PR #34, opened 2022-02-04
Wait for coordinator response before returning from 2PC client during mint

### LINKED ISSUE #32: Minting transaction never completes when minting a large number of coins (> 1000 UTXOs)
### Affected Branch

trunk

### Basic Diagnostics

### Description

After launching the system, run the following:
./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat mint 1000 2000

The above will try to mint one thousand (1000) $20 dollar bills.  After submission, this transaction will be stuck in pending state and never complete.  This happens when minting to a new wallet or an existing wallet.

### Code of Conduct

metalicjames:
The 2PC client should wait for the coordinator response to a mint operation. Otherwise, the client will exit before receiving the response meaning a manual sync is required to confirm the mint transaction. This PR adds a 5 second timeout before returning from send_mint_tx, which makes `client-cli` wait to exit until it has received a response or timed out.

Fixes #32.

### DIFF
--- a/src/uhs/client/twophase_client.cpp
+++ b/src/uhs/client/twophase_client.cpp
@@ -78,7 +78,9 @@ namespace cbdc {
     auto twophase_client::send_mint_tx(const transaction::full_tx& mint_tx)
         -> bool {
         auto ctx = transaction::compact_tx(mint_tx);
-        return m_coordinator_client.execute_transaction(
+        auto done = std::promise<void>();
+        auto done_fut = done.get_future();
+        auto res = m_coordinator_client.execute_transaction(
             ctx,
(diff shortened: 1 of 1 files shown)


dmtrinh:
Confirmed working @ https://github.com/mit-dci/opencbdc-tx/issues/32#issuecomment-1030679251.  Thanks @metalicjames!

HalosGhost (approved):
This is a very reasonable solution to #32: offer a timeout after-which we notify the user that the `mint` didn't complete as-expected (motivating the user to manually `sync` the wallet), and otherwise complete as-normal.

Merging.

PR #38, opened 2022-02-07
Retry joining followers to raft clusters indefinitely

metalicjames:
This PR adds a retry loop to the initial raft cluster setup phase executed by leaders. Previously, followers would already have to be running before the leader (node 0 in a cluster) could be started. Otherwise, the leader would fail to add all the follower nodes to the raft cluster and abort. While this was useful for testing and benchmarking, it is undesirable in a deployment environment as node startup order is hard to guarantee. Instead, the leader now retries adding followers to the raft cluster until all followers have been added.

### DIFF
--- a/src/uhs/atomizer/atomizer/atomizer_raft.cpp
+++ b/src/uhs/atomizer/atomizer/atomizer_raft.cpp
@@ -15,7 +15,8 @@ namespace cbdc::atomizer {
                                  const network::endpoint_t& raft_endpoint,
                                  size_t stxo_cache_depth,
                                  std::shared_ptr<logging::log> logger,
-                                 nuraft::cb_func::func_type raft_callback)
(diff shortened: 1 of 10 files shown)


HalosGhost (changes requested):
Really simple fix: put the cluster-building code in an infinite loop, and break out when it succeeds.

My one question is whether or not it would make sense to offer a config setting to set a limit on how long it would try to build the cluster? I'd hope that doing so would result in the ability to change the config to effectively restore the old behavior (to make benchmarking more accurate).

This commit is also missing a DCO sign-off.

HalosGhost (approved):
I think this is a good middle-ground: offer a boolean switch to enable cluster-build retry. In the future, maybe it makes sense to expand this to be a specific time-limit rather than a single on/off.

Looks good to me!

HalosGhost:
Merging.

PR #40, opened 2022-02-09
Fix #39: README issues w/ Docker Permissions

### LINKED ISSUE #39: README Incorrectly Details Docker Permissions
[here](https://github.com/mit-dci/opencbdc-tx#build-the-container), users are incorrectly told one needs a root shell for Docker. This is not true. The Docker installation instructions themselves detail specifically [how to avoid this](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user), and if you are using native mac/windows apps you probably don't even have to do that. 

I suggest either:
- no instructions are given, and users are trusted to know how to use Docker, or how to look that up themselves
- if such hand-holding is desired/necessary, we encourage best practices and point users towards the guide above rather than encourage unnecessary use of root permissions

tarfeef101:
- don't tell users to run as root
- use compose if we're going to use compose
- simplify commands for up/down/stop

Would be great if we could have a default `docker-compose` file that wouldn't require specifying the file each time, but not necessary. Would also be nice to not use a ubuntu base image, they're rather bloated. Would be even better to separate runtime and building images for faster builds and smaller images overall. I'd also like to not make the interactive container manually, but simply define a runtime container as a service in compose, and users could run `docker-compose run wallet`, for example, and that'd dump them into a shell within the environment specified (and since it's in the same compose file, the network is automatically correct).

If any/all of these additional changes are desired, that can be added to this or made as separate PRs too.

closes #39


README.md
```diff
@@ -55,26 +55,29 @@ The easiest way to compile the code and run the system locally is using [Docker]
 * [Install Docker](https://docs.docker.com/get-docker/)
 * [Install docker-compose](https://docs.docker.com/compose/install/)
 
-Don't forget to run the docker daemon!
+### Recommended post-install steps
+- Enable the Docker daemon as detailed [here](https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot)
```
```suggestion
- [Enable the Docker daemon](https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot)
```

Bit of a nit, but when possible, I like to follow  the [WCAG Guidelines for links](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html) and use more descriptive link text.

README.md
```diff
@@ -55,26 +55,29 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
-$ sudo -s                        # open a root shell (needed for docker)
-# docker build . -t opencbdc-tx  # build the container
+# docker-compose -f docker-compose-2pc.yml build
 ```
 
 ## Launch the System
+**Note:** You will need to both run the system and interact with it. To do this, you should run the system in **detached** mode (in the background), and then you can launch a second container in which to run commands. This is detailed below.
 
-**Note:** You will need to both run the system and interact with it; you can either use two shells, or you can add the `--detach` flag when launching the system (note that it will then remain running till you stop it, e.g., with `docker stop`).
 Additionally, you can start the atomizer architecture by passing `--file docker-compose-atomizer.yml` instead.
```
HalosGhost: This is a more general question: do you think it makes more sense for us to expand all flags that we mention (rather than using their short-option form) for clearer documentation?

If not, we should at least probably be consistent throughout the document so we're not referring to an option by its long and short form in different places.

README.md
```diff
@@ -55,26 +55,29 @@ The easiest way to compile the code and run the system locally is using [Docker]
 * [Install Docker](https://docs.docker.com/get-docker/)
 * [Install docker-compose](https://docs.docker.com/compose/install/)
 
-Don't forget to run the docker daemon!
+### Recommended post-install steps
+- Enable the Docker daemon as detailed [here](https://docs.docker.com/engine/install/linux-postinstall/#configure-docker-to-start-on-boot)
+  - It is recommended to read the other instructions here to customize other features you may find useful
+- Follow [these instructions](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) to enable the ability to run docker as a non-root user
```
HalosGhost: I'm trying to figure out if there's a sensible way to rephrase this to also have a more descriptive link text, but nothing's immediately coming to mind for me…

HalosGhost (changes requested):
This looks like a solid improvement!

I have a couple of small changes requested, and there are a few logistical things that should be handled as well:

* In accordance with [the Contribution Guide](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/contributing.md#intellectual-property), this commit needs a DCO to be considered for inclusion. (we're going to be adding a CI task to check this automatically soon so that this gets handled quickly and easily, sorry it's not in-place yet!)
* It would be great if the commit message header could also be more descriptive (e.g., `docs: improve README's docker walkthrough`)

HalosGhost:
> Would be great if we could have a default `docker-compose` file that wouldn't require specifying the file each time, but not necessary.

I think we're unlikely to go this route (because we want it to be as simple as possible to run each architecture, and not suggest that either one is official or default).

> Would also be nice to not use a ubuntu base image, they're rather bloated. 

This sounds interesting!
In part, we used ubuntu because it gives a simple target for people to use to know that a bare machine will work to run the code, but I'm definitely open to seeing alternatives!

> Would be even better to separate runtime and building images for faster builds and smaller images overall.

This is also interesting!
I'd love to see an example or hear more about how this would work (anything we can do to improve the code's usability is on-the-table).

> I'd also like to not make the interactive container manually, but simply define a runtime container as a service in compose, and users could run `docker-compose run wallet`, for example, and that'd dump them into a shell within the environment specified (and since it's in the same compose file, the network is automatically correct).
That sounds terrific.
I'd love to see a separate PR for that (and for several of the things above).

Thank you for pitching in a hand; welcome to the project!

README.md
tarfeef101: I would lean towards short-forms to encourage people to use things more efficiently, and where it is difficult to infer the meaning, to simply explain in the instruction relating to the command snippet.

README.md
tarfeef101: changed it a bit

tarfeef101:
> -- Snip --

re: more advanced dockering:

### Different base image
The de-facto standard of alpine for small/fast images is what I'd tend toward, I can make another PR using that instead. The biggest thing people run into there is that fact that it uses `musl` instead of `glibc` by default, but that may not actually be a problem, and if it is, it can be ameliorated, so that can be tested out fairly easily.

### Runtime vs Building images
Basically, we could do the following:
- define an image to build the executable(s). it will have all the dependencies to build the output program, and be layered in a way such that it can be built quickly, even if slightly larger, as compute time is generally more valuable than storage. For example:
```
# alpine:3.15.0
FROM alpine@sha256:e7d88de73db3d3fd9b2d63aa7f447a10fd0220b7cbf39803c803f2af9ba256b3
RUN install some deps
COPY some_infrequently_changed_files
RUN first_build_command
COPY frequently_changed_files
RUN second_build_command
```
- define a second runtime images with only runtime dependencies, and then the executable built by the builder image. For example:
```
# get binary from builder image
FROM previous_image:tag as builder
FROM alpine:latest
RUN install_runtime_deps
COPY --from=builder /opt/binary /opt
ENTRYPOINT ["/bin/sh", "-c", "/opt/binary", "--option", "value"]
```

This would be done so the runtime image can be distributed separately from the builder image, and be kept much smaller (it only needs runtime dependencies, and the binaries. In addition, the builder image would be smaller (for the same reasons, but only needing build dependencies), and could be built faster as a result.

### Wallet vs System Services
By leveraging the previous suggestion, you could have 2 different services laid out like this:
```
services:
  wallet:
    build: src/wallet/Dockerfile
    entrypoint: ["/bin/bash"]
    command: ["/opt/client-cli", "--relevant-option"]
  system:
    build: src/system/Dockerfile
    command: ["system_executable", "--relevant-option"]
```
Then you can run `docker-compose up -d system` (and add more services if more are required), and `docker-compose run wallet` will dump you right into a shell, and optionally (as shown above) in an interactive session within the executable if that's supported (if not just nix the `command` line and you'll just have a regular bash shell).

README.md
HalosGhost: I don't think that's unreasonable; can you walk through the rest of the README then and make sure that we refer to each option consistently?

README.md
```diff
@@ -55,26 +55,29 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
+- [Enable the ability to run docker as a non-root user](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)
+- Alternatively, advanced users may wish to run Docker in [rootless mode](https://docs.docker.com/engine/security/rootless/)
 
 ## Build the container
 
 ```terminal
-$ cd opencbdc-tx                 # change to the project directory
-$ sudo -s                        # open a root shell (needed for docker)
-# docker build . -t opencbdc-tx  # build the container
+# docker-compose -f docker-compose-2pc.yml build
```
HalosGhost: Here, and below, these commands are intended to be run as non-root, so they should be replaced with a `$` (Note though that the `#` inside the container is still correct and should be left as-is).

HalosGhost: I'm also wondering if we should call out that despite this command using the `docker-compose-2pc.yml` file, it's building the full image (so you don't need to rebuild to run the atomizer architecture). Thoughts?

HalosGhost (changes requested):

README.md
tarfeef101: I think that's worthwhile, one wouldn't intuitively expect multiple compose files to all use the same image, so that sounds like a good improvement.

HalosGhost:
> The de-facto standard of alpine for small/fast images is what I'd tend toward, I can make another PR using that instead. The biggest thing people run into there is that fact that it uses `musl` instead of `glibc` by default, but that may not actually be a problem, and if it is, it can be ameliorated, so that can be tested out fairly easily.

I was wondering if that's what you had in-mind. I like Alpine, and like the idea of smaller images, but I'm a little concerned that moving this direction might confuse the situation with `./scripts/configure.sh`. @metalicjames where do you come down on this?

> ### Runtime vs Building images

I think this could be really interesting. It offers a reasonable route to distributing images for people that just want to play around with the system (the runtime images). I do wonder though, code changes will necessitate rebuilding both images, right? How likely is it that you'd want to rebuild the runtime images without also rebuilding the builder images?

> ### Wallet vs System Services

This is also still very interesting to me, but I'm wondering where the line gets drawn. We have two separate compose files right now, at what point would it make more sense for this solution to define a service for each architecture, and combine the files? My immediate concern there is that it'd get harder to see the separation between the two architectures. What do you think about this?

metalicjames:
> where do you come down on this?

We don't use the configure script for the docker images right now, so switching to Alpine images is probably a good idea.

I'm a fan of all the proposals above, but they should be split out into separate PRs to keep this focused on the readme changes.

HalosGhost:
> We don't use the configure script for the docker images right now, so switching to Alpine images is probably a good idea.

I'm willing to accept that this may be an example of my being overly-cautious. ☺

> they should be split out into separate PRs to keep this focused on the readme changes.

Sounds perfect.

HalosGhost:
Note that we merged a small additional walkthrough to the README that should probably be updated in this PR.

HalosGhost:
@tarfeef101 following the correction mentioned in the NEWS post, this PR needs to be rebased.

PR #41, opened 2022-02-09
docs: Correct statistics in README

HalosGhost:
The statistics in the README are outdated, and do not refer to the
latest demonstrated numbers in the technical paper.

This minor change corrects the statistics to match the paper
appropriately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,12 @@ Both architectures handle multiple geo-distributed datacenter outages with a [re
 1. "Atomizer" architecture
     - Materializes a total ordering of all transactions settled by the system in a linear sequence of batches.
     - Requires vertical scaling as peak transaction throughput is limited by the performance of a single system component.
-    - Maximum demonstrated throughput ~100K transactions per second.
-    - Geo-replicated latency ~3 seconds.
+    - Maximum demonstrated throughput ~170K transactions per second.
+    - Geo-replicated latency <2 seconds.
 1. "Two-phase commit" architecture
     - Transaction history is not materialized and only a relative ordering is assigned between directly related transactions.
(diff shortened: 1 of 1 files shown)


anders94:
Good catch. I don't think we need to stipulate "(as of release of the technical paper)" though. In the future that might be slightly ambiguous.

HalosGhost:
@anders94 the reason I put in that parenthetical was so that we wouldn't need to keep updating the numbers if they change. Do you have a thought about how we can keep them evergreen?

README.md
```diff
@@ -24,12 +24,12 @@ Both architectures handle multiple geo-distributed datacenter outages with a [re
 1. "Atomizer" architecture
     - Materializes a total ordering of all transactions settled by the system in a linear sequence of batches.
     - Requires vertical scaling as peak transaction throughput is limited by the performance of a single system component.
-    - Maximum demonstrated throughput ~100K transactions per second.
+    - Maximum demonstrated throughput ~170K transactions per second.
     - Geo-replicated latency ~3 seconds.
```
metalicjames: I think we said <2 s in the paper.

metalicjames (approved):

anders94:
@HalosGhost Yeah, good point. Short answer is "no" - we'll have to remember to update here as things change.

README.md
anders94: Agreed. Let's update this as well.

anders94 (changes requested):
Per James' note, let's update to <2 seconds latency. Otherwise LGTM.

HalosGhost:
Requesting final double-check before merge!

anders94 (approved):
LGTM

PR #44, opened 2022-02-09
build: Check for buildtime support of cpu features

### LINKED ISSUE #42: Improve the build system's ability to react to the target platform
We've seen several instances of people trying to build the code on slightly (or wildly) different platforms from the one we “officially” support (namely, x64 Linux), and it resulting in similar build errors (cf. #35)

In particular, our build system currently hard-codes some flags that we want to set for performance, but aren't available on all platforms. It would be ideal if we could teach our CMake files to dynamically define performance-related flags based on the target architecture to fix these problems more systemically (than offering specific suggestions to people to fix the build for their target). This should also just generally make the code more portable!

theuni:
Addresses #42, though libsecp256k1 could use some feature detection as well. I suspect build is currently broken for 32bit targets.

Additionally, tests should be added for sse42 and crc for leveldb, but I don't believe their absence is causing any current build problems.

### DIFF
--- a/3rdparty/crypto/CMakeLists.txt
+++ b/3rdparty/crypto/CMakeLists.txt
@@ -7,12 +7,26 @@ add_library(crypto sha256_avx2.cpp
                    sha256.cpp
                    siphash.cpp)
 target_include_directories(crypto PRIVATE ".." ".")
-set_source_files_properties(sha256_avx2.cpp PROPERTIES COMPILE_FLAGS
-                                            "-DENABLE_AVX2 -mavx -mavx2")
-set_source_files_properties(sha256_shani.cpp PROPERTIES COMPILE_FLAGS
-                                             "-DENABLE_SHANI -msse4 -msha")
-set_source_files_properties(sha256_sse41.cpp PROPERTIES COMPILE_FLAGS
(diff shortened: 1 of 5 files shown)


theuni:
Test failure appears to be a false positive.

HalosGhost (changes requested):
@theuni We're working on getting the DCO check added to the CI so this check will happen automatically, but these commits do not contain a sign-off. Would you mind either amending both commits with a sign-off (`git commit --amend -s`) or squashing and signing-off on the squashed commit?

HalosGhost:
@metalicjames I think this is an excellent improvement to the build system, but would love a second set of eyes to review before I merge it.

HalosGhost (approved):
Separates out several x86-focused instruction sets and tests for their availability before actually leveraging them.

I think the idea is excellent and the implementation looks good. I'd like to get someone else to review quickly since cmake isn't the build-system I'm most familiar with.

metalicjames (approved):
I tested this PR locally on a Raspberry Pi running Ubuntu. With this patch I can configure and build the project out of the box.

HalosGhost:
Looks good on several counts. Merging!

PR #45, opened 2022-02-10
Align NuRaft version Dockerfile with configure.sh

wadagso-gertjaap:
The NuRaft version used in `scripts/configure.sh` when building locally is 1.3.0 presently.

However, when building using Docker we're using NuRaft version 1.2.0 as specified in the `Dockerfile`

While it doesn't seem to influence the ability to produce working binaries inside the Docker build system, I think we should make these two versions equal such that the binaries created in the Docker environment are the same as local builds.


metalicjames (approved):

HalosGhost:
Very simple change that brings Docker and local configuration in-line. Merging.

PR #46, opened 2022-02-10
Run make commands in Docker on all cores

wadagso-gertjaap:
In the regular build script `scripts/build.sh` we utilize the maximum number of cores when building by adding the `-j {cores}` argument. This is not present in the `Dockerfile`, which causes the Docker build to be unnecessarily slow - especially on systems that have a lot of cores.

This change will utilize the available number of cores when building.

On my machine (Ryzen 9 3900X) , this makes a huge difference:

Before:
```
real	4m54,257s
user	0m0,099s
sys	0m0,103s
```

After:
```
real	1m21,746s
user	0m0,087s
sys	0m0,069s
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN wget https://github.com/google/leveldb/archive/${LEVELDB_VERSION}.tar.gz && 
     rm -f ${LEVELDB_VERSION}.tar.gz && \
     cd leveldb-${LEVELDB_VERSION} && \
     cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DLEVELDB_BUILD_TESTS=0 -DLEVELDB_BUILD_BENCHMARKS=0 -DBUILD_SHARED_LIBS=0 . && \
-    make && \
+    make -j$(nproc) && \
     make install
 
 # Install NuRaft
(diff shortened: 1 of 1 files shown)


HalosGhost:
Confirmed dramatic improvement:

# Before

```console
$ time docker build . -t opencbdc-tx
[snip]
docker build . -t opencbdc-tx  0.10s user 0.10s system 0% cpu 4:20.23 total
```

# After

```console
$ time docker build . -t opencbdc-tx
[snip]
docker build . -t opencbdc-tx  0.06s user 0.07s system 0% cpu 1:03.14 total
```

Going to double-check that the initial README walkthrough is still functional (though it'd surprise me if it weren't) before merging.

wadagso-gertjaap:
@HalosGhost If you want to validate the efficiency gain, you'll have to run the builds without cache:

```time docker build . --no-cache -t opencbdc-tx```

HalosGhost:
In this case, I had rebuilt the image against different branches, so it should have been a reasonable comparison. But I'll double-check it before I walk-through the README.

HalosGhost:
Updated numbers: local clean docker image build drops from 5:11.84 to 2:15.66.

And I've confirmed README walkthrough works as-expected. Merging!

HalosGhost (commented):
Simple fix to dramatically improve docker image build times.

PR #47, opened 2022-02-10
Add missing cbdc::atomizer::request

wadagso-gertjaap:
In the `atomizer-cli-watchtower`, when testing without signing, the transactions are directly delivered to the atomizer. However, in 2/3 cases the request isn't wrapped in a `cbdc::atomizer::request{}` as introduced [here](https://github.com/mit-dci/opencbdc-tx/commit/8db8be1d5ce93066917932734f908e01596f03f1)

Without this fix, the atomizer is reporting

```Invalid request packet```

And the `atomizer-cli-watchtower` is reporting: 

```Failed to send pay tx to atomizer. ID:{tx id} h:{height}```


metalicjames (approved):

HalosGhost:
Looks like a simple fix! Merging.

PR #58, opened 2022-02-15
Add docs to readme.md for running test suite in Docker

cdesch:
- Add docs for running the test suite in docker for enabling contributors to test their code prior to submitting
- Add `lcov` to base `Dockerfile` as a dependency so tests that run in the docker container can produce code coverage metrics

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ RUN apt update && \
       libgtest-dev \
       libgmock-dev \
       net-tools \
+      lcov \
       git
 
 # Args
--- a/README.md
+++ b/README.md
@@ -139,3 +139,16 @@ In each of the below commands, you should pass `atomizer-compose.cfg` instead of
   # ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool1.dat wallet1.dat info
   Balance: $0.30, UTXOs: 1, pending TXs: 0
   ```
+
+## Testing
+
+Running Unit & Integration Tests
+
+1. Build the container
+   ```terminal
+   # docker build . -t opencbdc-tx
+   ```
+2. Run Unit & Integration Tests
+   ```terminal
+   # docker run -ti opencbdc-tx ./scripts/test.sh
+   ```


Dockerfile
```diff
@@ -6,13 +6,14 @@ ENV DEBIAN_FRONTEND noninteractive
...
-      net-tools \
-      git
+    build-essential \
+    wget \
+    cmake \
+    libgtest-dev \
+    libgmock-dev \
+    net-tools \
+    lcov \
+    git
```
HalosGhost: The purpose of the extra indent was for visual clarity that each of these lines is an argument to the command `apt install -y` rather than a new command. Additionally, by dedenting all these lines, it hides the real change (which is the addition of `lcov` to the list).

Unless there's a good reason we should remove the indentation, I think we should leave it as-is.

Dockerfile
```diff
@@ -6,13 +6,14 @@ ENV DEBIAN_FRONTEND noninteractive
...
-      libgmock-dev \
-      net-tools \
-      git
+    build-essential \
+    wget \
+    cmake \
+    libgtest-dev \
+    libgmock-dev \
+    net-tools \
+    lcov \
```
HalosGhost: Allowing the scripts we provide to be run inside the docker containers is probably a good idea (if for no other reason, because of the principle of least surprise).

However, just to note, we are compatible with [`act`](https://github.com/nektos/act), so you can run the full CI suite locally (in-docker) before opening a PR.

README.md
```diff
@@ -139,3 +139,45 @@ In each of the below commands, you should pass `atomizer-compose.cfg` instead of
...
+
+## Testing
+
+Running Unit Tests
+
+1. Build the container
+   ```terminal
+   # docker build . -t opencbdc-tx
+   ```
+2. Run Unit Tests
```
HalosGhost: This will actually run the full test-suite (both unit and integration tests).

README.md
```diff
@@ -139,3 +139,45 @@ In each of the below commands, you should pass `atomizer-compose.cfg` instead of
...
+   ```terminal
+   # docker run -ti opencbdc-tx ./scripts/test.sh
+   ```
+
+Running Integration Tests
```
HalosGhost: This is not an integration test; it's a *very* simple system test which lives outside our test suite. It's essentially just a short-cut for the walkthrough we provide above for ease-of-use.

I don't think it makes sense to document this script here; if it moves to become part of a more fully-featured system-test suite under `test/`, then I think that's a different story.

HalosGhost (changes requested):
Thank you for the PR! There are a few changes that I think should be made.

I also have a question: this change allows `./scripts/test.sh` to be run under docker; should we go the next step and also setup `./scripts/lint.sh` to be run under docker?

If not, where's the line for what we should enable to run inside docker and what we shouldn't? If so, does it make more sense for us to document using `act` to run the full CI locally? (It's definitely a simpler process than having to manually run the containers.)

Dockerfile
cdesch: I believe the indentation change was a side effect of the VS Code Docker plugin auto-formatting the file on save. I think it is okay to leave the indentation as it was, although this might become a recurring issue for contributors.

cdesch:
@HalosGhost I think it is a good idea to add the `clang-format` for linting.  If you're gonna have one docker image for development and running, allowing that one image to both build, test, lint and run makes it easier for new contributors to pick up the code and get started. If the docker image needs to be slimmed down because of these additional dependencies that aren't needed at runtime, I think it would be better to have a `dev` Dockerfile and a `"prod"` Dockerfile, even though the system is Production, by design 👍.

I can make a separate PR for  `clang-format` and `./scripts/lint.sh` if you would like.

README.md
cdesch: @HalosGhost Thanks. I think so too given that context. It helps to know those differences. I removed it here: https://github.com/mit-dci/opencbdc-tx/pull/58/commits/56150189dec9b6939d13f79673ea4d650cc232ea

Dockerfile
cdesch: I restored the indentation in this commit https://github.com/mit-dci/opencbdc-tx/pull/58/commits/a27f2100fa7da1ee4f27312440e93808a4b51b32 , although I had to disable the VS Code Docker Extension. You may run into this problem again with other contributors.

Dockerfile
cdesch: I think this is a good idea

cdesch (commented):
Updated PR per request

cdesch:
@HalosGhost I'm not sure if there is a github action for closing the review or not. Not sure if there is something else you want me to do.

HalosGhost:
@cdesch sorry about the delay! In the top-right of the page (in the Reviewers section), there's a button to re-request a review ([cf. step 7](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review)). That's usually my cue to take another look at a PR.

I think this PR looks pretty good now, but I think the commits should be squashed to a single commit combining the texts.

cdesch:
Thanks @HalosGhost , I squashed the commits into https://github.com/mit-dci/opencbdc-tx/commit/6c1abf0ce512d6e90ab746c21a004c450950c8b1 and performed the "Re-Review" action. That was kind of hidden up there. Thanks for guiding me.

HalosGhost (approved):
This looks good to me. Adds a dependency to the docker image so that `./scripts/test.sh` can be run in-docker, and includes a basic walkthrough for users to do so in the README.

Will merge after all checks have finished running (unless something surprising happens).

HalosGhost:
@cdesch happily! Welcome aboard!

PR #59, opened 2022-02-15
Calculate size of object and allocate before serialization

metalicjames:
This PR clears a TODO in the serialisation helpers. Pre-calculates the serialised size of the object and allocates the required memory in the buffer before performing the actual serialisation. This reduces the number of heap allocations to one per use of `make_buffer` or `make_shared_buffer`.


HalosGhost (approved):
This looks good to me. Leverages a helper we already have to pre-allocate memory for serialization rather than relying on the serialization itself automatically increasing the allocation size.

Merging.

PR #60, opened 2022-02-15
Support multiple consumers from blocking queue

metalicjames:
This PR fixes `cbdc::blocking_queue` so that it can support multiple consumers popping from the queue. Prior to this change `pop()` would set the wake-up flag even if no element was popped from the queue. This prevented multiple consumers from unblocking after a `clear()` call.

### DIFF
--- a/src/util/common/blocking_queue.hpp
+++ b/src/util/common/blocking_queue.hpp
@@ -66,9 +66,8 @@ namespace cbdc {
                     item = std::move(m_buffer.front());
                     m_buffer.pop();
                     popped = true;
+                    m_wake = !m_buffer.empty();
                 }
-
-                m_wake = !m_buffer.empty();
 
                 return popped;
             }


HalosGhost:
@metalicjames This seems pretty reasonable. I wonder, if an element is popped, should we *always* wake, or do we want to preserve the possibility that an element gets popped and we don't wake.

From my reading of the diff, it looks like we re-evaluate whether the buffer is empty to determine if we should wake. If we want to always wake-when-popped, it may make more sense to just set `m_wake` to `true` inside that block.

Thoughts?

metalicjames:
We shouldn't wake up again if there are no more elements in the queue. In that case we should wait to be notified by the producer. We should only change the wait flag if we popped a value from the queue. If we were woken up and the queue was empty, that means `clear()` has been called, and the goal is to unblock any consumers. So wake should remain the value it was before (true), so that all consumers are notified and unblock.

HalosGhost (approved):
Following some discussion with @metalicjames, this enables some more flexibility for the blocking queue, and preserves the behavior we expect.

Merging.

PR #61, opened 2022-02-15
Implement serialization for cbdc::buffer

metalicjames:
This PR implements serialisation for `cbdc::buffer`. The size is serialised first (as a 64-bit unit), followed by the actual buffer data.

### DIFF
--- a/src/util/serialization/format.hpp
+++ b/src/util/serialization/format.hpp
@@ -7,6 +7,7 @@
 #define OPENCBDC_TX_SRC_SERIALIZATION_FORMAT_H_
 
 #include "serializer.hpp"
+#include "util/common/buffer.hpp"
 #include "util/common/config.hpp"
 #include "util/common/variant_overloaded.hpp"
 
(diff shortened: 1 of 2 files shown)


src/util/serialization/format.hpp
```diff
@@ -34,6 +35,9 @@ namespace cbdc {
     /// \see \ref cbdc::operator<<(serializer&, std::byte)
     auto operator>>(serializer& packet, std::byte& b) -> serializer&;
 
+    auto operator<<(serializer& ser, const buffer& b) -> serializer&;
```
HalosGhost, line 44: Keeping inline with the rest of the serialization code, can these two methods be given doc comments?

HalosGhost (changes requested):
Minor change (just some comments for documentation) requested.

HalosGhost (approved):
Looks like a nice flexibility improvement. Allows us to directly {de,}serialize buffers themselves.

Merging.

PR #62, opened 2022-02-15
Expose parse_ip_port function from config namespace

metalicjames:
This PR exposes the `parse_ip_port` function used by the configuration parser so that it can be used outside of that context. Also adds a TODO for error handling, given the function currently requires a correctly parseable string as input.

### DIFF
--- a/src/util/common/config.cpp
+++ b/src/util/common/config.cpp
@@ -11,6 +11,7 @@
 
 namespace cbdc::config {
     auto parse_ip_port(const std::string& in_str) -> network::endpoint_t {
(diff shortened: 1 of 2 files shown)


HalosGhost:
Can you expand on the value of exposing this method in the commit text? This is a very small change, but it's not immediately clear to me what benefit it offers.

metalicjames:
The function is used by some other code I'm working on. This PR is in preparation for that.

HalosGhost (approved):
I don't think there's any drawback to this. Simply exposes a helper we have for wider-use.

Merging.

PR #63, opened 2022-02-15
Implement siphash algorithm for cbdc::buffer

metalicjames:
This PR implements a template specialisation for `const_sip_hash` using `cbdc::buffer`. This allows a buffer to be a key in hash maps.

### DIFF
--- a/src/util/common/hashmap.hpp
+++ b/src/util/common/hashmap.hpp
@@ -6,6 +6,7 @@
 #ifndef OPENCBDC_TX_SRC_COMMON_HASHMAP_H_
 #define OPENCBDC_TX_SRC_COMMON_HASHMAP_H_
 
+#include "buffer.hpp"
 #include "crypto/siphash.h"
 #include "hash.hpp"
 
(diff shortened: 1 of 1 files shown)


HalosGhost (commented):
This looks like a solid, simple specialization (mirroring the primary definition almost verbatim).

I'm inclined to immediately approve this, but I do want to stop for a moment and ask about the potential negatives of using a hard-coded initialization key. It's what we're already doing in the primary definition, but I'm wondering if now is a good moment to clear that to-do.

Ktamas794 (approved):

metalicjames:
If the initialisation key is hard-coded, the benefit of using sip hash is largely eliminated. The goal of the initialisation is that it's unknown to those outside the system, so an adversary can't predict the bucket a particular element will be assigned, and thus degrade the performance of hash maps.

Changing the initialisation would change a lot of files and I think should be left for a different PR.

HalosGhost (approved):
It probably makes sense for us to address the hard-coded initialization key separately.

Merging.

PR #55, opened 2022-02-14
Fix mac os builds

dmtrinh:
Resolves build failures on macOS/Darwin - issue #36.  The conditional block in CMake will check whether the environment is Darwin and add `-fno-rtti` to `CMAKE_CXX_FLAGS`.  We also need to detect where the default location for dev tools is.  That will be used as the base directory for the `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` in `build.sh`

Issue #36, opened 2022-02-07
Build fails on MacOS 


dmtrinh:
### Affected Branch

trunk

### Basic Diagnostics

- [X] I've pulled the latest changes on the affected branch and the issue is still present.

- [ ] The issue is reproducible in docker

### Description

Trying to build this directly after cloning `opencbdc-tx` repository.  Linking failed with:
```
Undefined symbols for architecture x86_64:
  "typeinfo for leveldb::Comparator", referenced from:
      typeinfo for cbdc::raft::index_comparator in libraft.a(index_comparator.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Environment
```
(base) ---------- 20:24:48 (duc@DucMacBook) :~/Dev2/opencbdc-tx ----------
==> clang --version
Apple clang version 12.0.5 (clang-1205.0.22.11)
Target: x86_64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
(base) ---------- 20:25:25 (duc@DucMacBook) :~/Dev2/opencbdc-tx ----------
==> system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro15,1
      Processor Name: 6-Core Intel Core i7
      Processor Speed: 2.6 GHz
      Number of Processors: 1
      Total Number of Cores: 6
      L2 Cache (per Core): 256 KB
      L3 Cache: 9 MB
      Hyper-Threading Technology: Enabled
      Memory: 16 GB
      System Firmware Version: 1554.140.20.0.0 (iBridge: 18.16.14759.0.1,0)
      Activation Lock Status: Disabled
```

### To Reproduce
1.  `git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx`
2.   Change into `opencbdc-tx` directory and run `scripts/configure.sh`
```
(base) ---------- 20:14:06 (duc@DucMacBook) :~/Dev2/opencbdc-tx ----------
==> scripts/configure.sh
Configuring...
Warning: leveldb 1.23 is already installed and up-to-date.
To reinstall 1.23, run:
  brew reinstall leveldb
Warning: llvm@11 11.1.0_4 is already installed and up-to-date.
To reinstall 11.1.0_4, run:
  brew reinstall llvm@11
[...]

PR #64, opened 2022-02-15
Disable RTTI and fix macOS builds

dmtrinh:
Resolves build failures on macOS/Darwin - issue #36.  Replacement for PR #55; please refer to that PR for prior thread with @metalicjames.  Note that disabling RTTI should also allow us to build against the latest LevelDB 1.23 on Linux.

### DIFF
--- a/3rdparty/crypto/CMakeLists.txt
+++ b/3rdparty/crypto/CMakeLists.txt
@@ -7,12 +7,26 @@ add_library(crypto sha256_avx2.cpp
                    sha256.cpp
                    siphash.cpp)
 target_include_directories(crypto PRIVATE ".." ".")
-set_source_files_properties(sha256_avx2.cpp PROPERTIES COMPILE_FLAGS
-                                            "-DENABLE_AVX2 -mavx -mavx2")
-set_source_files_properties(sha256_shani.cpp PROPERTIES COMPILE_FLAGS
-                                             "-DENABLE_SHANI -msse4 -msha")
-set_source_files_properties(sha256_sse41.cpp PROPERTIES COMPILE_FLAGS
-                                             "-DENABLE_SSE41 -msse4.1")
-set_source_files_properties(sha256.cpp PROPERTIES COMPILE_FLAGS
-                                       "-DUSE_ASM -DENABLE_SHANI \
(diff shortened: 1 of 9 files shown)


HalosGhost:
@dmtrinh there are still merge commits in this PR. `git rebase origin/trunk` should actually automatically remove merge commits, so that's probably the next thing to do (reading up on [rebasing](https://git-scm.com/docs/git-rebase) may be useful too).

If it's easier/quicker, I can also cherry-pick the two commits into a separate PR and merge.

HalosGhost (changes requested):
Needs squashing.

dmtrinh:
> Needs squashing.

@HalosGhost Merge commits are part of the set to be squashed so git is complaining. :-(  It might be quicker if you cherry-pick the one commit [4754f5d](https://github.com/mit-dci/opencbdc-tx/pull/64/commits/4754f5d7846e938ee4907e9f5376e6442f2a845f) on your side.  Just let me know how you'd like to proceed...  Thanks!

HalosGhost:
@dmtrinh okay, I just checked out your work locally and I know what happened (it's from this branch attempting to build off your trunk branch rather than starting sync'd with `upstream/trunk`). In short, it's trying to re-apply d415932 which is why the merge commits are showing up. 

I've opened #65 as a replacement PR and will merge after all checks pass.

After it's merged, I'd strongly suggest resyncing your `trunk` branch with upstream's `trunk` (it may make sense for us to add a bit of guidance on how to do this since it will be a frequent requirement). The easiest way to do this (in my opinion) is to add an `upstream` remote:

```console
$ git remote add upstream 'https://github.com/mit-dci/opencbdc-tx'
```

Once you have an `upstream` remote, you can do the following to sync your local `trunk`:

```console
$ git switch trunk
$ git pull upstream trunk
```

Because your fork's `trunk` differs from `upstream/trunk`, you may need to do something a little more aggressive (e.g., using ``--force``); this is a reason why it is much easier to leave your fork's `trunk` as little more than a shortcut and to do most/all work that you intend to PR on other branches (so that syncing `trunk` is always trivial). I hope that explanation helped!

dmtrinh:
Very helpful as always, thanks Sam!!

PR #66, opened 2022-02-17
Simplify shard seeder tool to use config file

metalicjames:
This PR simplifies the shard seeder tool usage. Rather than using command line options, use the configuration file to determine pre-seed generation parameters. Also unifies the file names generated by the shard seeder with those expected by the shards themselves when loading the pre-seed.


HalosGhost (approved):
This looks like an excellent refactor to improve the seeder's usability. Should hopefully make it easier for people to run experiments and benchmarks.

Merging.

PR #67, opened 2022-02-17
Update default config values to use fixed size transactions for benchmarking

metalicjames:
This PR updates the default configuration parameters to send fixed size 2-in, 2-out transactions from the benchmarking tools. This brings the default values in line with those we used for testing in the paper, and result in better performance out of the box than the previous default.


HalosGhost (approved):
Really simple and solid change: just updates the values for benchmarking-related settings to more reasonable defaults.

Merging.

PR #68, opened 2022-02-20
Added support for debugging with containers

dmtrinh:
Some folks may find it useful to be able to step through the code to understand how the system works.  This PR makes it easier for people to debug.  Support is added as a separate Docker image (see `Dockerfile.debug`), leaving the original image untouched for regular use of the system.

Key aspects of the new image are:
* Contains additional packages to support debugging
* Builds everything with debug information
* Exposes ports 22 and 1234 for remote use


dmtrinh:
Closing due to local merge commits.  Will add an upstream remote...

PR #69, opened 2022-02-20
Added support for debugging with containers

dmtrinh:
Some folks may find it useful to be able to step through the code to understand how the system works. This PR makes it easier for people to debug. Support is added as a separate Docker image (see Dockerfile.debug), leaving the original image untouched for regular use of the system.

Key aspects of the new image are:
* Contains additional packages to support debugging
* Builds everything with debug information
* Exposes ports 22 and 1234 for remote use


HalosGhost:
This is interesting! I do like the idea of making it easier to debug the system live. Out of interest, how large are the resultant images? @metalicjames, the docker files are serving as an easy way for people to test against a reasonably predictable target, but I don't think debugging is too far outside that goal. What are your thoughts?

dmtrinh:
> Out of interest, how large are the resultant images? 

@HalosGhost, the original image came in at 917MB in size.  The new image with additional packages for debugging is 2.25GB.  The new image is built via a separate Dockerfile and can also have its own tag (e.g. Debug) to not interfere with the original image:
```
(base) ---------- 21:49:41 (duc@DucMacBook) :~/Dev2/opencbdc-tx_dmtrinh ----------
==> docker images
REPOSITORY                             TAG              IMAGE ID       CREATED          SIZE
opencbdc-tx                            Debug            04b1794a5619   10 seconds ago   2.25GB
opencbdc-tx                            latest           3a60c3e81e6d   4 minutes ago    917MB
``` 
What we're providing here is some additional support for folks looking to better understand/get started on the project.

HalosGhost:
I spoke with @metalicjames offline and I think we're on the same page that this is a great idea. I'd love to see (à la #40) the possibility of all images we have based on one image to make building each image faster and the resulting images smaller, but I think this is very worth doing.

I'm going to be pulling this code over the next day or so to test it locally and see how it goes.

@dmtrinh can you clarify why debugging requires the disabling of seccomp? (e.g., is that because of the tracing syscalls used by debuggers or something else?) In particular, is it possible for us to define a seccomp profile that will allow debugging but won't be in the completely open `unconfined` mode?

.gitignore
```diff
@@ -34,6 +34,7 @@ CMakeCache.txt
 *.o-*
 *.a
 *.log
+*log*
```
HalosGhost, line 37: Why is this line added?

It actually results in running the CI locally with [`act`](https://github.com/nektos/act) failing (I guess because it uses `git` to determine which files should be copied?), and I'm not sure what the benefit is.

README.md
```diff
@@ -44,7 +44,9 @@ For more information on how to contribute, please see our [Contribution Guide](d
 
 1. [Install Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 1. Clone the repository (including submodules)
-    - `git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx`
+   ```
+   git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx
```
HalosGhost, line 48:
```suggestion
   ```terminal
   $ git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx
```

I think it's pretty reasonable to improve the formatting of this block despite that it's unrelated to the other work in this PR. But, if we're going to make the change, let's unify the style so the terminal blocks all look the same.

README.md
```diff
@@ -152,3 +154,18 @@ Running Unit & Integration Tests
...
+1. Build the Debug container
+   ```terminal
+   # docker build -t opencbdc-tx:Debug -f Dockerfile.debug .
+   ```
+2. Launch the Debug container.  **Note:**  The container exposes two ports, `22` (ssh) and `1234`.  The latter can be used for [gdbserver](https://sourceware.org/gdb/onlinedocs/gdb/Server.html) or [lldb-server](https://lldb.llvm.org/man/lldb-server.html).  Debugging requires running privileged operations so the container needs to be run in [unconfined mode](https://docs.docker.com/engine/security/seccomp/#run-without-the-default-seccomp-profile).
+   ```terminal
+   # docker run -d -p 2222:22 -p 1234:1234 --rm --security-opt seccomp:unconfined --name opencbdc-tx opencbdc-tx:Debug
+   ```
+
+Both the source and build output remain in the container's `/opt/tx-processor` as before.  Use `docker exec` or ssh into the container to start gdb or lldb.  Lastly, configure your favorite IDE as appropriate to make use of the container.
```
HalosGhost, line 171: I think this section should be expanded to actually show a short example of connecting to debugging container and executing `gdb` to start debugging; at the moment, it walks-through the overhead but then doesn't give an example of how a user could leverage this improvement.

README.md
```diff
@@ -152,3 +154,18 @@ Running Unit & Integration Tests
    ```terminal
    # docker run -ti opencbdc-tx ./scripts/test.sh
    ```
+
+## Debugging
+
+Running in debug mode can be useful for those who want more insights into how the system works under the covers.  A dedicated container is used for debugging.
```
HalosGhost, line 160: Here and below:

In any `.md` file, we try to have a single sentence per-line (so that diffs which modify a sentence don't have to modify the whole associated paragraph).

It makes review and semantic history much cleaner. Please line-break after each full-stop.

README.md
```diff
@@ -152,3 +154,18 @@ Running Unit & Integration Tests
...
+
+1. Build the Debug container
+   ```terminal
+   # docker build -t opencbdc-tx:Debug -f Dockerfile.debug .
+   ```
+2. Launch the Debug container.  **Note:**  The container exposes two ports, `22` (ssh) and `1234`.  The latter can be used for [gdbserver](https://sourceware.org/gdb/onlinedocs/gdb/Server.html) or [lldb-server](https://lldb.llvm.org/man/lldb-server.html).  Debugging requires running privileged operations so the container needs to be run in [unconfined mode](https://docs.docker.com/engine/security/seccomp/#run-without-the-default-seccomp-profile).
+   ```terminal
+   # docker run -d -p 2222:22 -p 1234:1234 --rm --security-opt seccomp:unconfined --name opencbdc-tx opencbdc-tx:Debug
```
HalosGhost, line 168: Can you expand on why `unconfined` is necessary? In particular, is `unconfined` actually necessary, or can we request other (less general) permissions that would still allow debugging to work?

HalosGhost (changes requested):
Sorry it took me longer to review than I had expected; a few things behaved somewhat strangely that I had to dig into.

I've requested a few changes in-line; let me know if you have any questions!

HalosGhost:
@dmtrinh, following the correction mentioned in the NEWS post, this PR needs to be rebased.

PR #77, opened 2022-03-11
Fix #4 Migrated SHA256 to SHA3

### LINKED ISSUE #4: Consider using SHA3
SHA256, currently used, has a number of points one could improve upon:
- it does not protect against [length extension attacks](https://en.wikipedia.org/wiki/Length_extension_attack) (not currently relevant for our work). This is cited as a reason [why Bitcoin uses double-SHA](https://bitcoin.stackexchange.com/questions/6037/why-are-hashes-in-the-bitcoin-protocol-typically-computed-twice-double-computed).
- it does not support explicit personalization fields for domain separation (so a solution for mit-dci/opencbdc-tx#5 is to use `H(H(tag) || H(tag) || data)` for SHA256).

While SHA256 design has been around for a long while, similar designs SHA0 and SHA1 are broken, and cryptographic community's hash function design techniques have improved in the 20 year period between designs of SHA256 and SHA3.

In this proposal, we would still keep SHA256 for Schnorr signatures as libsecp is a highly complex codebase we don't need to edit but could use SHA3 everywhere else. Most NIST PQC signature candidates also rely on SHA3 so it is likely we would bring it at some point in the future.

DeanWen:
1. Imported sha3.h, span.h and sha3.cpp from Bitcoin repo
2. Updated reference code from sha256 to sha3

closes #4  

@HalosGhost


src/uhs/transaction/transaction.cpp
```diff
@@ -44,11 +44,13 @@ namespace cbdc::transaction {
     auto input::hash() const -> hash_t {
         auto buf = cbdc::make_buffer(*this);
 
-        CSHA256 sha;
+        SHA3_256 sha;
         hash_t result;
 
-        sha.Write(buf.c_ptr(), buf.size());
-        sha.Finalize(result.data());
+        Span buf_span{(const unsigned char*)buf.data(), buf.size()};
```
HalosGhost: We generally try to avoid C-style casts (largely because C++ casts are much more easily searchable, get better compile-time checking, and convey intention much more clearly); so using either `static_cast` or `reinterpret_cast` (as makes sense at the call-site) is preferable. Here and below.

src/uhs/transaction/transaction.cpp
```diff
@@ -101,17 +105,20 @@ namespace cbdc::transaction {
     auto uhs_id_from_output(const hash_t& entropy,
                             uint64_t i,
                             const output& output) -> hash_t {
-        CSHA256 sha;
+        SHA3_256 sha;
         hash_t ret;
-        sha.Write(entropy.data(), entropy.size());
+        Span entropy_span{entropy.data(), entropy.size()};
```
HalosGhost: At the end of this function, you `Finalize` against a plain `hash_t` (which is really just an alias for a particular specialization of `std::array<>`), then it sounds like there's an implicit conversion between `std::array<>` and `Span`. If that's the case we may not need to explicitly convert our data to `Span`s (and, as a result, possibly avoid the need for a lot of the explicit casts as well). Is this viable?

HalosGhost (changes requested):
Welcome!

I left a couple of notes inline to be looked at; however, there are also a couple of house-keeping things to take a look at as well:

* Note that all commits must [include a DCO](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/contributing.md#intellectual-property) to be considered for inclusion.
* The lint-check failed; the CI task output includes more information on what needs to be fixed.
* Your commit messages do not convey what was changed or why. The [getting-started guide](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/getting-started.md#what-does-a-good-commit-look-like) includes more information on how to write good commit messages.

src/uhs/transaction/transaction.cpp
DeanWen: Updated

src/uhs/transaction/transaction.cpp
DeanWen: Updated, I still need to construct a Span{}, I did it inline with static_cast

HalosGhost:
This looks a lot cleaner; there's one failing CI check (just looks like you need to run `clang-format` on a file or two, so a pretty easy fix).

After you make that small fix, and the CI passes, I'll take a final look (and probably have one other person double-check it) and we'll get it merged!

DeanWen:
> This looks a lot cleaner; there's one failing CI check (just looks like you need to run `clang-format` on a file or two, so a pretty easy fix).
> 
> After you make that small fix, and the CI passes, I'll take a final look (and probably have one other person double-check it) and we'll get it merged!

I think this should be fixed, let me know if it still happens @HalosGhost

HalosGhost (approved):
Looks like all my concerns have been addressed and this is now looking pretty clean. ACK from me.

I've requested the review of a couple other folk who might be able to offer a second opinion.

3rdparty/crypto/sha3.h
```diff
@@ -0,0 +1,41 @@
...
+  public:
+    static constexpr size_t OUTPUT_SIZE = 32;
+
+    SHA3_256() {}
+    SHA3_256& Write(Span<const unsigned char> data);
+    SHA3_256& Finalize(Span<unsigned char> output);
+    SHA3_256& Reset();
+};
+
+#endif // BITCOIN_CRYPTO_SHA3_H
```
metalicjames, line 41: Add EOF newline

3rdparty/crypto/sha3.cpp
```diff
@@ -0,0 +1,250 @@
...
+    }
+    return *this;
+}
+
+SHA3_256& SHA3_256::Reset() {
+    m_bufsize = 0;
+    m_pos = 0;
+    std::fill(std::begin(m_state), std::end(m_state), 0);
+    return *this;
+}
```
metalicjames, line 250: EOF newline

metalicjames, line 250: EOF newline

3rdparty/crypto/span.h
```diff
@@ -0,0 +1,378 @@
...
+
+/** Like the Span constructor, but for (const) unsigned char member types only.
+ * Only works for (un)signed char containers. */
+template<typename V>
+constexpr auto MakeUCharSpan(V&& v)
+    -> decltype(UCharSpanCast(Span{std::forward<V>(v)})) {
+    return UCharSpanCast(Span{std::forward<V>(v)});
+}
+
+#endif // BITCOIN_SPAN_H
```
metalicjames, line 378: EOF newline

metalicjames, line 378: EOF newline

3rdparty/crypto/span.h
```diff
@@ -0,0 +1,378 @@
+// Copyright (c) 2018-2021 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_SPAN_H
```
metalicjames, line 5: I'd rather not bring in this custom implementation of span if we can avoid it (instead we should use `std::span` once C++20 is widely available).

Can we convert `Write` and `Finalize` to use the same interface as the existing `CSHA256` class? That way we don't need this header.

metalicjames, line 5: I'd rather not bring in this custom implementation of span if we can avoid it (instead we should use `std::span` once C++20 is widely available).

Can we convert `Write` and `Finalize` to use the same interface as the existing `CSHA256` class? That way we don't need this header.

3rdparty/crypto/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ add_library(crypto sha256_avx2.cpp
                    sha256_sse4.cpp
                    sha256_sse41.cpp
                    sha256.cpp
```
metalicjames, line 7: I don't think `CSHA256` is used anywhere in the codebase after this PR. We should remove the SHA256 source files.

metalicjames, line 7: I don't think `CSHA256` is used anywhere in the codebase after this PR. We should remove the SHA256 source files.

3rdparty/crypto/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ add_library(crypto sha256_avx2.cpp
                    sha256_sse4.cpp
                    sha256_sse41.cpp
```
metalicjames, line 6: The SHA256 implementation comes with several optimised variants with significantly better performance. Before merging this PR we should check what the performance difference is when using this un-optimized SHA3 implementation. If the difference is significant, we should either hold off on this PR, or find similarly optimised implementations as SHA256.

metalicjames, line 6: The SHA256 implementation comes with several optimised variants with significantly better performance. Before merging this PR we should check what the performance difference is when using this un-optimized SHA3 implementation. If the difference is significant, we should either hold off on this PR, or find similarly optimised implementations as SHA256.

metalicjames (changes requested):
Concept ACK.

There are still a few things I think we should change/discuss before this PR is merged, though.

3rdparty/crypto/sha3.h
HalosGhost: Is there a clang-tidy/format check for this (I've not been able to find one)? C++ does not require and EOF newline (though C does), but if we want to enforce it, it'd be excellent if we could add that as a part of the CI.

3rdparty/crypto/CMakeLists.txt
HalosGhost: This is one of the questions I was hoping @madars would weigh in on. In #5, we talked about using different hash functions in different places. I was wondering if there was a place in this PR where we should continue using sha256 and keep both.

3rdparty/crypto/span.h
HalosGhost: cppreference [suggests](https://en.cppreference.com/w/cpp/compiler_support/20) that C++20 (and particularly, `std::span`) is rather widely-supported at this point. Should we just assume it's fair-game for us to jump to C++20 now?

3rdparty/crypto/CMakeLists.txt
metalicjames: The only place it might still be needed is for witness programs to maintain compatibility with `OP_SHA256` in Bitcoin. But in practice, atomic swaps tend to use `OP_HASH160` which uses RIPEMD-160. Since neither are implemented presently, I don't see the reason to keep the dead SHA256 code.

In terms of domain separation in #5, that's more about labelling pre-image payloads depending on the domain. For example, we might choose to always hash transaction IDs as `H("TXID" + payload)`, so that hashes for other purposes (with different labels) will never collide even with the same pre-image payload.

3rdparty/crypto/span.h
metalicjames: There are still several languages and especially standard library features missing from GCC/libstdc++ and Clang/libc++ so I don't think it's the right time to upgrade from C++17 just yet. Either way, this is not the appropriate PR to make such a change.

3rdparty/crypto/sha3.h
metalicjames: There is, but it looks like the `3rdparty` directory is excluded from that check: https://github.com/mit-dci/opencbdc-tx/blob/trunk/scripts/lint.sh#L19.

Let's add the newlines here and fix the lint script in a follow-up PR so the entire repo is consistent with EOF newlines.

HalosGhost:
@DeanWen, following the correction mentioned in the NEWS post, this PR needs to be rebased.

3rdparty/crypto/span.h
DeanWen: action item: make SHA3 interface closer to SH256

3rdparty/crypto/CMakeLists.txt
DeanWen: No action for now, @HalosGhost to do perf testing using existing tool

PR #78, opened 2022-03-15
Reduce Docker Image Size by 870MB

cdesch:
Changes
* Changed `Dockerfile` to a multi-stage build and pruned dependencies, reducing the image size from 987MB to 362MB. _(There is an opportunity to reduce the docker image build time to just the code base in the future)_
* Added `Deploy.Dockerfile` to include only essential binaries for slim container runtime, reducing the image size from 870MB to 117MB.

Note: **This change may break CI/CD pipelines if they depend on the default docker image.** Tests will still need the dependencies such as `lcov` and thus need to have the image built with a target stage of `builder`. The Readme.md was updated to reflect this (e.g. `docker build --target builder . -t opencbdc-tx`).

Image Size Differences:

Original Docker Image Size
```
➜  opencbdc-tx git:(trunk) ✗ docker images | grep opencbdc-tx
opencbdc-tx                                      latest                231a7edf0b88   About a minute ago   987MB
```
Reduced Docker Image Size, with no build dependencies, only source code and build products
```
➜  opencbdc-tx git:(trunk) ✗ docker images | grep opencbdc-tx
opencbdc-tx                                      latest                5110556d24a3   10 seconds ago   362MB
```
Further Reduced Docker Image Size by only including essential binaries. Build with a deployment Docker file `Deploy.Dockerfile`. Build the deployment image with: `docker build . -t opencbdc-tx -f Deploy.Dockerfile`
```
➜  opencbdc-tx git:(trunk) ✗ docker images | grep opencbdc-tx                       
opencbdc-tx                                      latest                e65ac6493f4b   7 minutes ago    117MB
```


HalosGhost:
Doesn't look like this breaks our current CI/CD pipeline at all (which is good news)! I'll be pulling your changes locally here to test this evening or tomorrow morning.

HalosGhost:
I love the idea of offering a really minimal image for people to leverage if they just want to run experiments, and I like the idea of having stages (rebuilding the full builder image seems to go quite quickly). My question is if it is possible to have the deploy image be named something different (`opendbdc-tx-minimal`?) so that you could build both the builder and deployment images without needing to rebuild the other if your use-case switches.

Is that easy to do or quite hard?

cdesch:
I wouldn't label it so much as `hard`, rather than doing what fits the project best.  Tagging the docker image with `opendbdc-tx-release` or `opendbdc-tx-minimal` for building that `Dockerfile` will work, although I think getting the right workflow will require more than one PR. There is a lot of flexibility between CI/CD, multiple build images, multistage images, and having different `docker-compose` files, with their own trade-offs. 

What I really like about the current docker build workflow and other similar open-source projects, is that the bar for entry for new contributors to get up and running is very low. You can clone the project, build with docker-compose, and you’re off to the races. I like keeping with that spirit as much as possible, although there are certainly areas where making the optimizations will improve build times, images sizes, and dev workflow. 

What I would suggest is taking stage what is in the `base` stage in the `Dockerfile` and move into a separate `base.Dockerfile` with a different CI workflow that publishes this image to a container registry on a schedule or as needed, and labeled with something like `opendbdc-tx-base`. The layers in this part of the image don’t change frequently enough to need to be rebuilt every time you want to compile the code. This base image will serve as an input for building the source for development and testing

Next, I'd split the workflow into two build products from different Dockerfiles, `DEBUG` and `RELEASE`. (Choosing `DEBUG` and `RELEASE` is subjective, but may provide flexibility/alignment with the `cmake` compiler flag `DCMAKE_BUILD_TYPE`, although I think that is a decision the project team might want to discuss). 
 - DEBUG Dockerfile, builds an image containing everything needed for building, testing, linting, and running the project. (Targeted for core project contributors)
 - RELEASE Dockerfile, builds an image containing only the build outputs and essential dependencies to run, making it the most lightweight. (Targeted for testing in an optimized environment)

Additionally, you might consider keeping a third image as the default `Dockerfile` that builds the binaries from source, includes source but removes build dependencies, taking the best from `Debug` and `Release`. This image would be for those new contributors following `readme.md` and getting running quickly with `docker-compose --file docker-compose-2pc.yml up`.

I pushed a branch that illustrates the first step here: https://github.com/cdesch/opencbdc-tx/tree/docker-images

README.md
```diff
@@ -145,9 +151,9 @@ In each of the below commands, you should pass `atomizer-compose.cfg` instead of
 
 Running Unit & Integration Tests
 
-1. Build the container
+1. Build the container. Target `builder` stage as the tests need the dependencies.
```
HalosGhost, line 154:
```suggestion
1. Build the container.
   Target `builder` stage as the tests need the dependencies.
```

This is a bit of a nitpick, but one-sentence-per-line makes storing documentation in-VCS dramatically nicer in the long run.

HalosGhost:
I think this looks awesome, and it behaves quite well for me locally (every step goes faster and generally has smaller image sizes). @metalicjames, would you mind taking a look at this as well? I'm still getting more familiar with docker and having a second set of reviewing eyes is helpful!

Deploy.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+    cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} .. && \
+    make -j$(nproc)
+
+# Create Deployment Image
+FROM $IMAGE_VERSION AS deploy
+
+WORKDIR /opt/tx-processor
+
+# Copy All of the ./build directory
+# COPY --from=builder  /opt/tx-processor/build ./build
```
metalicjames, line 70: Should this comment still be here?

Deploy.Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# Only copy essential binaries
+COPY --from=builder  /opt/tx-processor/build/src/uhs/twophase/sentinel_2pc/sentineld-2pc ./build/src/uhs/twophase/sentinel_2pc/sentineld-2pc
+COPY --from=builder  /opt/tx-processor/build/src/uhs/twophase/coordinator/coordinatord ./build/src/uhs/twophase/coordinator/coordinatord
+COPY --from=builder  /opt/tx-processor/build/src/uhs/twophase/locking_shard/locking-shardd ./build/src/uhs/twophase/locking_shard/locking-shardd
```
metalicjames, line 75: What about the binaries for the atomizer architecture?

metalicjames, line 75: What about the binaries for the atomizer architecture?

Dockerfile
```diff
@@ -1,19 +1,21 @@
-FROM ubuntu:20.04
+ARG IMAGE_VERSION="ubuntu:20.04"
```
metalicjames, line 1: Why is this Dockerfile still needed? Seems to be mostly a copy of `Deploy.Dockerfile`.

metalicjames, line 1: Why is this Dockerfile still needed? Seems to be mostly a copy of `Deploy.Dockerfile`.

metalicjames (changes requested):
Concept ACK.

We should try to minimise duplication if we can. There are now multiple docker files specifying roughly the same thing.

I think the Dockerfile can be further reduced in size and build time, I don't think we need `lcov` or `gmock`, or to build the tests/benchmarks for NuRaft.

HalosGhost:
@cdesch, following the correction mentioned in the NEWS post, this PR needs to be rebased.

PR #85, opened 2022-03-31
Sentinel compact transaction attestations

### LINKED ISSUE #84: Guard against malicious sentinels
In the current UHS-based data model, sentinels validate the transaction-local invariants of a full transaction, before converting the transaction to a compact representation which is the data structure processed by the backend (2PC or Atomizer). Since the data required to validate the full transaction is lost after being converted to a compact transaction, it would be trivial for a compromised sentinel to submit an invalid transaction for processing. This would allow the adversary to spend any UHS ID, and mint arbitrary UHS IDs representing any value (that an honest sentinel would later accept as a valid output). This is possible because the upstream component from the sentinels (shards in the atomizer architecture, coordinators in 2PC) blindly accepts compact transactions (deleting input UHS IDs, and creating output UHD IDs) and has no knowledge of the pre-image data. The current implementation does not restrict the origin of compact transactions either, so any entity with internal network access can send compact transactions directly to backend components, bypassing the sentinels.

It would be desirable to protect against up to `n` sentinels being compromised. `n` should be configurable based on the threat model and number of nodes in the system. As long as `<=n` sentinels are malicious, it should be impossible to settle an invalid transaction. Furthermore, upstream components should only accept compact transactions for processing from known sentinels.

metalicjames:
This PR implements a potential solution for #84.

Each sentinel is assigned a unique public/private key, defined in the system configuration. When sentinels validate a transaction, they sign the resulting compact transaction, if valid. There is a configuration parameter `attestation_threshold` which defines the number of distinct sentinels required to sign a compact transaction before upstream components will consider it valid. The sentinel that received the transaction first will distribute the transaction to additional sentinels which will also validate and return signatures. The originating sentinel gathers all of the signatures until `attestation_threshold` are available. Finally, the sentinel forwards the compact transaction and associated signatures to the upstream component for settlement. The upstream components (shards, atomizer cluster, coordinator cluster) check the signatures on the compact transaction, ensuring that `attestation_threshold` distinct, known sentinels have signed it. This solution guards against up to `attestation_threshold - 1` malicious sentinels and prevents upstream components from accepting compact transactions from unknown sentinels.


HalosGhost:
@metalicjames, following the correction in the NEWS post, this PR needs to be rebased.

PR #76, opened 2022-03-10
Add Docker Publishing Workflow

kylecrawshaw:
Adding a Github Actions Workflow to publish Docker images to Github Container Registry. Images will be published with the branch name, `latest` and commit sha only on the default branch. In this case every time a branch is merged into `trunk` there will be a new Docker image published with the tags `sha-<commit>`, `latest` and `trunk`.

This docker image can be pulled using the following:
```
docker pull ghcr.io/mit-dci/opencbdc-tx:latest
```
It can also be used in a downstream Docker image like this:
```
FROM ghcr.io/mit-dci/opencbdc-tx:latest
```

-  I've added a new file called `docker-compose-dev-2pc.yml` that only containers the image and build args necessary to make sure that docker compose will be build a local image.
- The changes to `docker-compose-2pc.yml` will ensure that docker compose uses a pre-built image from Github Container Registry.
    - The benefit of this change is that it will be significantly faster to get up and running with the system if you are not making any code changes.
- `README.md` has been updated with a new section for `Development` that explains how to run the system with both docker-compose files

Signed-off-by: Kyle Crawshaw <kyle.crawshaw@gmail.com>

PR #86, opened 2022-04-05
Adding a Github Actions Workflow to publish Docker images to Github C…

kylecrawshaw:
Adding a Github Actions Workflow to publish Docker images to Github Container Registry. Images will be published with the branch name, `latest` and commit sha only on the default branch. In this case every time a branch is merged into `trunk` there will be a new Docker image published with the tags `sha-<commit>`, `latest` and `trunk`.

This docker image can be pulled using the following:

```
docker pull ghcr.io/mit-dci/opencbdc-tx:latest
```

It can also be used in a downstream Docker image like this:

```
FROM ghcr.io/mit-dci/opencbdc-tx:latest
```

   -  I've added a new file called `docker-compose-prebuilt-2pc.yml` that only contains the image and build args necessary to make sure that docker compose will use a prebuilt image if this file is provided. Documentation has been updated with instructions.
        - The benefit of this change is that it will be significantly faster to get up and running with the system if you are not making any code changes.

Re-opening #76 
    Signed-off-by: Kyle Crawshaw <kyle.crawshaw@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,26 +62,39 @@ The easiest way to compile the code and run the system locally is using [Docker]
 
 Don't forget to run the docker daemon!
 
-## Build the container
-
-```terminal
-$ cd opencbdc-tx                 # change to the project directory
-$ sudo -s                        # open a root shell (needed for docker)
-# docker build . -t opencbdc-tx  # build the container
-```
-
 ## Launch the System
 
 **Note:** You will need to both run the system and interact with it; you can either use two shells, or you can add the `--detach` flag when launching the system (note that it will then remain running till you stop it, e.g., with `docker stop`).
 Additionally, you can start the atomizer architecture by passing `--file docker-compose-atomizer.yml` instead.
 
+_The commands below will build a new image every time that you run it.
+You can remove the `--build` flag after the image has been built to avoid rebuilding.
+To run the system with our pre-built image proceed to the [next section](#launch-the-system-with-a-pre-built-image) for the commands to run._
+
 1. Run the System
    ```terminal
-   # docker-compose --file docker-compose-2pc.yml up
+   # docker compose --file docker-compose-2pc.yml up --build
    ```
 1. Launch a container in which to run wallet commands (use `--network atomizer-network` instead of `--network 2pc-network` if using the atomizer architecture)
    ```terminal
    # docker run --network 2pc-network -ti opencbdc-tx /bin/bash
+   ```
+
+## Launch the System With a Pre-built Image
+
+We publish new docker images for all commits to `trunk`.
+You can find the images [in the Github Container Registry](https://github.com/mit-dci/opencbdc-tx/pkgs/container/opencbdc-tx).
+
+**Note:** You must use `docker compose` (not `docker-compose`) for this approach to work or you will need to pull the image manually `docker pull ghcr.io/mit-dci/opencbdc-tx`.
+Additionally, you can start the atomizer architecture by passing `--file docker-compose-atomizer.yml --file docker-compose-prebuilt-atomizer.yml` instead.
+
+1. Run the system
+    ```terminal
+    # docker compose --file docker-compose-2pc.yml --file docker-compose-prebuilt-2pc.yml up --no-build
+    ```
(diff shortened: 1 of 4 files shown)


HalosGhost:
Thank you @kylecrawshaw!

kylecrawshaw:
After some discussion on the side with @HalosGhost I've decided to take a different approach. Instead of changing the main docker-compose-2pc.yml file to use the prebuilt image I will create a different docker-compose file to be used in combination with the current one for people who just want to run the system without building.

This will ensure that the current behavior when running the system with docker compose remains the same, while also providing an easier and faster avenue for people who simply only want to run the code.

README.md
```diff
@@ -62,28 +62,34 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
-_The below configuration uses the pre-built image from Github Container Registry.
-To run and test your changes to the codebase you can follow the [Development section](#development)._
+_The commands below will build a new image every time that you run it.
+You can remove the `--build` flag after the image has been built to avoid rebuilding.
+To run the system with our pre-built image proceed to the [next section](#launch-the-system-with-a-pre-built-image) for the commands to run._
 
 1. Run the System
    ```terminal
-   # docker compose --file docker-compose-2pc.yml up
+   # docker compose --file docker-compose-2pc.yml up --build
```
kylecrawshaw, line 76: This is just removing a step in the process by using the flag instead of a separate run of `docker build`

README.md
```diff
@@ -62,28 +62,34 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
    ```terminal
-   # docker-compose --file docker-compose-2pc.yml up
+   # docker compose --file docker-compose-2pc.yml up --build
    ```
 1. Launch a container in which to run wallet commands (use `--network atomizer-network` instead of `--network 2pc-network` if using the atomizer architecture)
    ```terminal
    # docker run --network 2pc-network -ti opencbdc-tx /bin/bash
    ```
 
+## Launch the System With a Pre-built Image
```
HalosGhost, line 83: I think there's one more change here I'll request before merging. With this change, a user following the document will skip a step if they choose to pull images rather than build them (namely, they'll skip running a second container where they'd run the commands in the following steps).

I think there are two solutions to that and I would be fine with either:

* Separate out the second step in *Launch the System* into its own section, and combine this section with *Launch the System* as two different ways you could launch it
* Fold this section into *Launch the System* as a note underneath the first step as an alternative

I think the first option I mention is likely to be cleaner, but I'd be happy with either.

README.md
```diff
@@ -62,28 +62,34 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
+
+We publish new docker images for all commits to `trunk`.
+You can find the images [in the Github Container Registry](https://github.com/mit-dci/opencbdc-tx/pkgs/container/opencbdc-tx).
+
+**Note:** You must use `docker compose` instead of `docker-compose` for this approach to work or you will need to pull the docker image `docker pull ghcr.io/mit-dci/opencbdc-tx`.
```
```suggestion
**Note:** You must use `docker compose` (not `docker-compose`) for this approach to work or you will need to pull the image manually with `docker pull ghcr.io/mit-dci/opencbdc-tx`.
```

HalosGhost (changes requested):
One minor wording improvement (fine without it), and one change to the docs that I think needs to be done for clarity.

README.md
kylecrawshaw: Great call! I really appreciate the suggestions to improve documentation!

README.md
kylecrawshaw: Good catch. I'm going to duplicate that second step in the _Launch the System With a Pre-Built Image_ section since the command is slightly different given the different image that will be run.

HalosGhost:
@kylecrawshaw this looks excellent. Can you squash into a single commit, I'll let the CI run one more time and then merge?

kylecrawshaw:
> @kylecrawshaw this looks excellent. Can you squash into a single commit, I'll let the CI run one more time and then merge?

I have squashed all of my commits. This should be ready to go.

HalosGhost (approved):
Adds automatically-publishing a pre-built image for people to leverage if they're only doing testing (rather than development) and associated documentation for usage.

Looks good to me!

HalosGhost:
Merging.

PR #87, opened 2022-04-11
Sentinel compact transaction attestations

metalicjames:
This PR implements a potential solution for #84.

Each sentinel is assigned a unique public/private key, defined in the system configuration. When sentinels validate a transaction, they sign the resulting compact transaction, if valid. There is a configuration parameter `attestation_threshold` which defines the number of distinct sentinels required to sign a compact transaction before upstream components will consider it valid. The sentinel that received the transaction first will distribute the transaction to additional sentinels which will also validate and return signatures. The originating sentinel gathers all of the signatures until `attestation_threshold` are available. Finally, the sentinel forwards the compact transaction and associated signatures to the upstream component for settlement. The upstream components (shards, atomizer cluster, coordinator cluster) check the signatures on the compact transaction, ensuring that `attestation_threshold` distinct, known sentinels have signed it. This solution guards against up to `attestation_threshold - 1` malicious sentinels and prevents upstream components from accepting compact transactions from unknown sentinels.


davebryson:
@metalicjames @HalosGhost   Wow.  Nice work!  If I understand correctly, the transaction returned from the sentinel check now includes a list of signatures from the set of sentinels that successful validated the Tx.   If a quorum is reached, this should be a confirmed transaction that is guaranteed to be a valid transfer.  Couldn't this also be the basis for an offline transaction?  Either the originator or recipient could submit this to the transaction processor.

HalosGhost:
It looks like 12c2e58216a82f984090a6521d83cc81f953eb7f is unrelated to the sentinel attestations; should it be split out?

cc @metalicjames

src/uhs/twophase/locking_shard/interface.hpp
```diff
@@ -15,12 +16,8 @@
 namespace cbdc::locking_shard {
     /// Transaction type processed by locking shards.
     struct tx {
-        /// The TX ID of the transaction, if provided.
-        std::optional<hash_t> m_tx_id{};
-        /// Vector of input hashes for the shard to process as spent.
-        std::vector<hash_t> m_spending{};
-        /// Vector of output hashes to create on the shard.
-        std::vector<hash_t> m_creating{};
+        /// Compact TX.
```
HalosGhost, line 19: Is there a point in keeping this data structure if it's reduced to just contain a `compact_tx`? Is it just to avoid a greater delta for the PR?

src/uhs/transaction/validation.cpp
```diff
@@ -403,4 +403,24 @@ namespace cbdc::transaction::validation {
...
+        if(!std::all_of(tx.m_attestations.begin(),
+                        tx.m_attestations.end(),
+                        [&](const auto& att) {
+                            return pubkeys.find(att.first) != pubkeys.end()
+                                && tx.verify(secp_context.get(), att);
+                        })) {
+            return false;
+        }
+
+        return true;
```
HalosGhost:
```suggestion
        return std::all_of(
            tx.m_attestations.begin(),
            tx.m_attestations.end(),
            [&](const auto& att) {
                return pubkeys.find(att.first) != pubkeys.end()
                       && tx.verify(secp_context.get(), att);
            });
```

Any reason to avoid the immediate return?

HalosGhost (commented):
This took me a while to comb through, but I think it all looks pretty solid.

I do have a couple of small inline questions (none of them necessarily must be changed for merge).

src/uhs/twophase/locking_shard/interface.hpp
metalicjames: It avoids a greater delta, and leaves the structure available for future expansion for other tamper detection features.

davebryson:
@HalosGhost @metalicjames   Could this have an impact on the performance of the collecting sentinel as the number of sentinels grow?   What if you flipped the model?  Allow the sentinel to remain passive.  And have either the client or a "gateway" service contact and collect signatures from the sentinels.

HalosGhost:
> Could this have an impact on the performance of the collecting sentinel as the number of sentinels grow?

It will definitely have an impact on throughput as the threshold for required signatures grows (which I think is what you were referring to). Logically though, the more sentinels you have, the easier it will be to collect the required number of attestations (and some very minimal and preliminary testing suggests this may be correct).

Are you instead asking if it affects the performance of the specific sentinel that's collecting attestations? In some sense, that's not a huge concern because it's quite simple to spin up more sentinels as-needed.

> What if you flipped the model? Allow the sentinel to remain passive. And have either the client or a "gateway" service contact and collect signatures from the sentinels.

That's an interesting thought; to clarify, you're suggesting that either a client could make independent connections to sentinels to gather attestations, and then forward the attested transaction to a sentinel once the threshold has been gathered (or, adding another component to sit between the client and sentinel to do this work)?

If the client is to be responsible for it, then there'd also have to be some way for the system to tell the client what the attestation threshold is (right now, it's part of the component configuration, but it seems reasonable to imagine that you might (or need, depending on the threat model) that threshold to scale; and it also means that the client would then be responsible for making a lot of additional connections. Additionally, it might not be ideal to reveal that information from a security perspective.

Adding an additional system-side component to do this work would alleviate the revelation of that information to the client, and would decrease load on any specific sentinel; however, I'm not sure we know enough about this solution to say whether or not the affect on performance of a specific sentinel is worth investing effort into minimizing (given the ease with which you can simply spin up new sentinels).

src/uhs/atomizer/sentinel/controller.cpp
```diff
@@ -83,30 +109,81 @@ namespace cbdc::sentinel {
...
     void controller::send_transaction(const transaction::full_tx& tx) {
-        const auto compact_tx = cbdc::transaction::compact_tx(tx);
-        auto ctx_pkt = std::make_shared<cbdc::buffer>();
-        auto ctx_ser = cbdc::buffer_serializer(*ctx_pkt);
-        ctx_ser << compact_tx;
+        auto compact_tx = cbdc::transaction::compact_tx(tx);
+        auto attestation = compact_tx.sign(m_secp.get(), m_privkey);
+        compact_tx.m_attestations.insert(attestation);
+
+        gather_attestations(tx, std::move(compact_tx), {});
```
HalosGhost: Why make `gather_attestations()` actually broadcast the inputs? For the sake of clarity and composability, could it instead either return the set of gathered attestations to be included in the compact transaction, or graft them onto the transaction directly, but leave the actual input broadcasting to be in `send_transaction()` (or another extracted method)?

This isn't a deal-breaker, but the change surprised me.

src/uhs/twophase/sentinel_2pc/controller.hpp
```diff
@@ -49,20 +52,57 @@ namespace cbdc::sentinel_2pc {
...
+                                   const execute_result_callback_type& res_cb);
+
+        void
+        validate_result_handler(validate_result v_res,
+                                const transaction::full_tx& tx,
+                                execute_result_callback_type result_callback,
+                                transaction::compact_tx ctx,
+                                std::unordered_set<size_t> requested);
+
+        void gather_attestations(const transaction::full_tx& tx,
```
HalosGhost, line 80: Same question here as for the atomizer's new `gather_attestations()` (called out so we don't forget it if this should be changed).

HalosGhost (commented):
This looks like a solid implementation; it enables sentinels to validate a transaction (but not execute it), and sign off on its validity, then leverages that capability to allow requiring a threshold of N sentinel sign-offs before actually executing a transaction.

I only really have one question about the implementation, and it concerns clarity/maintainability rather than anything more serious. I made my question inline, but I suspect I'll be comfortable approving this for merge after it's answered.

src/uhs/atomizer/sentinel/controller.cpp
HalosGhost: cc @metalicjames

metalicjames: I moved the send logic into its own method for both the atomizer and 2PC sentinels.

metalicjames:
I split out #104.

src/uhs/atomizer/sentinel/controller.cpp
HalosGhost: This update still leaves the actual call to send the inputs inside `gather_attestations()` (which is the part that I find confusing)—on the surface, I wouldn't expect gathering attestations to require sending the inputs after the attestations have been gathered. Those feel like discrete steps to me (such that I'd expect the `send_compact_tx()` call to take place at the end of `send_transaction()` *after* `gather_attestations()` rather than those two concepts being coupled together).

Does that make sense? Is there an aspect to why gathering the attestations necessitates forwarding the inputs immediately that I'm missing?

metalicjames: Where else would you put it? `gather_attestations` is a series of callbacks to request the attestations from remote sentinels, which will respond asynchronously. Once we receive the last callback, we have enough attestations and are ready to send the transaction. `gather_attestations` (and also `send_transaction`) returns right away after sending the first attestation request, it does not block and return the attestation.

HalosGhost: That's quite clarifying; sorry for the confusion there. I suppose then it's less about anything than the name (it's not so much gathering attestations as it is actually sending the transaction, the first step of which is to gather the attestations). I'll call it resolved for now. Thank you!

HalosGhost (approved):
This looks to be excellent work, and a solid potential mitigation against sentinel compromise.

Looks good to me.

PR #92, opened 2022-04-23
get variant issue fixed including tests for default and non default constructibles

david8544:
get_variant will support variant of all default constructibles or all none default constructibles.


HalosGhost:
@davidmag854 it looks like this PR pulls in secp from the submodule in-tree (which is why it's 25k+ lines added, and 101 files changed). Can you fix that in your branch, and force-push?

PR #93, opened 2022-04-25
get_variant modified to work with variants which either hold 

david8544:
all default constructible alternatives or all none.

### DIFF
--- a/tests/unit/serialization/format_test.cpp
+++ b/tests/unit/serialization/format_test.cpp
@@ -420,8 +420,6 @@ TEST_F(format_test, get_variant_default_constructibles) {
     // empty set
     std::unordered_set<uint64_t> s0;
     std::variant<std::unordered_set<uint64_t>, uint64_t> v0{s0};
-    // how ser and deser are connected?
-    // my answer is the variant
     ser << v0;
     auto r0 = cbdc::get_variant<std::unordered_set<uint64_t>, uint64_t>(deser);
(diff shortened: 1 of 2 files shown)


tests/unit/serialization/format_test.cpp
```diff
@@ -414,3 +415,54 @@ TEST_F(format_test, malformed_unordered_sets_cannot_roundtrip) {
     // member)
     EXPECT_FALSE(deser);
 }
+
+TEST_F(format_test, get_variant_default_constructibles) {
+    // empty set
+    std::unordered_set<uint64_t> s0;
+    std::variant<std::unordered_set<uint64_t>, uint64_t> v0{s0};
+    // how ser and deser are connected?
```
HalosGhost: This comment seems like it might be left over from implementation (doesn't look to be documenting anything).

HalosGhost (changes requested):
All-in-all, this looks like a good improvement for one of our serialization helpers (`get_variant`) to make it a bit more flexible.

I only have one small thing I'd request to be changed, which is to remove a comment that I don't think provides much value.

I'd also love for @metalicjames to take a look at this. In particular, do you think it makes sense to try and find a way to support variants with a mixture of default-constructible and non-default-constructible alternatives?

tests/unit/serialization/format_test.cpp
david8544: removed

PR #96, opened 2022-04-29
adjust scripts, Cmake for OSX. Updates README

### LINKED ISSUE #94: Unable to build on OSX with bash scripts
### Affected Branch

trunk

### Basic Diagnostics

- [ ] The issue is reproducible in docker

### Description

There are several small issues with both `scripts/configure.sh` and `scripts/build.sh` that prevents building on OSX.  Here are few I've run across so far:

- brew installs leveldb but the header files are not available to the build.  There's also the issue of using different versions of leveldb between platforms: Linux build uses version 1.2.2.  brew uses the latest which is 1.2.3
- `configure.sh` uses `wget` in a few places but `wget` is not installed by default on OSX.  `curl` is available on OSX
- `build.sh` hardcodes a path to `gmake` which is different from the default location from brew
- copying `nuraft` headers, etc... to `/use/local/` requires `sudo`
- `git` submodules are not pulled in either script.

Note: I'm working on a fix for all this and will submit a pull request once I've worked through all the issues.

### Code of Conduct

davebryson:
Signed-off-by: Dave Bryson

Changes:
- Removed brew installs from configure.sh and added to build information in README.  Why? calling `sudo ./configure.sh` fails as brew can't be called with sudo
- Moved OSX specific CMAKE_FLAGS to CmakeFile.txt
- Added `build` information to README 

References: https://github.com/mit-dci/opencbdc-tx/issues/94

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,34 @@ For more information on how to contribute, please see our [Contribution Guide](d
 1. Clone the repository (including submodules)
     - `git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx`
 
+# Build
+
+Use these directions if you want to build the source on your machine.
+If you just want to run the system, see "Run the Code" below.
+
+## macOS
+Ensure your development environment is set correctly for clang:
+
+`sudo xcode-select -switch /Library/Developer/CommandLineTools`
+
+Or, if you've changed this in the past, you can reset to point to commandline tools with:
+
+`sudo xcode-select --reset`
+
+
+1. Install dependencies: `brew install leveldb llvm@11 googletest lcov make wget cmake`
+2. `./scripts/configure.sh`
+3. `./scripts/build.sh`
+
+Note: To run clang-tidy and clang-format (required by `lint.sh`), you must add them both to your path.
+
+Ex: `ln -s /usr/local/opt/llvm@11/bin/clang-tidy /usr/local/bin/clang-tidy`
+
+## Linux
+
+1. `./scripts/configure.sh`
+2. `./scripts/build.sh`
+
 # Run the Code
 
 The easiest way to compile the code and run the system locally is using [Docker](https://www.docker.com).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,6 @@ try_compile(USE_STDFS_LIB ${CMAKE_BINARY_DIR} "${CMAKE_SOURCE_DIR}/cmake-tests/t
 check_cxx_compiler_flag(-Wshadow-all W_SHADOW_ALL)
 check_cxx_compiler_flag(-Wnewline-eof W_NEWLINE_EOF)
 
-
 set(SSE41_CXXFLAGS "-msse4.1")
 set(AVX2_CXXFLAGS  "-mavx -mavx2")
 set(SHANI_CXXFLAGS "-msse4 -msha")
@@ -35,12 +34,11 @@ find_library(NURAFT_LIBRARY nuraft REQUIRED)
 message("Disabling Run Time Type Information (RTTI) features.")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
 
-include_directories(3rdparty 3rdparty/secp256k1/include)
+include_directories(3rdparty 3rdparty/secp256k1/include /usr/local/include /opt/homebrew/include)
 
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
     add_compile_options(-fprofile-arcs -ftest-coverage)
 endif()
-
 
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
     add_link_options(--coverage)
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -10,15 +10,14 @@ fi
 mkdir -p $BUILD_DIR
 cd $BUILD_DIR
 
-
-CPUS=1
 CMAKE_FLAGS=""
+CPUS=1
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
     CPUS=$(grep -c ^processor /proc/cpuinfo)
 elif [[ "$OSTYPE" == "darwin"* ]]; then
     CPUS=$(sysctl -n hw.ncpu)
     XCODE_CMDLINE_DIR=$(xcode-select -p)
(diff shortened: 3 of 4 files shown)


HalosGhost:
> calling `sudo ./configure.sh` fails as brew can't be called with sudo

Does it make sense, then, to instead inline the use of `sudo` for the linux commands (resulting in the correct invocation for both platforms being `./scripts.configure.sh` (no `sudo`))? This PR as-it-stands makes it so that there's an additional step for macOS (the `brew install` commands) beyond what's necessary for linux (which, in an ideal world, I think we should try to avoid).

Additionally, it looks like this PR includes some trailing whitespace in a few files (namely, `CMakeLists.txt`, and `README.md`) which is why the lint CI check failed.

davebryson:
@HalosGhost   Thanks for the feedback.   Resolving the linting issues now.   

As for `sudo`,  this is something I went back and forth on a couple of times.  I even attempted having `cmake` install leveldb and nuraft to eliminate the need for `sudo`.  But in the end, I settled on the path of least resistance (for now) so as not to cause a ripple effect with `ci.yml` etc... I'm open to suggestions...  There's definitely room for improvement.  For example, OSX is using leveldb from brew that's a different version than the one used by the linux build.  And is llvm@11 a brew dependency solely for `clang-tidy` and `clang-format`?

README.md
```diff
@@ -51,6 +51,27 @@ For more information on how to contribute, please see our [Contribution Guide](d
 1. Clone the repository (including submodules)
     - `git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx`
 
+# Build
+
+Use these directions if you want to build the source on your machine.
+If you just want to run the system, see "Run the Code" below.
+
+## OSX
```
jonwiggins: I think they switched to `macOS` in 2016: https://en.wikipedia.org/wiki/Macintosh_operating_systems#Releases

jonwiggins (commented):
This worked for me on an M1 Mac. Although at first after following the steps when I ran `./scripts/build.sh`, I got:
```
CMake Error at CMakeLists.txt:4 (project):
  The CMAKE_C_COMPILER:

    /Applications/Xcode.app/Contents/Developer/usr/bin/clang

  is not a full path to an existing compiler tool.
```
I had to run 
```
sudo xcode-select -switch /Library/Developer/CommandLineTools
```
And then everything worked.

HalosGhost:
> ```
> sudo xcode-select -switch /Library/Developer/CommandLineTools
> ```

This might make sense to include in the `README` in the macOS instructions. @davebryson what do you think?

davebryson:
@HalosGhost  Maybe `xcode-select --install` would be better to make sure it's installed.  I believe installing it will automatically set the path.

jonwiggins:
> @HalosGhost Maybe `xcode-select --install` would be better to make sure it's installed. I believe installing it will automatically set the path.

I had already installed it, it looks like the issue can be caused by:
1) Installing other programs which mess with the path: https://stackoverflow.com/questions/17980759/xcode-select-active-developer-directory-error
2) Updating macOS: https://developer.apple.com/forums/thread/666584

README.md
```diff
@@ -51,6 +51,29 @@ For more information on how to contribute, please see our [Contribution Guide](d
...
+
+Use these directions if you want to build the source on your machine.
+If you just want to run the system, see "Run the Code" below.
+
+## macOS
+Ensure your development environment is set correctly for clang:
+
+`sudo xcode-select -switch /Library/Developer/CommandLineTools`
+
+1. Install dependencies: `brew install leveldb llvm@11 googletest lcov make wget`
```
AlexRamRam:
```suggestion
1. Install dependencies: `brew install leveldb llvm@11 googletest lcov make wget cmake`
```

AlexRamRam (changes requested):
Add cmake to brew install list

README.md
```diff
@@ -51,6 +51,29 @@ For more information on how to contribute, please see our [Contribution Guide](d
 1. Clone the repository (including submodules)
     - `git clone --recurse-submodules https://github.com/mit-dci/opencbdc-tx`
 
+# Build
+
+Use these directions if you want to build the source on your machine.
+If you just want to run the system, see "Run the Code" below.
+
+## macOS
+Ensure your development environment is set correctly for clang:
```
AlexRamRam, line 60:
```suggestion
Ensure your development environment is set correctly for clang (to reset: sudo xcode-select --reset):
```
This tip might be useful for those using the full Xcode.app

AlexRamRam (changes requested):
My apologies for the multiple reviews (open-source/GH newbie ...)

README.md
HalosGhost: When mentioning code (or commands) in documentation, they should probably be enclosed in backticks for code formatting (`like so`).

README.md
davebryson: @AlexRamRam  Thanks for the feedback.  Did you run into an error without installing `cmake`?  Because it's also installed with `make`

AlexRamRam: I did. I did a separate "brew install cmake" to resolve it. Thanks.

HalosGhost:
@davebryson I took another look, and Alex has had occasion to do another clean build using your changes, and they all look set. Can you squash a single commit, and then I shall merge it?

HalosGhost:
@davebryson it looks like you merged back in from trunk, but didn't squash (resulting in 10 commits to pull rather than 1).

davebryson:
@HalosGhost  Yea I pulled to merge the latest by accident. But haven't done the squash yet.  Sorry for the mis-leading message

davebryson:
@HalosGhost Squash is done.  I had to back out the latest merges from dci/trunk.

HalosGhost:
@davebryson these changes look solid and I'm ready to merge but I just wanted to double-check one thing with you. You've included a DCO sign-off, so as far as we're concerned, this is set from an IP perspective. However, because your sign-off isn't at the end of the commit message, `git` isn't aware of it:

```console
$ git interpret-trailers --parse <this-commit-as-a-patch-file>
$
```

This might mean that you're not credited appropriately for the commit. If you're worried about that, you can amend the commit message so that the sign-off is in its own paragraph at the end of the commit message (contiguously with any other trailers):

```
Support for OSX Build

Updated scripts, cmake, and README for building on OSX

Signed-off-by: Dave Bryson <daveb@miceda.org>
```

Which `git` understands correctly:

```console
$ git interpret-trailers --parse <the-updated-commit-patch>
Signed-off-by: Dave Bryson <daveb@miceda.org>
```

Feel free to push the amended commit if you want to fix it; otherwise, ping me and I'll merge!

HalosGhost:
Conferred offline; @davebryson asked for it to be merged.

PR #99, opened 2022-05-10
Check the watchtower is syncing with the atomizer

metalicjames:
If the watchtower fails to sync blocks from the atomizer in the end-to-end integration test, this was previously a silent failure. The tests would still fail but due to confusing side effects like transactions not being confirmed. This commit updates the test to explicitly check the watchtower is syncing blocks and that the mint transactions confirm. It also adds some log messages to components that were useful while investigating the issue that led to this PR.

### DIFF
--- a/src/uhs/atomizer/atomizer/controller.cpp
+++ b/src/uhs/atomizer/atomizer/controller.cpp
@@ -104,6 +104,8 @@ namespace cbdc::atomizer {
             });
         }
 
+        m_logger->info("Atomizer started...");
+
         return true;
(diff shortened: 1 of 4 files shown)


HalosGhost:
@metalicjames looks like there's a small merge conflict to be resolved.

metalicjames:
@HalosGhost should be good to review/merge now

HalosGhost (approved):
Simple logging improvements and a stronger check in the atomizer E2E test. Looks good to me.

PR #103, opened 2022-05-13
Supply auditing by storing values in the UHS

### LINKED ISSUE #100: Enable supply auditing by storing values in the UHS
If we include the value of the output in the preimage used to create the UHS ID, and store the values along-side the IDs in the UHS, then we've effectively made a hash-based commitment to the value which can be verified after-the-fact.

This makes it easy to keep all our invariants, and elegantly adds a simple mechanism to verify that the full monetary supply continues to hold the total-minted-value.

metalicjames:
This PR modifies the data model of both UHS-based architectures to contain the UTXO value corresponding to each UHS ID. It contains a tool to gather audit logs from each of the shards and aggregate them to acquire a consistent view of the supply at a particular time. In 2PC, we attach "epochs" to transactions allowing us to get a consistent snapshot across shards. In the Atomizer architecture, we use the blocks for coordination.

Implements #100.


HalosGhost:
@metalicjames looks like it failed the lint check solely due to use of `ul` rather than `UL` for some literals.

HalosGhost:
closing in-favor of #224

PR #104, opened 2022-05-17
Balance load across sentinels from load generators

metalicjames:
Pulled out from #87. Implements load balancing across multiple sentinels from load generators. Previously load generators were paired one-to-one with a sentinel. This PR makes load generators send transactions to all sentinels randomly selecting between them. This improves performance when a single sentinel is unable to absorb the load from a load generator.


HalosGhost (approved):
Thank you for splitting this out! This change to clear (well, replace) a todo for load-balancing across the sentinels looks solid and self-contained.

PR #105, opened 2022-05-18
Always start tx notify thread in atomizer

metalicjames:
Previously the transaction notification thread in the atomizer was only started when `m_opts.m_batch_size > 1`. Since this thread is always needed (notifications are always batched in the current code), remove that condition.


HalosGhost (approved):
A very self-contained and self-explanatory change; this ensures we always start the tx-notify thread in the atomizer.

Looks good to me!

PR #106, opened 2022-05-18
Wait to process incoming requests until syncronized with the atomizer

metalicjames:
This PR prevents the atomizer shard from listening on its server endpoint until it has received at least one block from the atomizer. This ensures that the shard will not be considered "up" until it has successfully connected to an atomizer. Without being synchronized with the atomizer, the shard cannot process any requests, so this is desirable behavior.

### DIFF
--- a/src/uhs/atomizer/shard/controller.cpp
+++ b/src/uhs/atomizer/shard/controller.cpp
@@ -63,6 +63,19 @@ namespace cbdc::shard {
             return atomizer_handler(std::forward<decltype(pkt)>(pkt));
         });
 
+        constexpr auto max_wait = 3;
+        for(size_t i = 0; i < max_wait && m_shard.best_block_height() < 1;
+            i++) {
+            m_logger->info("Waiting to sync with atomizer");
+            constexpr auto wait_time = std::chrono::seconds(1);
+            std::this_thread::sleep_for(wait_time);
+        }
+
+        if(m_shard.best_block_height() < 1) {
+            m_logger->warn(
+                "Shard still not syncronized with atomizer, starting anyway");
+        }
+
         auto ss = m_shard_network.start_server(
             m_opts.m_shard_endpoints[m_shard_id],
             [&](auto&& pkt) {
--- a/src/uhs/atomizer/shard/controller.cpp
+++ b/src/uhs/atomizer/shard/controller.cpp
@@ -136,7 +136,14 @@ namespace cbdc::shard {
         // If the block is not contiguous, catch up by requesting
         // blocks from the archiver.
         while(!m_shard.digest_block(blk)) {
-            m_logger->warn("Block", blk.m_height, "not contiguous.");
+            m_logger->warn("Block",
+                           blk.m_height,
+                           "not contiguous with previous block",
+                           m_shard.best_block_height());
+
+            if(blk.m_height <= m_shard.best_block_height()) {
+                break;
+            }
 
             // Attempt to catch up to the latest block
             for(uint64_t i = m_shard.best_block_height() + 1; i < blk.m_height;


src/uhs/atomizer/shard/controller.cpp
```diff
@@ -63,6 +63,19 @@ namespace cbdc::shard {
...
 
+        constexpr auto max_wait = 3;
+        for(size_t i = 0; i < max_wait && m_shard.best_block_height() < 1;
+            i++) {
+            m_logger->info("Waiting to sync with atomizer");
+            constexpr auto wait_time = std::chrono::seconds(1);
+            std::this_thread::sleep_for(wait_time);
+        }
+
+        if(m_shard.best_block_height() < 1) {
```
HalosGhost, line 74: This is not enforcing a wait-till-sync behavior; this is adding a sleep hoping it will be long enough for that behavior to have happened, and then continuing on after a time-out if not.

Because the shard cannot process requests until synchronized, it makes sense to me to enforce that synchronization before listening; but I would expect that to take one of two forms:

1. wait indefinitely until we have synched
2. wait until we have synched or a timeout is reached, at which point we fail and exit out

As it stands, I'm not clear on what benefit this code offers.

metalicjames: > wait until we have synched or a timeout is reached, at which point we fail and exit out

That is what this code does, as far as I understand it. The timeout is there for the test suite to work.

HalosGhost: So, to clarify, this addition isn't enforcing the behavior of failing out (which will be hit by `ss` not having a value on now-line 79). Rather, this is adding the timeout to offer an alternative to immediately failing out if the synchronization hasn't already happened?

metalicjames: Before this change the shard server (listening for sentinel requests) would start listening even if it was unable to connect to any atomizers. It would then start processing requests, all of which would silently fail, leading to needless errors /retries at the start of a test run. This change delays the start until we're definitely connected to the atomizer cluster (indicated by receiving at least one block), or a timeout elapses.

HalosGhost: That is closer to what I had initially thought then. So, after the timeout elapses, the shard server doesn't fail out and stop; it will go into that same silent failure mode, right?

I think by adding a retry-timeout, we would be masking the problem. I think it makes more sense to have the server fail out and stop running after failure to synchronize. Adding the timeout in this way makes it more likely slower systems will fail when faster systems operate properly, which is something I think we should try to avoid.

metalicjames: Indeed. Eventually the shard might connect though.

If I remove the timeout, the integration tests fail. I think we'd need to rearchitect our tests to handle components starting in an asynchronous manner.

We could choose not to merge this change. It's not required to pass the CI and we haven't needed up until now. I made the change in response to some challenges @AlexRamRam was facing.

HalosGhost: Just to double-check I'm understanding: when you remove the timeout (i.e., it will wait indefinitely until it synchronizes), that results in the tests failing? That's not what I would expect (I'd have expected that to result in much more consistent behavior for testing).

metalicjames: Indeed, if we remove the timeout the integration tests fail because the shard controller never returns from `init()`. Similarly, if we leave in the timeout but make failure to connect an error, `init()` returns false and the shard fails to initialize. This is because of the startup order in our integration tests: https://github.com/mit-dci/opencbdc-tx/blob/trunk/tests/integration/shard_integration_test.cpp#L25. `init()` must return before the test framework can connect to it to simulate messages.


e.g. making the timeout an error
```
[ RUN      ] shard_integration_test.error_initial_sync
[2022-06-03 11:18:03.521] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:18:04.524] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:18:05.528] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:18:06.533] [ERROR] Shard still not syncronized with atomizer
/build/tests/integration/shard_integration_test.cpp:24: Failure
Value of: m_ctl->init()
  Actual: false
Expected: true
[  FAILED  ] shard_integration_test.error_initial_sync (3148 ms)
```

e.g. removing the timeout
```
[ RUN      ] shard_integration_test.error_non_existant_input
[2022-06-03 11:26:37.088] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:38.093] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:39.098] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:40.100] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:41.102] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:42.105] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:43.109] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:44.112] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:45.118] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:46.118] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:47.120] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:48.123] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:49.128] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:50.133] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:51.136] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:52.139] [INFO ] Waiting to sync with atomizer
[2022-06-03 11:26:53.142] [INFO ] Waiting to sync with atomizer
...
```

HalosGhost: I see. That is an unfortunate coincidence of less-than-ideal behavior. I'm not against merging this, but I'm a little concerned about adding workarounds for the test-suite's fragility into the code.

HalosGhost: This is a bigger issue than can be reasonably dealt with in this PR. I've split out the larger issue to #127. I will think some more on whether or not it is sensible to merge in this work-around in the meantime (recognizing it's not needed to pass CI).

src/uhs/atomizer/shard/controller.cpp
```diff
@@ -63,6 +63,19 @@ namespace cbdc::shard {
             return atomizer_handler(std::forward<decltype(pkt)>(pkt));
         });
 
+        constexpr auto max_wait = 3;
+        for(size_t i = 0; i < max_wait && m_shard.best_block_height() < 1;
+            i++) {
+            m_logger->info("Waiting to sync with atomizer");
```
HalosGhost, line 69: @metalicjames I have one last question on this one (and then, I think I'm ready for it to be merged): would it be simpler / more stable / more intuitive to use `condition_variable`s and `wait_for`? The result should be about the same, but is pretty easy to generalize.

metalicjames: Maybe, but we don't have access to the internal state of the shard here, so I'm not sure how that could be implemented without changing the shard interface?

HalosGhost (approved):
Sorry this took so long for me to revisit, but this is a very reasonable reliability improvement for atomizer shards. Looks good to me.

PR #107, opened 2022-05-23
Update doxygen github action for gh-pages

jallen-frb:
- Updates the doxygen github action to use the new edge version
- Fix allows pushes to gh-pages


HalosGhost (approved):
Very simple solution to get the documentation building (and should minimize this problem in the future by pointing to `edge`).

PR #109, opened 2022-05-26
Test atomizer raft node recovery via snapshot, and small test-related fixes

AlexRamRam:


scripts/test.sh
```diff
@@ -14,7 +14,7 @@ run_test_suite () {
     LOCATION=$2
     rm -rf $LOCATION
     mkdir -p $LOCATION
-    find . \( -name '*.gcno' -or -name '*.gcda' \) -and -not -path '*coverage*' -exec cp --parents \{\} "$LOCATION" \;
+    find . \( -name '*.gcno' -or -name '*.gcda' \) -and -not -path '*coverage*' -exec rsync -R \{\} "$LOCATION" \;
```
HalosGhost, line 17: I think this particular change is not a good idea. We're currently using a non-portable extension of `cp` (which is what you ran into), but your replacement is also non-portable. I'd recommend we instead leverage POSIX-specified variants to accomplish this.

1. `cp`'s `--parents` option does the same thing as `mkdir`'s `-p` (which *is* POSIX-specified); so we could do this:
   ```suggestion
       find . \( -name '*.gcno' -or -name '*.gcda' \) -and -not -path '*coverage*' -exec mkdir -p $(dirname \{\}) && cp \{\} "$LOCATION/\{\}" \;
   ```

That looks terrible, so I'd love to not do that. 

2. Instead, we could use `install`'s `-D` which works much more closely to `cp --parents` and is also POSIX-specified:
   ```suggestion
       find . \( -name '*.gcno' -or -name '*.gcda' \) -and -not -path '*coverage*' -exec install -D -m644 \{\} "$LOCATION/\{\}" \;
   ```
   `-m644` specifies the files should have `rw-r--r--` permissions which is, I believe, what we would want for all `gcno`/`gcda` files

@metalicjames thoughts?

src/uhs/atomizer/atomizer/state_machine.cpp
```diff
@@ -203,7 +203,6 @@ namespace cbdc::atomizer {
     void state_machine::create_snapshot(
         nuraft::snapshot& s,
         nuraft::async_result<bool>::handler_type& when_done) {
-        assert(s.get_last_log_idx() == last_commit_index());
```
HalosGhost: Why is this removed? I don't see any additions to the test that look like they test this holds.

tests/integration/atomizer_end_to_end_test.cpp
```diff
@@ -207,25 +213,17 @@ TEST_F(atomizer_end_to_end_test, double_spend) {
                                                       ctx.m_uhs_outputs[1],
                                                   }}}});
 
-    auto want = cbdc::watchtower::status_request_check_success{
```
HalosGhost, line 215: Is breaking this (and below) out just to make it more clear what is being asserted?

tests/integration/two_phase_end_to_end_test.cpp
```diff
@@ -35,14 +35,14 @@ class two_phase_end_to_end_test : public ::testing::Test {
                                                                m_opts,
                                                                m_logger);
 
-        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
HalosGhost: Am I remembering right that you made this change to just ensure the tests passed locally for you? If so, I think that change probably doesn't belong in this PR (and frankly, probably signals that we need to make more significant changes so that the test's sleeping is less fragile).

HalosGhost (changes requested):
The core of this PR (the actual test itself) looks pretty solid to me. However, I have a few requests and questions I've made inline to address before merge.

src/uhs/atomizer/atomizer/state_machine.cpp
AlexRamRam: I believe this `assert` to be a bug. However, the a different assert may make sense depending on the intent of the original author. @metalicjames - please comment.

tests/integration/atomizer_end_to_end_test.cpp
AlexRamRam: No. The original check asserts a block height. The block height can only be estimated as empty blocks may be added at the prescribed interval. Hence the goal is the maintain the other checks while removing the check for block height.

tests/integration/two_phase_end_to_end_test.cpp
AlexRamRam: Spot on - More changes are needed to properly address the timing issue that appears to exist. I will revert this code change, but add some comments.

src/uhs/atomizer/atomizer/state_machine.cpp
metalicjames: The assertion is there because it's an assumption the rest of the method makes. `create_snapshot` only ever creates a snapshot as of `last_commit_index()`. Thus, if NuRaft puts something else in `s.get_last_log_idx()`, the generated snapshot will be for the incorrect log index. It wasn't super clear, but I assumed NuRaft would only ever call this method with `s.get_last_log_idx() == last_commit_index()`. If that's not true, we need to edit this method to support generating a snapshot at any log index rather than the current one.

scripts/test.sh
metalicjames: What was wrong with the script previously? This code hasn't changed in quite a while.

AlexRamRam: `cp --parents` is not available on MacOS

tests/integration/atomizer_end_to_end_test.cpp
AlexRamRam: New Issue created: https://github.com/mit-dci/opencbdc-tx/issues/119

tests/integration/two_phase_end_to_end_test.cpp
AlexRamRam: Changes reverted, new Issue created: https://github.com/mit-dci/opencbdc-tx/issues/119

scripts/test.sh
HalosGhost: In talking with @AlexRamRam about this a bit more offline; I think it probably makes sense for us to switch to `rsync` and add it to `./scripts/configure.sh` to-be-installed.

src/uhs/atomizer/atomizer/state_machine.cpp
AlexRamRam: @metalicjames : please see #125

HalosGhost:
If we can split out the test-script fixes (perhaps to #122?), I think it's ready-to-merge; good to actually test atomizer recovery via-snapshot.

HalosGhost (approved):
I think your changes are solid, and I'm ready to merge this.

Issue #98, opened 2022-05-04
Secure Minting


davebryson:
### Question

How can the system support secure minting of currency?

### Benefit

A secure minting solution will ensure that only pre-authorized, designated, wallets can mint new currency

### Proposed Solution

For an initial solution we will add the ability for sentinels to recognize and validate mint transactions.  This will require:

- Adding configuration information to include a list of public keys authorized to sign mint transaction outputs
- Changing the validation logic.  Instead of rejecting a transaction with zero inputs.  This will signal the logic to use the configured (mint) public keys to validate signatures
- Update the wallet as needed to support the above. For example, the wallet will need to identify and use the pre-authorized public key to sign mint transactions

### Possible Difficulties

The proposed solution is a starting point - it may not be the final (best) solution.  More research will be needed to ensure minting is as secure as possible. 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

davebryson:
@HalosGhost  You can assign this to me.  I've started working on it.

davebryson:
@HalosGhost  Working through an implementation of secure minting, I've ran into an issue with the Atomizer sentinel.  Minting transactions do not have Tx inputs.  However, the Atomizer sentinel expects Tx inputs in order to forward the transaction: https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/uhs/atomizer/sentinel/controller.cpp#L97.   It *appears* the atomizer sentinel uses Tx inputs to validate a shard id.    

The logic I've added for secure minting performs specific minting validation if a transaction has no inputs.  Otherwise a non-mint transaction, with no inputs, will be rejected.   So technically valid non-mint transactions will work with the current sentinel logic.   

However, if I adjust the logic to support a mint transaction (no inputs), what's the best way to determine which `pid` ( https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/uhs/atomizer/sentinel/controller.cpp#L116 ) the transaction should be sent to?  Are there any other potential issues with a compact transaction having no inputs in the Atomizer?

**UPDATE**: On further investigation, it appears the shard also checks for inputs:  https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/uhs/atomizer/shard/shard.cpp#L123

HalosGhost:
@davebryson I just want to double-check, it sounds like you managed to find an answer to the primary question you had; is there anything else tripping you up or blocking you?

davebryson:
[...]

PR #110, opened 2022-05-26
Add secure minting

davebryson:
This PR addresses issue [#98](https://github.com/mit-dci/opencbdc-tx/issues/98) adding support for secure minting.  Mint transactions now run through the sentinel and validation.  Minters are identified via pre-approved key pairs used during sentinel validation to determine who is allowed to execute a mint transaction.  Public keys associated with approved minters are added to the `cfg` file.

For testing and demonstrations a pre-calculated minter key pair is available.  Existing `cfg` files have been updated with the information.

A mint transaction is identified as having no inputs and 1 or more outputs.   Validation has been adjusted to support this to ensure only valid minting transactions are forwarded from the sentinel.  This required a change in the Atomizer as it will no longer reject a transaction that has no inputs (as it did previously).  

PR [#96](https://github.com/mit-dci/opencbdc-tx/pull/96) should be applied before this PR


HalosGhost:
@davebryson looks like there's a small merge conflict (and some linting issues); but a general rebase is probably a good idea now that the macOS script updates have been merged in.

davebryson:
@HalosGhost  Rebased.  But still need to address cognitive complexity issue (clang-tidy) in `client-cli.cpp`

davebryson:
@HalosGhost   clang-tidy is failing due to the number of `if/else` clauses in `client-cli.cpp main()` exceeding the default threshold of 25 (there are currently 26) for the `readability-function-cognitive-complexity.Threshold` setting. 

Any objections to increasing the threshold setting in `.clang-tidy` to provide a little room for growth in the command line client?

HalosGhost:
I think a longer-term solution would be to factor out a basic parsing mechanism so that we can have less of an `if`/`else` soup (ideally allowing use of a function-pointer map, or `switch`/`case`); but that's definitely a larger refactoring effort. I don't have any immediate problem raising the threshold. @metalicjames is there a middle-ground that you think we should pursue instead?

metalicjames:
I would prefer that we move the command dispatch code to its own function than increase the cognitive complexity limit.

davebryson:
@HalosGhost   Not sure what the issue is with the CI failing on unit/integration tests.   Note commit `9fa789e` above passed all tests.  Last commit (that failed tests) made no changes that would effect tests.   Additionally I've tested locally several times and all unit/integration tests are passing.

HalosGhost:
Seems it's the atomizer end-to-end tests:

```
/home/runner/work/opencbdc-tx/opencbdc-tx/tests/integration/atomizer_end_to_end_test.cpp:58: Failure
Value of: m_ctl_watchtower->get_block_height() > 0
  Actual: false
Expected: true
scripts/test.sh: line 9:  8480 Aborted                 (core dumped) $PWD/$1
```

I will try to pull the changes and test locally today to see what's going wrong.

HalosGhost:
@davebryson your suspicions were correct; timing issue during the run; rerunning it passed all checks in a breeze. I'll dig in and begin review in the next few days!

src/uhs/atomizer/sentinel/controller.cpp
```diff
@@ -86,7 +86,7 @@ namespace cbdc::sentinel {
 
     auto controller::execute_transaction(transaction::full_tx tx)
         -> std::optional<cbdc::sentinel::execute_response> {
-        const auto res = transaction::validation::check_tx(tx);
+        const auto res = transaction::validation::check_tx(tx, m_opts);
```
HalosGhost: I definitely think we should avoid passing the full config through to `check_tx`; I think passing in the vector of blessed keys is probably a lot cleaner.

src/uhs/atomizer/sentinel/controller.cpp
```diff
@@ -167,7 +167,17 @@ namespace cbdc::sentinel {
             return;
         }
 
-        send_compact_tx(ctx);
+        // If the tx has no inputs, it's a mint. Send it directly to one of the
```
HalosGhost, line 175: By actually including minting machinery, we should now remove this work-around (minting transactions should now be considered a special-case of normal transactions, and should go through the same workflow) unless there's a compelling reason to avoid the coordinator/atomizer.

src/uhs/atomizer/shard/shard.cpp
```diff
@@ -120,10 +120,13 @@ namespace cbdc::shard {
...
 
+        atomizer::tx_notify_request msg;
+
+        // If the tx has no inputs, it's a mint.
         if(tx.m_inputs.empty()) {
-            return cbdc::watchtower::tx_error{
-                tx.m_id,
-                cbdc::watchtower::tx_error_inputs_dne{{}}};
+            msg.m_tx = std::move(tx);
+            msg.m_block_height = best_block_height();
```
HalosGhost: now-line 164 is as follows:

```
msg.m_block_height = snp_height;
```

Should we be doing that same thing here rather than calling `best_block_height()`?

src/uhs/client/client-cli.cpp
```diff
@@ -204,9 +246,46 @@ auto confirmtx_command(cbdc::client& client,
...
+    }
+    return result;
+}
+
 // LCOV_EXCL_START
 auto main(int argc, char** argv) -> int {
     auto args = cbdc::config::get_args(argc, argv);
+
+    // Create a demo wallet
+    if(args.size() == 3 && args[2] == "demowallet") {
```
HalosGhost: This should probably be moved to live under `dispatch_command()` as well. That probably means reformatting how `generate_demo_minter_wallet()` works right now (namely, so that it follows the expected argument set).

HalosGhost: Honestly, I'd also be okay with modifying our routines so that all new wallets created have the minting key as their first key (so that all wallets can actually still mint, but we are checking that minting transactions still come from the blessed key) and then avoiding an additional command.

Thoughts?

src/uhs/transaction/validation.cpp
```diff
@@ -419,4 +455,113 @@ namespace cbdc::transaction::validation {
                                    && tx.verify(secp_context.get(), att);
                            });
     }
+
+    auto check_mint_witness(const cbdc::transaction::full_tx& tx,
+                            size_t idx,
+                            const cbdc::config::options& options)
```
HalosGhost: Here, I also think we should avoid passing in configurations. When possible, let's minimize the amount of irrelevant data we pass to the `check_*` functions.

src/uhs/transaction/wallet.cpp
```diff
@@ -121,6 +154,35 @@ namespace cbdc {
         return ret;
     }
 
+    auto transaction::wallet::generate_minter_key() -> pubkey_t {
+        if(m_pubkeys.empty()) {
+            // new wallet. create the minter key
+            return generate_key();
```
HalosGhost: This looks a little odd to me (it seems like I'm missing something). If there are no public keys, `generate_minter_key()` will generate a new keypair (that seems reasonable), but the comment suggests that calling `generate_key()` will ensure that `m_pubkeys[0]` holds the minting key. As far as I can tell, `generate_key()` will return a new key created rather than the minting key, and it doesn't include any calls to functions that would generate the minting key or ensure that `m_pubkeys[0]` is the minting key.

src/uhs/transaction/wallet.cpp
```diff
@@ -121,6 +154,35 @@ namespace cbdc {
...
+        }
+        // first key is always the minter key
```
HalosGhost: As a design-note, I think this is kind of limiting (recognizing this is the first run at secure minting, perhaps that's not that big of a deal); for example, this makes it impossible for a wallet to hold more than one minting-key.

A more general solution might be to offer an additional mapping that takes a public key and returns whether or not that key has been blessed for minting. This could be in the form of a simple map (`std::unordered_map<pubkey_t, bool>`), to more complex things that would need to be tailor-made (e.g., [sparse arrays](https://research.swtch.com/sparse)).

tests/unit/validation_test.cpp
```diff
@@ -336,3 +340,71 @@ TEST_F(WalletTxValidationTest, sign_verify_compact) {
     ASSERT_FALSE(
         cbdc::transaction::validation::check_attestations(ctx, m_pubkeys, 2));
 }
+
```
HalosGhost: I think this should be split into its own file.

tests/unit/validation_test.cpp
```diff
@@ -336,3 +340,71 @@ TEST_F(WalletTxValidationTest, sign_verify_compact) {
...
+        const auto minter_pub = m_minter.generate_minter_key();
+        m_opts.m_minter_pubkeys.insert(minter_pub);
+    }
+
+    cbdc::transaction::wallet m_minter{};
+    cbdc::transaction::wallet m_not_minter{};
+    cbdc::config::options m_opts{};
+};
+
+// To Test:
```
HalosGhost: Looks like this list is no longer needed (all these tests are present below).

src/uhs/client/client-cli.cpp
HalosGhost: Having read further down, it actually seems like this is planned to be the case anyway. If so, is `generate_demo_minter_wallet()` necessary?

HalosGhost (changes requested):
Sorry it has taken me a little while to review this. I think there's a lot of great work here (and a solid step towards having some rough means of secure minting).

I do have a few questions to make sure I'm understanding the flow right, and a few requests for changes that I've made inline.

Please feel free to ask for clarification!

src/uhs/atomizer/sentinel/controller.cpp
davebryson: @HalosGhost  The reason for this is `send_compact` checks for inputs in 2 different places (lines 195 & 202) . Which means a mint will fail in the call requiring some other work around.  Overall, I tried to minimize altering other logic as much as possible.

src/uhs/atomizer/sentinel/controller.cpp
davebryson: @HalosGhost   Thought about this initially.  Decided on passing the ref to `opt` as I suspect future functionality may need access to `opts` as well.  But I can change if you like.

src/uhs/atomizer/shard/shard.cpp
davebryson: I think technically they are the same value at the point of the call,  but I'll change for consistency

src/uhs/transaction/wallet.cpp
davebryson: @HalosGhost  `m_pubkey[0]` is always the minting key.  If there are no `pubkeys` then `generate_key()` creates one and inserts it at `m_pubkeys[0]` (this is the minting key), otherwise, we use the existing minting key at `m_pubkey[0]`.

src/uhs/transaction/wallet.cpp
davebryson: @HalosGhost  Agree.  Just wanted to keep it simple as possible on the first pass.

tests/unit/validation_test.cpp
davebryson: Do you mean the `MinterValidationTest`?  I actually had it as a separate file initially.  I can change it back to that.

src/uhs/client/client-cli.cpp
davebryson: Any wallet created will have the first key marked as the minter key. 

`generate_demo_minter_wallet()` is used to denote that you're using a pre-calculated, deterministic minter key.  This key is used to provide a fixed, known key for tests and demos.  It corresponds to the minter public key you see in the current `cfg` files.

davebryson: Overall, I'd say the `cli` could probably use a little polishing to make it a little more user-friendly and extensible.   For example, it'd be nice to not have to copy and past return values for the `import_inputs` call.  I may look in to that a bit more when this is wrapped up.

davebryson:
@HalosGhost  Latest commit addresses requested changes

tests/unit/validation_test.cpp
HalosGhost: Yeah, I think it makes sense to be standalone.

src/uhs/atomizer/sentinel/controller.cpp
HalosGhost: Without specific functionality in mind for relying on the configuration, I'd rather default to [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

src/uhs/client/client.hpp
```diff
@@ -237,6 +251,8 @@ namespace cbdc {
         /// \brief Sends the given minting transaction to a service that will
         ///        accept and process it.
         ///
+        /// TODO:  Remove. No longer needed
```
HalosGhost: Is the point of leaving this just to minimize the change in this PR? (it's otherwise leaving specifically-dead code in the repo.)

src/uhs/transaction/wallet.cpp
HalosGhost: An offline conversation finally helped this click for me. The first key is always *treated* as the minting key from the wallet's perspective; it's then up to someone configuring the system to specify the correct public key in the configured minting keys.

For the sake of testing determinism, this PR also includes some small machinery to set the minting key for a wallet to be a pre-defined key (a nothing-up-my-sleeve key); but in any tests or real configuration, that should not be used.

I'm all good from this perspective.

HalosGhost (approved):
@davebryson these changes look solid (sorry there was such a delay in-review). I have a small comment, and I'd like @metalicjames to take a quick look at it before merge, but I'm comfortable with merging.

src/uhs/client/client.hpp
davebryson: @HalosGhost  I primarily left it in the event someone objected to the changes and/or wanted the functionality for other reasons (testing?).

HalosGhost: Nothing occurs to me on why it would still be needed with the adopting of this (pretty simple even) minting mechanism, but perhaps @metalicjames will think otherwise (I'm open to hearing alternative opinions). :)

src/uhs/atomizer/sentinel/controller.cpp
```diff
@@ -167,7 +169,17 @@ namespace cbdc::sentinel {
...
+        // shards
+        if(ctx.m_inputs.empty()) {
+            auto ctx_pkt
+                = std::make_shared<cbdc::buffer>(cbdc::make_buffer(ctx));
+            if(!m_shard_network.send_to_one(ctx_pkt)) {
+                m_logger->error("Failed to send mint tx to shard");
+            }
+        } else {
```
metalicjames, line 183: Style point: prefer `return;` over `else` here to avoid the unnecessary condition nesting (which causes a cognitive complexity penalty).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting

src/uhs/atomizer/shard/shard.cpp
```diff
@@ -120,10 +120,13 @@ namespace cbdc::shard {
...
         }
 
+        atomizer::tx_notify_request msg;
+
+        // If the tx has no inputs, it's a mint.
         if(tx.m_inputs.empty()) {
-            return cbdc::watchtower::tx_error{
-                tx.m_id,
-                cbdc::watchtower::tx_error_inputs_dne{{}}};
+            msg.m_tx = std::move(tx);
```
metalicjames, line 127: Don't we need the sentinel attestations here?

`m_attestations` is not set.

src/uhs/client/client-cli.cpp
```diff
@@ -31,10 +33,48 @@ auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
...
+    if(!tx.has_value()) {
+        std::cout << "Could not generate valid mint tx." << std::endl;
+        return false;
+    }
+
+    std::cout << "tx_id:" << std::endl
+              << cbdc::to_string(cbdc::transaction::tx_id(tx.value()))
               << std::endl;
+
+    if(resp.has_value()) {
```
metalicjames, line 51: This code seems duplicated from `print_tx_result`, can we refactor it?

src/uhs/client/client.hpp
metalicjames: If we're not using it, let's remove it.

src/uhs/transaction/validation.cpp
```diff
@@ -33,8 +33,13 @@ namespace cbdc::transaction::validation {
         return std::tie(m_code, m_idx) == std::tie(rhs.m_code, rhs.m_idx);
     }
 
-    auto check_tx(const cbdc::transaction::full_tx& tx)
+    auto check_tx(const cbdc::transaction::full_tx& tx,
+                  const std::unordered_set<pubkey_t, hashing::null>& minters)
```
metalicjames, line 38: Is it possible to use a more general container here (like a vector, array etc)? Requiring a `std::unordered_set` with a custom hasher seems too specific to me.

src/uhs/transaction/validation.cpp
```diff
@@ -72,6 +77,38 @@ namespace cbdc::transaction::validation {
         return std::nullopt;
     }
 
+    auto
+    check_mint_tx(const cbdc::transaction::full_tx& tx,
+                  const std::unordered_set<pubkey_t, hashing::null>& minters)
+        -> std::optional<tx_error> {
+        // ensure there are outputs
```
metalicjames, line 85: Let's `assert(tx.m_inputs.empty())` at the start of this function so we don't misuse it accidentally.

src/uhs/transaction/validation.cpp
```diff
@@ -419,4 +456,115 @@ namespace cbdc::transaction::validation {
                                    && tx.verify(secp_context.get(), att);
                            });
     }
+
+    auto check_mint_witness(
```
metalicjames, line 461: These mint witness check functions seem almost identical to the existing functions for input witnesses. Can we refactor them somehow?

Here and below.

src/uhs/transaction/validation.hpp
```diff
@@ -118,8 +121,15 @@ namespace cbdc::transaction::validation {
     /// \note This function returns immediately on the first-found error.
     ///
     /// \param tx transaction to validate
+    /// \param minters public keys of authorized minters
     /// \return null if transaction is valid, otherwise error information
-    auto check_tx(const transaction::full_tx& tx) -> std::optional<tx_error>;
+    auto check_tx(const transaction::full_tx& tx,
+                  const std::unordered_set<pubkey_t, hashing::null>& minters)
+        -> std::optional<tx_error>;
+    auto
```
metalicjames, line 130: Can you add doc comments for the new functions in this file?

src/uhs/transaction/wallet.cpp
```diff
@@ -25,22 +25,55 @@ namespace cbdc {
...
+            tx.m_outputs.push_back(out);
+        }
 
-            ret.m_outputs.push_back(out);
+        const auto sighash = transaction::tx_id(tx);
+        tx.m_witness.resize(n_outputs);
+        const privkey_t seckey = m_keys.at(pubkey);
+
+        // Sign each output
+        for(size_t i = 0; i < n_outputs; i++) {
```
metalicjames, line 127: This signing code seems duplicated with the code in `wallet::sign`, can we refactor it?

src/uhs/transaction/wallet.cpp
```diff
@@ -25,22 +25,55 @@ namespace cbdc {
     auto transaction::wallet::mint_new_coins(const size_t n_outputs,
                                              const uint32_t output_val)
         -> transaction::full_tx {
-        transaction::full_tx ret;
```
metalicjames, line 28: Any reason to rename this variable?

src/uhs/transaction/wallet.cpp
```diff
@@ -121,6 +154,35 @@ namespace cbdc {
         return ret;
     }
 
+    auto transaction::wallet::generate_minter_key() -> pubkey_t {
+        if(m_pubkeys.empty()) {
+            // new wallet. create the minter key
+            return generate_key();
+        }
+        // first key is always the minter key
+        return m_pubkeys[0];
```
metalicjames: I think you need to hold the lock to access `m_pubkeys` here.

src/uhs/transaction/wallet.cpp
```diff
@@ -121,6 +154,35 @@ namespace cbdc {
...
+    }
+
+    auto transaction::wallet::generate_test_minter_key() -> pubkey_t {
```
metalicjames: I'm confused about this. Is the test minter key the one the validation code will check signed the TX? How is that different from the key generated when I can `generate_minter_key()`?

What if I call `generate_minter_key()` first, and then `generate_test_minter_key()` after? Now `m_pubkeys[1]` contains the minter key. Same thing if I call `generate_test_minter_key()` multiple times.

metalicjames (changes requested):
I have a general question. Can I make multiple mint transactions with this PR?

It seems that all mint transactions will be identical, resulting in the same outpoints for every mint TX, meaning I can actually only make one valid mint TX without overwriting (or re-adding!! previously used outputs). One can see how this could lead to replay attacks where a previously submitted transaction spending a mint output becomes valid again when its input(s) are recreated.

The signatures on the outputs are missing any kind of sequence number or nonce. Does that matter? Is there a replay attack vector on mint transactions without one?

src/uhs/transaction/wallet.cpp
davebryson: Some of the code is duplicate.  As I was touching a lot of the code base, I didn't want to be too intrusive

src/uhs/transaction/wallet.cpp
davebryson: `generate_test_minter_key` generates a pre-determined key that's preset in all existing `.cfg` files for testing and/or demos.  Again, I was looking for a way to minimize changes across the codebase. See https://github.com/mit-dci/opencbdc-tx/blob/eb16c651e649042ddbbfbea3b05eb0b219dae981/src/uhs/transaction/wallet.hpp#L129 for a descriptoin

src/uhs/transaction/wallet.cpp
metalicjames: I think it's best to refactor the duplicate code in this PR. That will make the code more maintainable in the long run. You can split the refactors into separate commits to reduce the review burden.

src/uhs/transaction/wallet.cpp
metalicjames: Can we not read the key from the config file? Why do we need code to re-generate it?

davebryson: @metalicjames  Mainly because tests call wallet code to mint (especially the integration tests) and you need a deterministic key for that.  Putting the minter's private key in the configuration file for testing would mean you should put all minter private keys in the configuration file (testing and non-testing) to keep the design consistent.   But IMO, putting minter (or any) private keys in the configuration file seems like a really bad idea from a security perspective.

src/uhs/atomizer/shard/shard.cpp
davebryson: @metalicjames  This is related to an issue that I asked about early in the process of working on this.  The Atomizer, unlike 2PC,  actually requires `inputs` in the compact transaction.  A mint transaction doesn't have any inputs.  So making mint transactions work with the Atomized feels like a hack and deviates from the normal transaction flow.  Any thoughts or insights you have on this would be appreciated.

src/uhs/transaction/wallet.cpp
HalosGhost: Worth noting that this is not a production system, so we don't have to be quite as worried about such things at the moment.

For example, for sentinel attestation checks, the sentinel private keys *are* stored in the configuration files.

davebryson: @HalosGhost Agree and I understand what you're saying.  But on the other hand, adding security late to the process usually doesn't turn out well.   If security is not a high-priority in the current context of the project, then is secure minting even necessary right now?  Just thinking out loud on how to best bound the problem...

HalosGhost: I take your point. I suppose my view is that we should try to scope our questions as small as possible.

I think this PR is answering the question “How could secure minting be supported?” That doesn't require all other best-practices to be addressed at the same time. To that end, having the keys in-configuration or not is an orthogonal concern (in my opinion).

davebryson:
> The signatures on the outputs are missing any kind of sequence number or nonce. Does that matter? Is there a replay attack vector on mint transactions without one?

@metalicjames Great question and I believe you are correct.  Adding a nonce/sequence number would be an easy solution.  Of course a sequence number would require some form of state known by the system for validation.  But either way you have minters using the same key over and over.  Policy could dictate a minter must rotate a key after each use, but this complicates configuration.

src/uhs/atomizer/shard/shard.cpp
metalicjames: Hmm, I think the atomizer should be able to understand mint transactions. Otherwise, how would the mint transaction be included in a block and applied by the shards? If the transaction is included in a block, the atomizer should check the sentinel attestations. A valid mint transaction should still be validated by multiple sentinels and attested to, the same way as any other transaction. Otherwise, how can we be confident the mint transaction wasn't validated by a malicious sentinel, ignoring the authorized minter keys?

davebryson: @metalicjames A mint transaction flows through the sentinel and accumulates attestations the same as any other transaction.  The controller checks the attestations before it forwards it to the shard: https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/uhs/atomizer/shard/controller.cpp#L152.  All the shard appears to do is collect the input  index into the msg attestations set https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/uhs/atomizer/shard/shard.cpp#L151.  What concerns me more is the fact that atomizer allow this (at least passes all tests) even though the msg.m_attestations is empty for a mint transaction.

PR #116, opened 2022-06-02
Merge macOS build steps

jonwiggins:
Hi,

I tried to make some additional improvements to the build/run steps for mac.
1. The additional build steps for mac have been moved from `README.md` into `./scripts/configure.sh`. This means that the instructions for linux and macOS are now the same. I left the note on installing xcode tools in the instructions because I believe it triggers a prompt to accept an agreement. 
2. The install location for `brew` is different on intel and arm macs (https://docs.brew.sh/Installation), this updates the `clang-tidy` linking command to use the correct location.
3. I noticed that the `build.sh` script could also fail due to missing `gmake`, so it is now linked if needed.

Let me know what you think. I'm happy to make any requested changes.

Thanks

Signed-off-by: Jon Wiggins <me@jonwiggins.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,28 +56,21 @@ For more information on how to contribute, please see our [Contribution Guide](d
 Use these directions if you want to build the source on your machine.
 If you just want to run the system, see "Run the Code" below.
 
+1. Install the necessary libraries and resources
+  ```terminal
+  # ./scripts/configure.sh
+  ```
+2. Run the build
+  ```terminal
+  # ./scripts/build.sh
+  ```
+
 ## macOS
-Ensure your development environment is set correctly for clang:
+Note that if you have not already installed the xcode cli tools you will need to:
 
-`sudo xcode-select -switch /Library/Developer/CommandLineTools`
-
-Or, if you've changed this in the past, you can reset to point to commandline tools with:
-
-`sudo xcode-select --reset`
-
-
-1. Install dependencies: `brew install leveldb llvm@11 googletest lcov make wget cmake`
-2. `./scripts/configure.sh`
-3. `./scripts/build.sh`
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,29 +56,14 @@ For more information on how to contribute, please see our [Contribution Guide](d
...
-
-## Linux
-
 1. `./scripts/configure.sh`
 2. `./scripts/build.sh`
 
+## macOS
+Note that if you have not already installed the xcode cli tools you will need to:
+
+`xcode-select --install`
```
HalosGhost: The original PR probably should have had code snippets like this as full code blocks (as we do most elsewhere in the readme). Can that be done here?

scripts/configure.sh
```diff
@@ -24,6 +24,17 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
   CPUS=$(grep -c ^processor /proc/cpuinfo)
 elif [[ "$OSTYPE" == "darwin"* ]]; then
   CPUS=$(sysctl -n hw.ncpu)
+  # ensure development environment is set correctly for clang
```
HalosGhost, line 27: I do love the idea of simplifying down so that the workflow remains mostly just:

1. configure
2. build
3. run

(harkening back to the `Makefile` days of old). @davebryson does this seem like a reasonable simplification to you?

README.md
jonwiggins: Sounds good.
I also added blocks to the build steps themselves so they should all match.
Let me know if you actually wanted the configure and build commands to stay the way they were.

scripts/configure.sh
davebryson: Yes.  Looks good to me!

HalosGhost (approved):
Excellent simplification of the getting-started process for macOS.

Looks good to me.

PR #118, opened 2022-06-02
Add unit test to cover non-error cases of cbdc::nuraft_serializer::read()

### LINKED ISSUE #117: Add unit test to cover non-error cases of cbdc::nuraft_serializer::read()
This issue references a task in https://github.com/mit-dci/opencbdc-tx/issues/3, Test Coverage / Quality Umbrella.

The cbdc::nuraft_serializer::read() does not cover non-error cases

To resolve this issue, a unit test was added to raft_test.cpp called test_raft_serializer_read which makes sure the test hash which is read from the cbdc::nuraft_serializer is in fact correct. It also moves cursor, reads and makes sure the data is once again what it should be relative to test_hash.

ykaravas:
This pull request addresses https://github.com/mit-dci/opencbdc-tx/issues/117#issue-1257715247 

Added a unit test to cover non error case for cbdc::nuraft_serializer::read()

Closes https://github.com/mit-dci/opencbdc-tx/issues/117

### DIFF
--- a/tests/unit/raft_test.cpp
+++ b/tests/unit/raft_test.cpp
@@ -321,6 +321,24 @@ TEST_F(raft_test, test_raft_serializer_basic) {
     ASSERT_TRUE(ser.end_of_buffer());
 }
 
+TEST_F(raft_test, test_raft_serializer_read) {
+    auto new_log = nuraft::buffer::alloc(32);
+    auto test_hash = cbdc::hash_from_hex(
+        "cb7b43951ffcfe400a5432749a79096e632ef2e6328a28049c9af55b85fb260d");
+    auto ser = cbdc::nuraft_serializer(*new_log);
+    ser << test_hash;
+    ser.reset();
+    ASSERT_FALSE(ser.end_of_buffer());
+    auto read_output = nuraft::buffer::alloc(32);
+    ASSERT_TRUE(ser.read(read_output->data(), 32));
+    ASSERT_EQ(*read_output->data(), *test_hash.data());
+    ser.reset();
+    // Test advance_cursor method with read to follow as well.
+    ser.advance_cursor(10);
+    ASSERT_TRUE(ser.read(read_output->data(), 22));
+    ASSERT_EQ(*read_output->data(), *(test_hash.data() + 10));
+}
+
 TEST_F(raft_test, test_raft_serializer_out_of_bounds) {
     auto new_log = nuraft::buffer::alloc(32);
     auto test_hash = cbdc::hash_from_hex(


HalosGhost:
@ykaravas this looks simple and solid to me (adding a happy-path, round-trip test case for the raft serialization). Can you squash to a single commit and I'll merge?

ykaravas:
Hello @HalosGhost I seem to be having trouble squashing the commits. It is not a practice i am used to doing and having the commits merged in from changes to trunk interspersed with mine is confusing me a bit as to how i would accomplish this. Namely, i am wondering how to deal with commits interspersed with mine from merging trunk into my branch along the way. I think my mistake was merging trunk into my branch along the way instead of rebasing. Should i simply create new branch with my changes, and in turn, a new pull request? Please advise.

HalosGhost:
@ykaravas it's a bit new for many; no worries! The easiest way to handle newly-added commits upstream is (assuming you have this fork as a remote locally named `upstream`) `git pull --rebase upstream trunk`. That will replay all your local commits on top of `trunk`, and should put you in a good place. After that `git rebase -i` for an interactive rebase (appropriately choosing `squash` or `fixup` to get to a single commit) should have everything cleaned up!

ykaravas:
@HalosGhost Awesome thank you! I had to do a force push to get it to update, however; is that normal? If so, it should be all set now.

HalosGhost:
> I had to do a force push to get it to update, however; is that normal?

Git is very cautious about allowing history-rewriting (which is what rebasing is), so yeah; after rebasing, you'll always need to force-push.

tests/unit/raft_test.cpp
```diff
@@ -321,6 +321,23 @@ TEST_F(raft_test, test_raft_serializer_basic) {
...
+    ser << test_hash;
+    ser.reset();
+    ASSERT_FALSE(ser.end_of_buffer());
+    auto read_output = nuraft::buffer::alloc(32);
+    ASSERT_TRUE(ser.read(read_output->data(), 32));
+    ASSERT_EQ(*read_output->data(), *test_hash.data());
+    ser.reset();
+    ser.advance_cursor(10);
+    ASSERT_TRUE(ser.read(read_output->data(), 22));
+    ASSERT_EQ(*read_output->data(), *(test_hash.data() + 10));
```
HalosGhost, line 339: So, I believe I understand what this is testing (skipping the first 10 bytes and reading the other 22, do those 22 bytes match what we expect), but I'm not quite sure why that should be tested. Lines 333 and 334 test that the full 32-bytes are equal.

ykaravas: I suppose it is simply a test of the advance_cursor but if you like, i can simply remove it.

HalosGhost: I don't mind it, just wasn't clear on it. Maybe it's worth adding a comment above 336 to clarify that's the aim?

ykaravas: Sounds good. I added a comment.

HalosGhost (approved):
Adds a simple test to cover the happy-path of part of our serialization/deserialization code.

Looks good to me.

PR #115, opened 2022-06-01
Fix compilation on MacOS-arm64 using Apple Clang

metalicjames:
This PR adds type postfixes to the literals in the tests. This is required by Apple Clang which complains about mismatched types.

```
[build] /usr/bin/clang++ -DGTEST_LINKED_AS_SHARED_LIBRARY=1 -D_NO_EXCEPTION -I/Users/james/workspace/cbdc-universe0/3rdparty -I/Users/james/workspace/cbdc-universe0/3rdparty/secp256k1/include -I/usr/local/include -I/opt/homebrew/include -I/Users/james/workspace/cbdc-universe0/tests/. -I/Users/james/workspace/cbdc-universe0/tests/../src -I/Users/james/workspace/cbdc-universe0/tests/../tools/watchtower -I/Users/james/workspace/cbdc-universe0/tests/../3rdparty -I/Users/james/workspace/cbdc-universe0/tests/../3rdparty/secp256k1/include -isystem /opt/homebrew/anaconda3/include -isystem /usr/local/include -fno-rtti -g -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk -fprofile-arcs -ftest-coverage -Wall -Wextra -pedantic -Werror -fno-exceptions -Wshadow-all -Wnewline-eof -std=c++17 -MD -MT tests/unit/CMakeFiles/run_unit_tests.dir/serialization_test.o -MF tests/unit/CMakeFiles/run_unit_tests.dir/serialization_test.o.d -o tests/unit/CMakeFiles/run_unit_tests.dir/serialization_test.o -c /Users/james/workspace/cbdc-universe0/tests/unit/serialization_test.cpp
[build] In file included from /Users/james/workspace/cbdc-universe0/tests/unit/serialization_test.cpp:11:
[build] /opt/homebrew/include/gtest/gtest.h:1545:11: error: comparison of integers of different signs: 'const unsigned long' and 'const int' [-Werror,-Wsign-compare]
[build]   if (lhs == rhs) {
[build]       ~~~ ^  ~~~
[build] /opt/homebrew/include/gtest/gtest.h:1564:12: note: in instantiation of function template specialization 'testing::internal::CmpHelperEQ<unsigned long, int>' requested here
[build]     return CmpHelperEQ(lhs_expression, rhs_expression, lhs, rhs);
[build]            ^
[...]

PR #120, opened 2022-06-02
Fix another instance of mismatched types with Apple Clang

metalicjames:
Follows on from #115.


HalosGhost:
This is a really small and easy-evaluated change. Is this the last instance so far, or are there other places that should be included as well?

metalicjames:
I think this is the last (it compiles from scratch with this change).

HalosGhost (approved):
Does what it says on the tin. Fix (hopefully) the last instance of mismatched literal types.

Issue #122, opened 2022-06-03
Extend test.sh


mszulcz-mitre:
The script `opencbdc-tx/scripts/test.sh` runs integration and unit tests and measures test coverage.   It has a few limitations:
- It must be run from the top-level folder in the repo (`opencbdc-tx/`).
- It assumes the build directory is called `build` and is located in the top-level folder of the repo (i.e. `opencbdc-tx/build`).
- It always runs unit and integration tests and measures coverage.  

I think it could be improved by:
- Allowing the user to run it from any folder.
- Allowing the user the option of defining the build folder as a command-line argument.
- Allowing the user to choose via command-line arguments which tests to run (e.g. unit tests, integration tests, or both) and whether to measure coverage.

I think these changes can be made easily and can preserve backward compatibility (e.g. running the modified script with no command-line arguments would do what the script currently does).

PR #123, opened 2022-06-07
Extend test script

mszulcz-mitre:
This pull request addresses [Issue 122, 'Extend test.sh'](https://github.com/mit-dci/opencbdc-tx/issues/122).  `test.sh` is the script that runs unit tests and integration tests and measures coverage.  This pull request contains commits that extend it in the following ways:

- Users can now choose via command-line arguments which tests to run (e.g. unit tests, integration tests, or both) and whether to measure coverage.
- Users can now define the build folder via a command-line argument.
- Users can now run the script from any folder.
- Users can now read about what the script does and how to use it by calling it with `-h` or `--help` flags.

In addition, `test.sh` now adheres to the 80-char. line limit followed in the repo's C++ code.  It also now has variable references enclosed in double quotes, as recommended in the [Advanced Bash Scripting Guide](https://tldp.org/LDP/abs/html/quotingvar.html).

The new `test.sh` in this pull request is completely _backward compatible_:  running it with no command-line arguments does exactly what the previous version of the script does.  As a result, documentation that references the script and code that invokes it should not need to change.  Also, users don't need to change how they use the script if they don't want to.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -10,12 +10,19 @@ DESCRIPTION:
     This script runs unit and integration tests and measures test coverage. 
 
 FLAGS: 
+    -nu, --no-unit-tests            Do not run unit tests.
+                                    Default:  false
+    -ni, --no-integration-tests     Do not run integration tests.
+                                    Default:  false
     -nc, --no-coverage              Do not measure test coverage.
                                     Default:  false
     -h, --help                      Show usage.
 	"
 }
 
+# Parse command-line arguments.
+RUN_UNIT_TESTS="true"
+RUN_INTEGRATION_TESTS="true"
 MEASURE_COVERAGE="true"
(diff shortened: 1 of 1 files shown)


HalosGhost:
Rather than opening multiple pull-requests and closing them, it probably makes more sense to open one and force-push to override changes. Feel free to use Draft PRs as well if something isn't quite ready for review! :)

mszulcz-mitre:
Sorry about that!  I'll open a new draft PR.  I probably should've done that from the start.

If there's any way to delete these closed PRs, they're probably not useful to keep.

HalosGhost:
Don't worry about it (just reduces a bit of noise for me). And Github actually doesn't support deleting PRs (requires specifically asking support to do so), but they're not doing any harm.

PR #126, opened 2022-06-07
Extend test script

### LINKED ISSUE #122: Extend test.sh

mszulcz-mitre:
This pull request addresses [Issue 122, 'Extend test.sh'](https://github.com/mit-dci/opencbdc-tx/issues/122).  `test.sh` is the script that runs unit tests and integration tests and measures coverage.  This pull request contains commits that extend it in the following ways:

- Users can now choose via command-line arguments which tests to run (e.g. unit tests, integration tests, or both) and whether to measure coverage.
- Users can now define the build folder via a command-line argument.
- Users can now run the script from any folder.
- Users can now read about what the script does and how to use it by calling it with `-h` or `--help` flags.

In addition, `test.sh` now adheres to the 80-char. line limit followed in the repo's C++ code.  It also now has variable references enclosed in double quotes, as recommended in the [Advanced Bash Scripting Guide](https://tldp.org/LDP/abs/html/quotingvar.html).

The new `test.sh` in this pull request is completely _backward compatible_:  running it with no command-line arguments does exactly what the previous version of the script does.  As a result, documentation that references the script and code that invokes it should not need to change.  Also, users don't need to change how they use the script if they don't want to.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# exit script on failure
+# Exit script on failure.
 set -e
 
 if [ -z ${BUILD_DIR+x} ]; then
@@ -7,9 +7,9 @@ if [ -z ${BUILD_DIR+x} ]; then
 fi
 
 run_test_suite () {
-    cd $BUILD_DIR
+    cd "$BUILD_DIR"
     find . -name '*.gcda' -exec rm {} \;
-    $PWD/$1
+    "$PWD"/"$1"
 
     LOCATION=$2
     rm -rf $LOCATION
@@ -31,5 +31,5 @@ echo "Running unit tests..."
 run_test_suite "tests/unit/run_unit_tests" "unit_tests_coverage"
 
 echo "Running integration tests..."
-cp tests/integration/*.cfg $BUILD_DIR
+cp tests/integration/*.cfg "$BUILD_DIR"
 run_test_suite "tests/integration/run_integration_tests" "integration_tests_coverage"
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -2,6 +2,45 @@
 # Exit script on failure.
 set -e
 
+function usage() { echo "
+USAGE:
+    ./test.sh
+
+DESCRIPTION:
+    This script runs unit and integration tests and measures test coverage. 
+
+FLAGS: 
+    -nc, --no-coverage              Do not measure test coverage.
+                                    Default:  false
+    -h, --help                      Show usage.
+	"
+}
+
+MEASURE_COVERAGE="true"
+while [[ $# -gt 0 ]]
+do
+    case $1 in
+        -h|--help)
+            usage
+            exit 0
+            ;; 
+        -nc|--no-coverage)
+            MEASURE_COVERAGE="false"
+            shift
+            ;; 
+        -*)
+            echo "ERROR:  unknown command-line option '${1}'."
+            usage
+            exit 1
+            ;;
+        *)
+            echo "ERROR:  unknown command-line option '${1}'."
+            usage
+            exit 1
+            ;;
+    esac
+done
+
 if [ -z ${BUILD_DIR+x} ]; then
     export BUILD_DIR=build
 fi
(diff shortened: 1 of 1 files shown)


HalosGhost:
@mszulcz-mitre I love this; even minimal QoL improvements like this make a world of difference in the development cycle.

I have one thing that's been in the back of my head for a while that I'd wonder if it would make sense to include in this change: googletest supports pattern-based test filters (allowing you to run only a subset of tests) via [`--gtest_filter`](https://google.github.io/googletest/advanced.html#running-a-subset-of-the-tests).

Do you think it would be reasonable to include support for passing through a test filter to either/both of the suites?

mszulcz-mitre:
@HalosGhost I love the idea to include `--gtest_filter`!  I just added it.  While I'm at it, I'll also include support for the other gtest options (`--gtest_output`, `--gtest_repeat`, and `--gtest_break_on_failure`).

After adding `--gtest_filter`, I ran some tests with junk filters to confirm that no tests would be run and that the coverage would be 0%.  For example, I ran 
```
$ sudo ./scripts/test.sh --gtest_filter=blarg
```
As expected, no unit tests or integration tests were run.  Unexpectedly, though, the coverage wasn't 0%.  For both unit and integration tests, it was
```
Overall coverage rate:
  lines......: 0.2% (11 of 6120 lines)
  functions..: 0.1% (2 of 1888 functions)
  branches...: 0.1% (2 of 2008 branches)
``` 
All the lines, functions, and branches that were hit are in `util/common/random_source.hpp` and `util/common/random_source.cpp`.  It doesn't make sense to me why these two files would be hit, and when I look at line-by-line coverage in the the .html output, it gets stranger still.  The .html output shows that the only line hit in the header is the destructor, but the only lines hit in the .cpp file were for the constructor.

Any thoughts on why the coverage is non-zero?  I ran the test and coverage commands manually from the command line and got exactly the same results, so the issue isn't related to how `test.sh` is passing the `--gtests_filter` args to the test suites.

mszulcz-mitre:
@HalosGhost I finished enabling test.sh to accept all Google Test flags.  I also clarified the documentation.  These changes have been added to the pull request.

HalosGhost:
> As expected, no unit tests or integration tests were run. Unexpectedly, though, the coverage wasn't 0%. For both unit and integration tests, it was
> 
> ```
> Overall coverage rate:
>   lines......: 0.2% (11 of 6120 lines)
>   functions..: 0.1% (2 of 1888 functions)
>   branches...: 0.1% (2 of 2008 branches)
> ```

That is a bit puzzling to say the least…

> All the lines, functions, and branches that were hit are in `util/common/random_source.hpp` and `util/common/random_source.cpp`. It doesn't make sense to me why these two files would be hit, and when I look at line-by-line coverage in the the .html output, it gets stranger still. The .html output shows that the only line hit in the header is the destructor, but the only lines hit in the .cpp file were for the constructor.

My immediate guess would be that, for some reason, the random_source is being initialized before any tests are run (and thus being destroyed at the end); but I'm not sure why that would be happening for it and none of the other utility classes…

scripts/test.sh
```diff
@@ -1,35 +1,185 @@
...
+            ;;
+        --gtest_*)
+            GTEST_FLAGS+=("$1")
+            shift
+            ;;
+        -*)
+            echo "ERROR:  unknown command-line option '${1}'."
+            exit 1
+            ;;
+        *)
```
HalosGhost:
```suggestion
        *)
```

Is there a reason to separate these cases?

scripts/test.sh
```diff
@@ -1,35 +1,185 @@
...
+    esac
+done
+
+# If the build folder is not provided as a command-line argument or as an
+# environment variable, assume it's named 'build' and is located in the
+# top-level directory of the repo.  By defining the top-level directory relative
+# to the location of this script, the user can run this script from any folder.
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+REPO_TOP_DIR="${SCRIPT_DIR}/.."
+if [ -z ${BUILD_DIR+x} ]
```
HalosGhost: We should probably quote `${BUILD_DIR+x}` (in general, all shell expansions should be double-quoted); additionally, since we're relying on bash, we can count on `[[` to be present, and it avoids a lot of odd pitfalls that `[` has. Unless there is a specific reason to prefer it (the one that would normally come up for me would be portability), I'd prefer we standardize our conditionals to be of the form

```bash
if [[ condition ]]; then
    body
fi
```

scripts/test.sh
```diff
@@ -1,35 +1,185 @@
...
+then
+    BUILD_DIR="${REPO_TOP_DIR}/build"
+fi
+
+if [ ! -d "$BUILD_DIR" ]
```
HalosGhost: This is not unreasonable, but we could also `mkdir -p "$BUILD_DIR"`; is there a reason to prefer just failing out rather than creating the build directory?

HalosGhost (changes requested):
This looks like such a fabulous QoL improvement. I'm quite excited to merge it. I have a few small questions/suggestions to address first (made inline).

HalosGhost:
@AlexRamRam, would you be okay with 98928a14792192e934105c6ab71b599a96d8ebc7 being pulled into this PR?

AlexRamRam:
> 

Yes. Thank you.

AlexRamRam:
> @AlexRamRam, would you be okay with [98928a1](https://github.com/mit-dci/opencbdc-tx/commit/98928a14792192e934105c6ab71b599a96d8ebc7) being pulled into this PR?

Please use [this commit](https://github.com/mit-dci/opencbdc-tx/pull/109/commits/484214791c3388b101838be6b54ad486cf762aa9) which contains the two changed elements for replacing `cp --parent` with `rsync`. Thank you.

HalosGhost:
@mszulcz-mitre can you pull 484214791c3388b101838be6b54ad486cf762aa9 from Alex's fork into this PR (fits better here in my opinion)?

HalosGhost:
For the sake of avoiding staleness, I merged Alex's PR which included 4842147. This PR needs a rebase and a conflict resolution (should be quick).

scripts/test.sh
mszulcz-mitre: Nope--no good reason.  Good catch!

scripts/test.sh
mszulcz-mitre: Good call!  I think I grabbed that line from the original version and didn't even think about it, but you're right.  Thanks for flagging it!

scripts/test.sh
mszulcz-mitre: We could create the build folder as you suggest, but then all the tests will crash or fail anyways since the new folder won't contain any code to test.  In that case, I was thinking it would be clearer for users to create the build folder themselves through whatever step they're taking to build the code (maybe by running build.sh).

However, I'm happy to do it either way!

mszulcz-mitre:
@HalosGhost I rebased, resolved conflicts, and made the requested changes (except for the one under discussion in https://github.com/mit-dci/opencbdc-tx/pull/126#discussion_r909618903, but am happy to if you like).  After pushing, though, it looks like one of the unit tests timed out (raft_test.basic_raft_cluster_fail_test).  This doesn't happen when I run the tests in Docker locally.  Any thoughts on the cause?

scripts/test.sh
HalosGhost: No worries; it looked like it was straight from the original (and I didn't mean to assign blame), just figured it's possible for us to improve it now. ☺

scripts/test.sh
HalosGhost: Oh, I got myself turned around. I wholly agree that it makes sense to fail out if we're running tests and there's no build directory. Pardon the noise.

HalosGhost:
@mszulcz-mitre yeah, this is one of those test-suite-fragility issues that we need to address in the near future (if only to increase confidence that the test-suite should provide). I reran the unit/integration CI/CD step, and it all passed. I'll update my review today!

HalosGhost (approved):
This is a fabulous test script improvement. I'm quite excited to leverage the changes.

PR #132, opened 2022-06-13
Add NuRaft handler commit_config()

### LINKED ISSUE #125: [atomizer::state_machine] fix assertion failure in create_snapshot() and clarify meaning of last committed log index
### Affected Branch

trunk

### Basic Diagnostics

### Synopsis
In `cbdc::atomizer::state_machine::create_snapshot(snapshot& s ...)` we 
[`assert(s.get_last_log_idx() == last_commit_index())`](https://github.com/mit-dci/opencbdc-tx/blob/7e3dd76ec4e4d13ab78f8c0f372fad670240affa/src/uhs/atomizer/atomizer/state_machine.cpp#L206)
where `last_commit_index()` returns the last log index processed by [`::commit()`](https://github.com/mit-dci/opencbdc-tx/blob/7e3dd76ec4e4d13ab78f8c0f372fad670240affa/src/uhs/atomizer/atomizer/state_machine.cpp#L42).

When snapshot is enabled, the assertion fails because log index last processed by [`::commit()`](https://github.com/mit-dci/opencbdc-tx/blob/7e3dd76ec4e4d13ab78f8c0f372fad670240affa/src/uhs/atomizer/atomizer/state_machine.cpp#L42) is often smaller than the requested snapshot log index.
This is because NuRaft creates log entries for cluster configuration changes which invoke [`state_machine::commit_config(...)`](https://github.com/eBay/NuRaft/blob/9d7c4cb04d0fe0e18e6b2ae4fbef503bfa7da8f0/include/libnuraft/state_machine.hxx#L77-L84) instead of the application log commit [state_machine::commit(...)](https://github.com/eBay/NuRaft/blob/9d7c4cb04d0fe0e18e6b2ae4fbef503bfa7da8f0/include/libnuraft/state_machine.hxx#L51-L67).  See NuRaft's [commit thread](https://github.com/eBay/NuRaft/blob/9d7c4cb04d0fe0e18e6b2ae4fbef503bfa7da8f0/src/handle_commit.cxx#L233-L238) for  details.
This optional hook [`state_machine::commit_config(...)`](https://github.com/eBay/NuRaft/blob/9d7c4cb04d0fe0e18e6b2ae4fbef503bfa7da8f0/include/libnuraft/state_machine.hxx#L77-L84) is currently unimplemented for `atomizer::state_machine`, therefore the state machine's self-maintained last commit index is not in sync with the system's last log index.

### Two potential solutions:
1) Pretend the in-memory state machine "comits" cluster configuration change logs. Keep the assertion in place, and sync the last commited log index (`::m_last_commited_idx`) with the last snapshot log index by implementing the optional and otherwise empty callback `::commit_config()` like so:
```
    void state_machine::commit_config(const nuraft::ulong log_idx,
                                      nuraft::ptr<nuraft::cluster_config>&) {
        m_last_committed_idx = log_idx;
    }
```
Keeping `m_last_committed_idx` in sync in this way with the system log index provides additional assurance that the state machine is up-to-date. However, this assurance is likely already provided by the framework and may not be necessary (I'm not sure about this!). In any case, this provides clarity and comfort.
This may become fragile as NuRaft implementation changes over time (i.e. different cluster management message types are added invoking different hooks).

2) Manage only what we care about. Maintain the last commit index as we do now with the understanding that, currently, the state_machine only cares about application log entries via `::commit()`.  We relax the assertion accordingly (`assert(s.get_last_log_idx() >= last_commit_index());`), and explicitly save/re-populate this value in `create_snapshot()`/`apply_snapshot()` by adding new member `m_last_committed_idx`
to member struct [`cbdc::atomzier::state_machine::snapshot`](https://github.com/mit-dci/opencbdc-tx/blob/7e3dd76ec4e4d13ab78f8c0f372fad670240affa/src/uhs/atomizer/atomizer/state_machine.hpp#L115-L124)

### In order to reproduce the issue, follow these steps:

1. Enable snapshot feature: in `replicated_atomizer.cfg` add line `snapshot_distance=2`
2. `tests/integration/run_integration_tests --gtest_filter=replicated_atomizer_integration_tests.*`

### Code of Conduct

AlexRamRam:
Add NuRaft handler commit_config() to maintain synchronization of last committed index with latest log index

Fix #125
Signed-off-by: Alexander Jung <104335080+AlexRamRam@users.noreply.github.com>

### DIFF
--- a/src/uhs/atomizer/atomizer/state_machine.cpp
+++ b/src/uhs/atomizer/atomizer/state_machine.cpp
@@ -41,6 +41,7 @@ namespace cbdc::atomizer {
 
     auto state_machine::commit(nuraft::ulong log_idx, nuraft::buffer& data)
         -> nuraft::ptr<nuraft::buffer> {
+        assert(log_idx == m_last_committed_idx + 1);
         m_last_committed_idx = log_idx;
         auto req = from_buffer<request>(data);
         assert(req.has_value());
@@ -102,6 +103,7 @@ namespace cbdc::atomizer {
     void state_machine::commit_config(
         const nuraft::ulong log_idx,
         nuraft::ptr<nuraft::cluster_config>& /*new_conf*/) {
+        assert(log_idx == m_last_committed_idx + 1);
         m_last_committed_idx = log_idx;
     }
 
--- a/src/uhs/atomizer/atomizer/state_machine.cpp
+++ b/src/uhs/atomizer/atomizer/state_machine.cpp
@@ -99,6 +99,12 @@ namespace cbdc::atomizer {
             resp.value());
     }
 
+    void state_machine::commit_config(
+        const nuraft::ulong log_idx,
+        nuraft::ptr<nuraft::cluster_config>& /*new_conf*/) {
+        m_last_committed_idx = log_idx;
+    }
+
     auto
     state_machine::read_logical_snp_obj(nuraft::snapshot& s,
                                         void*& /* user_snp_ctx */,
--- a/src/uhs/atomizer/atomizer/state_machine.hpp
+++ b/src/uhs/atomizer/atomizer/state_machine.hpp
@@ -45,6 +45,12 @@ namespace cbdc::atomizer {
         [[nodiscard]] auto commit(nuraft::ulong log_idx, nuraft::buffer& data)
             -> nuraft::ptr<nuraft::buffer> override;
 
+        /// Handler for the raft cluster configuration changes
+        /// \param log_idx Raft log number of the configuration change.
+        void commit_config(
+            nuraft::ulong log_idx,
+            nuraft::ptr<nuraft::cluster_config>& /*new_conf*/) override;
+
         /// Read the portion of the state machine snapshot associated with
         /// the given metadata and object ID into a buffer.
         /// \param s metadata of snapshot to read.
--- a/src/uhs/twophase/coordinator/state_machine.cpp
+++ b/src/uhs/twophase/coordinator/state_machine.cpp
@@ -13,6 +13,7 @@
 namespace cbdc::coordinator {
     auto state_machine::commit(uint64_t log_idx, nuraft::buffer& data)
         -> nuraft::ptr<nuraft::buffer> {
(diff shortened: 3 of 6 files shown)


src/uhs/atomizer/atomizer/state_machine.cpp
```diff
@@ -99,6 +99,12 @@ namespace cbdc::atomizer {
             resp.value());
     }
 
+    void state_machine::commit_config(
+        const nuraft::ulong log_idx,
+        nuraft::ptr<nuraft::cluster_config>& /*new_conf*/) {
+        m_last_committed_idx = log_idx;
```
metalicjames, line 107: Is it worth asserting here that `log_idx > m_last_committed_idx`, or even stronger that `log_idx == m_last_committed_idx + 1`?

src/uhs/atomizer/atomizer/state_machine.hpp
```diff
@@ -45,6 +45,12 @@ namespace cbdc::atomizer {
         [[nodiscard]] auto commit(nuraft::ulong log_idx, nuraft::buffer& data)
             -> nuraft::ptr<nuraft::buffer> override;
 
+        /// Handler for the raft cluster configuration changes
+        /// \param log_idx Raft log number of the configuration change.
+        void commit_config(
```
metalicjames, line 50: I think we should add this to all the state machines, not just atomizer.

src/uhs/atomizer/atomizer/state_machine.cpp
AlexRamRam: I agree about `log_idx == m_last_committed_idx + 1`. Will add this to both `commit()` and `commit_config()`

src/uhs/atomizer/atomizer/state_machine.hpp
AlexRamRam: Agreed.

HalosGhost (approved):
This looks like a very reasonable solution to the problem. Both minimal and clear.

Looks good to me.

PR #135, opened 2022-06-19
tcp_client::init bug fix

### LINKED ISSUE #131: Impossible to trigger error "Failed to start coordinator client"
### Affected Branch

trunk

### Basic Diagnostics

- [ ] The issue is reproducible in docker

### Description

Consider the following branch in `sentinel_2pc::controller::init`(src/uhs/twophase/sentinel_2pc/controller.cpp, lines 38-41):
```
        if(!m_coordinator_client.init()) {
            m_logger->error("Failed to start coordinator client");
            return false;
        }
```
The top line checks if the coordinator client fails to start. If it does fail, the next lines should log the error "Failed to start coordinator client" and should return from the method with a value of false.  Surprisingly, it is _impossible_ to execute this branch because `m_coordinator_client.init` _always_ returns true.

The culprit seems to be `rpc::tcp_client::init` (src/util/rpc/tcp_client.hpp, lines 54 - 68), which is in the call stack of `m_coordinator_client.init`.  According to its description, `rpc::tcp_client::init`  “connects to … server endpoints”.  Its body indicates that if the connection fails, it should return false:
```
     [[nodiscard]] auto init() -> bool {
            if(!m_net.cluster_connect(m_server_endpoints, false)) {
                return false;
            }
	…
        }
```
However, it can never return false.  This is due to the fact that `cluster_connect` is called with its 2nd argument equal to false.  Since `cluster_connect(m_server_endpoints, false)` always returns true, its caller `rpc::tcp_client::init` always returns true, and ultimately `m_coordinator_client.init` always returns true.  Here's a summary of the call stack:

- m_net.cluster_connect(m_server_endpoints, false) (always true)
  - rpc::tcp_client::init (always true)
    - coordinator::rpc::init (always true)
      - m_coordinator_client.init (always true)

Because `sentinel_2pc::controller::init` never raises an error if the coordinator client fails to start, an infinite loop can be triggered.  The loop is in the body of `sentinel_2pc::controller::send_compact_tx` (src/uhs/twophase/sentinel_2pc/controller.cpp, lines 194-203):

        // TODO: add a "retry" error response to offload sentinels from this
        //       infinite retry responsibility.
        while(!m_coordinator_client.execute_transaction(ctx, cb)) {
            // TODO: the network currently doesn't provide a callback for
            //       reconnection events so we have to sleep here to
            //       prevent a needless spin. Instead, add such a callback
            //       or queue to the network to remove this sleep.
            static constexpr auto retry_delay = std::chrono::milliseconds(100);
            std::this_thread::sleep_for(retry_delay);
        };

When the coordinator client fails to start but an error is not raised, calls to `m_coordinator_client.execute_transaction` will return false and the loop will execute indefinitely.  The author apparently recognized this possibility, but there’s currently no code to log error or debug messages and gracefully exit the loop.   

Triggering the infinite loop is easy:  just provide a junk coordinator IP address (e.g. “abcdefg”) for the unit tests in controller_test.cpp.  Specifically, replace lines 24-25 in `SetUp` in controller_test.cpp with the following:
```
        const auto coordinator_ep
            = std::make_pair("abcdefghijklmnopqrstuvwxyz", m_coordinator_port);
```

How should this be addressed?  It seems reasonable to just replace the call to `m_net.cluster_connect(m_server_endpoints, false)` with `m_net.cluster_connect(m_server_endpoints, true)`, which would allow it to return false if the cluster connection fails.  This would allow `sentinel_2pc::controller::init` to fail if the coordinator client fails to start and would preclude triggering the infinite loop.  

### Code of Conduct

mszulcz-mitre:
GitHub [Issue #131](https://github.com/mit-dci/opencbdc-tx/issues/131) describes a bug that makes it impossible to trigger the error "Failed to start coordinator client" in `sentinel_2pc::controller::init` (src/uhs/twophase/sentinel_2pc/controller.cpp, lines 38-41).  This pull request contains a commit that fixes the bug using the method described in Issue #131.  The fix causes the unit test `tcp_rpc_test.send_fail_test` to fail, so the next commit in this pull request fixes it by testing for behavior that's expected after the bug fix.  The final commit in this pull request adds a new unit test that triggers the error that was previously impossible due to the bug.

### DIFF
--- a/src/util/rpc/tcp_client.hpp
+++ b/src/util/rpc/tcp_client.hpp
@@ -55,7 +55,7 @@ namespace cbdc::rpc {
         /// starts the response handler thread.
         /// \return true.
         [[nodiscard]] auto init() -> bool {
-            if(!m_net.cluster_connect(m_server_endpoints, false)) {
+            if(!m_net.cluster_connect(m_server_endpoints, true)) {
                 return false;
             }
 
--- a/tests/unit/coordinator/controller_test.cpp
+++ b/tests/unit/coordinator/controller_test.cpp
@@ -4,6 +4,7 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "uhs/twophase/coordinator/controller.hpp"
+#include "uhs/twophase/locking_shard/controller.hpp"
 #include "util.hpp"
 
 #include <gtest/gtest.h>
@@ -55,13 +56,4 @@ TEST_F(coordinator_controller_test, out_of_range_node_id) {
                                                           m_opts,
                                                           m_logger);
     ASSERT_FALSE(m_ctl_coordinator->init());
-}
-
-TEST_F(coordinator_controller_test, successful_init) {
-    m_ctl_coordinator
-        = std::make_unique<cbdc::coordinator::controller>(0,
-                                                          0,
-                                                          m_opts,
-                                                          m_logger);
-    ASSERT_TRUE(m_ctl_coordinator->init());
 }
--- a/tests/unit/rpc/tcp_test.cpp
+++ b/tests/unit/rpc/tcp_test.cpp
@@ -126,7 +126,7 @@ TEST(tcp_rpc_test, send_fail_test) {
     auto client = cbdc::rpc::tcp_client<request, response>(
         {{cbdc::network::localhost, 55555},
          {cbdc::network::localhost, 55556}});
-    ASSERT_TRUE(client.init());
+    ASSERT_FALSE(client.init());
 
     auto req = request{0};
     auto resp = client.call(req);
--- a/tests/unit/sentinel_2pc/controller_test.cpp
+++ b/tests/unit/sentinel_2pc/controller_test.cpp
@@ -169,3 +169,60 @@ TEST_F(sentinel_2pc_test, tx_validation_test) {
           });
     ASSERT_TRUE(res);
 }
+
+TEST_F(sentinel_2pc_test, bad_coordinator_endpoint) {
+    // Replace the valid coordinator endpoint defined in the fixture
+    // with an invalid endpoint.
+    m_opts.m_coordinator_endpoints.clear();
+    const auto bad_coordinator_ep
+        = std::make_pair("abcdefg", m_coordinator_port);
+    m_opts.m_coordinator_endpoints.resize(1);
+    m_opts.m_coordinator_endpoints[0].push_back(bad_coordinator_ep);
+
+    // Initialize a new controller with the invalid coordinator endpoint.
+    auto ctl = std::make_unique<cbdc::sentinel_2pc::controller>(0,
+                                                                m_opts,
+                                                                m_logger);
+
+    // Check that the controller with the invalid coordinator endpoint
+    // fails to initialize correctly.
+    ASSERT_FALSE(ctl->init());
+}
+
+TEST_F(sentinel_2pc_test, bad_sentinel_client_endpoint) {
+    // Test that a sentinel client fails to initialize
+    // when given a bad endpoint.
+    constexpr auto bad_endpoint = std::make_pair("abcdefg", m_sentinel_port);
+    const std::vector<cbdc::network::endpoint_t> bad_endpoints{bad_endpoint};
+    auto client = cbdc::sentinel::rpc::client(bad_endpoints, m_logger);
+    ASSERT_FALSE(client.init());
+
+    // Test that the controller fails to initialize when given a bad endpoint
+    // for a sentinel client.
+    m_opts.m_sentinel_endpoints.emplace_back(bad_endpoint);
+    auto ctl = std::make_unique<cbdc::sentinel_2pc::controller>(0,
+                                                                m_opts,
+                                                                m_logger);
+    ASSERT_FALSE(ctl->init());
+}
+
+TEST_F(sentinel_2pc_test, bad_rpc_server_endpoint) {
+    // The sentinel endpoint defined below (which corresponds to sentinel_id
+    // also defined below) is used by the sentinel_2pc controller to initialize
+    // an rpc server.  Replacing the valid endpoint defined in the fixture with
+    // an invalid endpoint should cause the rpc server to fail to initialize.
+    m_opts.m_sentinel_endpoints.clear();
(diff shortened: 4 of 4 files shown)


src/util/rpc/tcp_client.hpp
```diff
@@ -55,7 +55,7 @@ namespace cbdc::rpc {
         /// starts the response handler thread.
         /// \return true.
         [[nodiscard]] auto init() -> bool {
-            if(!m_net.cluster_connect(m_server_endpoints, false)) {
```
ykaravas, line 58: I saw this line a few weeks ago and i was banging my head trying to figure out what i was missing!

mszulcz-mitre:
@HalosGhost The unit and integration tests pass when I run them in Docker on my machine but are failing here.  Here's the last few lines from the output:
```
/home/runner/work/opencbdc-tx/opencbdc-tx/tests/integration/atomizer_end_to_end_test.cpp:58: Failure
Value of: m_ctl_watchtower->get_block_height() > 0
  Actual: false
Expected: true
scripts/test.sh: line 9:  8659 Aborted                 (core dumped) $PWD/$1
Error: Process completed with exit code 134.
```
This appears to be the same problem that @davebryson encountered in PR #110.  In [your response](https://github.com/mit-dci/opencbdc-tx/pull/110#issuecomment-1146376666) to the problem in that PR, you mentioned that you ran the tests again and they passed.  Could you please give that a shot?

HalosGhost:
@mszulcz-mitre I pulled the changes locally and needed to add the below patch to get the code to compile; however, once I did everything built and tested fine. I cannot imagine this is what was causing the issue for CI (if so, what a weird way for it to fail); but please go ahead and merge in these missing includes. Regardless, after that's merged, I'll dive in and we'll see how the CI feels about it.

```patch
diff --git a/src/uhs/twophase/coordinator/interface.hpp b/src/uhs/twophase/coordinator/interface.hpp
index 291cf2be..79a5276f 100644
--- a/src/uhs/twophase/coordinator/interface.hpp
+++ b/src/uhs/twophase/coordinator/interface.hpp
@@ -6,6 +6,8 @@
 #ifndef OPENCBDC_TX_SRC_COORDINATOR_INTERFACE_H_
 #define OPENCBDC_TX_SRC_COORDINATOR_INTERFACE_H_
 
+#include <functional>
+
 #include "uhs/transaction/transaction.hpp"
 
 namespace cbdc::coordinator {
diff --git a/src/util/common/logging.hpp b/src/util/common/logging.hpp
index 8b12f376..ff1dcc0c 100644
--- a/src/util/common/logging.hpp
+++ b/src/util/common/logging.hpp
@@ -13,6 +13,7 @@
 #include <mutex>
 #include <optional>
 #include <sstream>
+#include <memory>
 
 namespace cbdc::logging {
     /// No-op stream destination for log output.
```

mszulcz-mitre:
@HalosGhost Thanks for doing that!  

I'm surprised the patches were needed to build the code.  Did I mess something up?  The code without them built on my machine and it seems to have passed the GitHub build check without them too.  And none of my commits touched the patched files.

At any rate, I'm more than happy to add the patches.  When I do, though, running `scripts/lint.sh` raises errors about them:
```shell
mszulczewski@sa-dev:~/Code/OpenCBDC/opencbdc-tx$ ./scripts/lint.sh 
Linting...
src/uhs/twophase/coordinator/interface.hpp:9:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <functional>
^
src/util/common/logging.hpp:8:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <chrono>
^
```

I copied and pasted the patches from your message, but to confirm that nothing got messed up, here's my output from git diff.  Aside from the lines that start with 'index' , it matches your git diff output exactly (I diff'ed the diffs).
```diff
mszulczewski@sa-dev:~/Code/OpenCBDC/opencbdc-tx$ git diff
diff --git a/src/uhs/twophase/coordinator/interface.hpp b/src/uhs/twophase/coordinator/interface.hpp
index 291cf2b..79a5276 100644
--- a/src/uhs/twophase/coordinator/interface.hpp
+++ b/src/uhs/twophase/coordinator/interface.hpp
@@ -6,6 +6,8 @@
 #ifndef OPENCBDC_TX_SRC_COORDINATOR_INTERFACE_H_
 #define OPENCBDC_TX_SRC_COORDINATOR_INTERFACE_H_
 
+#include <functional>
+
 #include "uhs/transaction/transaction.hpp"
 
 namespace cbdc::coordinator {
diff --git a/src/util/common/logging.hpp b/src/util/common/logging.hpp
index 8b12f37..ff1dcc0 100644
--- a/src/util/common/logging.hpp
+++ b/src/util/common/logging.hpp
@@ -13,6 +13,7 @@
 #include <mutex>
 #include <optional>
 #include <sstream>
+#include <memory>
 
 namespace cbdc::logging {
     /// No-op stream destination for log output.
```

I've been scratching my head about the format errors, but I can't figure out the problem.  Any thoughts?

HalosGhost:
@mszulcz-mitre this is one of those odd things about compilation environments that I genuinely don't understand. It seems that my local environment is a little more picky (in *some* circumstances) about `#include` directives than others. Additionally, I made a mistake and added those two includes, but didn't `clang-format` them; if you run `clang-format` on those files, that will fix whatever formatting issue (probably include ordering) it's seeing.

As for whether you did anything wrong, I seriously doubt it; just a compiler idiosyncrasy.

metalicjames (approved):

HalosGhost:
@mszulcz-mitre I think we're all good to merge. Can you squash to a single commit, and I'll merge it.

HalosGhost:
@mszulcz-mitre sorry to have gotten back to this quite a while later, but can you please modify the commit subject line to be a little more clear? (“Fix bug” doesn't really convey anything when people are looking through the commit list.)

HalosGhost:
@mszulcz-mitre I do think this is ready-to-merge, but I'd much prefer a more descriptive commit heading than “Fix bug”. Would you mind updating that message when you get the chance? (e.g., “Allow tcp_client::init to fail correctly”)

mszulcz-mitre:
@HalosGhost No problem!  I updated the commit to use your more descriptive commit message.

HalosGhost:
All set and ready. Merging.

PR #137, opened 2022-06-22
Config file Auto-generation tool

### LINKED ISSUE #136: Craft More Intricate Config Files
This issue references a task in https://github.com/mit-dci/opencbdc-tx/issues/3, Test Coverage / Quality Umbrella; specifically, related to the fourteenth issue down the list which reads: 

" Craft a config that can be used as a basis for most tests"

ykaravas:
This pull request closes #136

Added configuration file autogeneration tool. I think there is still a lot that could be done here and there is potential for a lot more customization of tests/runs. The conversation is ongoing, however, this is a good spot to review and maybe even merge as it has basic functionality and is stable.

Moved configuration files around repo to their own directory, opencbdc-tx/config, for organizational purposes. Updated related scripts and docker yml files with new location of configs. This config dir houses an integration dir which holds integration test configs, unit dir, which holds unit test configs, a general dir, which holds the configs which were formerly in the project root dir and a tools dir which holds the config generator.

Created simple unit test for config generation as well which could be expanded a bit.


ykaravas:
By the way, i am not sure what is causing this or if it is normal, but when i run the atomizer docker (on trunk or on this branch) it seems to fail:

image

2pc docker, all unit tests and integration tests seem to run just fine though.

Issue #119, opened 2022-06-02
Dependencies and timing issues while initializing servers in 2PC architecture causing integration test failure


AlexRamRam:
### Affected Branch

trunk

### Basic Diagnostics

- [X] I've pulled the latest changes on the affected branch and the issue is still present.

- [ ] The issue is reproducible in docker

### Description

Currently, `SetUp()` in `tests/integration/two_phase_end_to_end_test.cpp` contains a number of `sleep(100 msecs)` statements. The purpose is to guard against failures that may be caused by timing dependencies among servers.  Despite these sleep intervals, the test fails intermittently if run with verbose logging on a slow terminal (e.g. a shell within an Emacs editor) which magnifies timing differences among threads. If the sleep duration is increased (e.g. 200 milliseconds), the likely of failure lessens significantly.

This Issue item asks whether dependencies in order of initialization of the component servers in the 2PC architecture exist by design. If so, client needs guidance on how to avoid failures due to these dependencies.

In order to reproduce the issue, follow these steps:
1. Set logging level to `trace` to create more stdout/stderr output
2. Run the following test in a slow terminal (emacs shell, for example) repeatedly until failure is observed
```
tests/integration/run_integration_tests --gtest_filter="two_phase_end_to_end_test.*"
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #138, opened 2022-06-22
Config file Auto-generation tool

### LINKED ISSUE #136: Craft More Intricate Config Files

ykaravas:
This pull request closes #136

Added configuration file autogeneration tool. I think there is still a lot that could be done here and there is potential for a lot more customization of tests/runs. The conversation is ongoing, however, this is a good spot to review and maybe even merge as it has basic functionality and is stable.

Moved configuration files around repo to their own directory, opencbdc-tx/config, for organizational purposes. Updated related scripts and docker yml files with new location of configs. This config dir houses an integration dir which holds integration test configs, unit dir, which holds unit test configs, a general dir, which holds the configs which were formerly in the project root dir and a tools dir which holds the config generator.

Created simple unit test for config generation as well which could be expanded a bit.


config/tools/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+project(config_gen)
+
+include_directories(../../src ../../config/tools ../../3rdparty/secp256k1/include)
```
HalosGhost: I believe `../../config/tools` is this directory; it probably should not be included here.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+#include <cassert>
+#include <filesystem>
+#include <functional>
+#include <random>
+#include <set>
+
+// NORMAL CONFIGS
+
+// Integer that determines if this is 2PC or Atomizer. 1 means 2PC
+static constexpr auto two_phase_mode = "2pc";
```
HalosGhost: This comment doesn't match the next line.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+// Number of sentinels to be created
+static constexpr auto sentinel_count_key = "sentinel_count";
+// Number of coordinators to be created
+static constexpr auto coordinator_count_key = "coordinator_count";
+// Number of archivers to be created
+static constexpr auto archiver_count_key = "archiver_count";
+// Number of atomizers to be created
+static constexpr auto atomizer_count_key = "atomizer_count";
+// Number of watchtowers to be created
+static constexpr auto watchtower_count_key = "watchtower_count";
```
HalosGhost: Can these be reused from the included `config.hpp` rather than redefined?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+            m_current_port++;
+        }
+        return port;
+    }
+
+    [[nodiscard]] auto config_generator::create_repeatable_key_pair()
+        -> std::pair<std::string, std::string> {
+        cbdc::privkey_t seckey;
+        for(auto&& b : seckey) {
+            b = rand();
```
HalosGhost: `rand()` is definitely more repeatable than using some of the more modern tooling (like you use in `create_random_key_pair()`), but I'm unclear on how this is repeatable.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+                std::chrono::system_clock::now().time_since_epoch().count());
+        }
+        std::normal_distribution<double> distribution(mean, std_dev);
+        return distribution(generator);
+    }
+
+    // This is not a failsafe because we are simply generating a
+    // configuration file here, however, it is still good to check that the
+    // port is available
+    [[nodiscard]] auto config_generator::get_open_port() -> unsigned short {
```
HalosGhost: This is a very interesting idea. We mulled over a similar idea in a task in #3 (`Add support for OS-assigned ephemeral port numbers and leverage them in-tests`), but this is a nice middle-ground.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+        -> std::variant<size_t, double, std::string> {
+        auto it = config_map.find(option);
+        if(it != config_map.end()) {
+            value_t parsed_val = parse_value(it->second, false);
+            if(std::holds_alternative<size_t>(parsed_val)) {
+                return std::get<size_t>(parsed_val);
+            } else if(std::holds_alternative<double>(parsed_val)) {
+                return std::get<double>(parsed_val);
+            } else if(std::holds_alternative<std::string>(parsed_val)) {
+                return std::get<std::string>(parsed_val);
```
HalosGhost: Would this be cleaner as a type-matching [`visit`or](https://en.cppreference.com/w/cpp/utility/variant/visit)?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+            } else {
+                std::string error_msg
+                    = "Warning: Unrecognized type for param, " + option + ".";
+                std::cout << error_msg << std::endl;
+                return error_msg;
```
HalosGhost: This looks impossible to reach. `parse_value()` returns a `value_t` which doesn't contain any inhabitants other than `size_t`, `double`, and `std::string`.

Perhaps a reasonable time to use `__builtin_unreachable`?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+                                                    const double value) {
+        m_new_config << key << "=" << value << '\n';
+    }
+
+    void config_generator::set_log_level(const std::string key,
+                                         std::string& log_level) {
+        if(find_value(key, log_level) == false) {
+            find_value(tmpl_universal_override_log_level, log_level);
+        }
+        if(log_levels.find(log_level) == log_levels.end()) {
```
HalosGhost: This check to see if the passed-in `log_level` is valid happens after we pass it through to `find_value()`; Is there a reason we shouldn't check first?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+        std::ofstream outFile;
+        outFile.open(_config_file);
+        outFile << m_new_config.str();
+        outFile.close();
+    }
+
+    // This method assumes project root is "opencbdc-tx" and build dir is
+    // "build"
+    [[nodiscard]] auto
+    config_generator::copy_to_build_dir(const std::string filename) -> bool {
```
HalosGhost: I'm not sure this is a good idea. The build-system supports using build directories other than `build`, so that assumption isn't stable.

Maybe it would make sense to keep this, but add an optional parameter to specify the build directory (defaulting that parameter to `build`)?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+                    set_param_to_config_file(config_map_it->first,
+                                             std::get<double>(parsed_val));
+                } else if(std::holds_alternative<std::string>(parsed_val)) {
+                    set_param_to_config_file(
+                        config_map_it->first,
+                        std::get<std::string>(parsed_val));
+                } else {
+                    return_msg += "Warning: Unrecognized type for param, "
+                                + config_map_it->first + ". \n";
+                }
```
HalosGhost: See comments above. (I'm not sure if this would actually be cleaner with a type-matching visitor, but it does look like the `else` case here is also unreachable.)

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+                                         std::get<size_t>(sentinel_count));
+                }
+            }
+            if(std::holds_alternative<size_t>(shard_count)) {
+                if(std::get<size_t>(shard_count) == 0) {
+                    return_msg
+                        += "Warning: Two-phase mode requires at least one "
+                           "configured shard. Fix configuration template "
+                           "and "
+                           "rerun.\n";
```
HalosGhost:
```suggestion
                           "and rerun.\n";
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,897 @@
...
+                        std::get<size_t>(archiver_count));
+                }
+            }
+            if(std::holds_alternative<size_t>(sentinel_count)) {
+                if(std::get<size_t>(sentinel_count) == 0) {
+                    return_msg
+                        += "Warning: Sentinels require at least one "
+                           "configured shard. Fix configuration template "
+                           "and "
+                           "rerun. \n";
```
HalosGhost:
```suggestion
                           "and rerun. \n";
```

config/tools/config_generator.hpp
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2022 MIT Digital Currency Initiative,
+//                    Federal Reserve Bank of Boston
+//                    MITRE Corporation
```
HalosGhost: It's probably time for us to split out the Copyright holders into an `AUTHORS` file. It's fine to leave these lines, but I'll work on splitting them out in the next little while.

config/tools/generate_configd.cpp
```diff
@@ -0,0 +1,43 @@
...
+//                    MITRE Corporation
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "config_generator.hpp"
+
+#include <iostream>
+#include <random>
+
+auto main(int argc, char** argv) -> int {
```
HalosGhost: This doesn't seem to be a daemon; what does the `d` in the filename refer to?

docker-compose-2pc.yml
```diff
@@ -7,7 +7,7 @@ services:
     image: opencbdc-tx
     tty: true
     restart: always
-    command: ./build/src/uhs/twophase/sentinel_2pc/sentineld-2pc 2pc-compose.cfg 0
+    command: ./build/src/uhs/twophase/sentinel_2pc/sentineld-2pc ./config/general/2pc-compose.cfg 0
```
HalosGhost, line 10: For all the `yml` files, this is a needed change and good; but it looks like this PR doesn't currently update the README's walk-through. We should make sure the documentation is fully-updated as well since the config paths will all change.

HalosGhost (changes requested):
This looks like a solid first crack. I have a few requests inline and a couple of questions.

config/tools/CMakeLists.txt
ykaravas: Yup! removing it.

DONE

config/tools/config_generator.cpp
ykaravas: Clarified this

DONE

config/tools/config_generator.cpp
ykaravas: Yes. However, as we had discussed in the issue related to this, i was hoping to eventually make the config stuff in config.cpp a class so i can utilize a lot more of the functionality from there. But i think we said that could be in a later version right? For now though, i changed this to reuse the ones in src/common/config.hpp.

DONE

config/tools/config_generator.cpp
ykaravas: It is repeatable depending on the template configuration. If you dont want it to be repeatable, if you look around line 92 (in the constructor), you will see that it will use srand to set the seed with a value related to the current time. If not set to randomize, srand is not called and the default value is used and is thus repeatable from run to run.

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: Fixed this.

DONE

config/tools/config_generator.cpp
ykaravas: Fixed this.

DONE

config/tools/generate_configd.cpp
ykaravas: Misunderstood the convention. Removing the d.

DONE

config/tools/config_generator.cpp
HalosGhost: Indeed, on all counts.

config/tools/config_generator.cpp
ykaravas: Most Likely yes! I had not used visit or variant prior to this project so I am still learning about them. Thanks!

However, If i use std::visit the way it is used in the link you provided, then there doesn't seem to be much difference between the way it is currently being done and the way it is done in the "type-matching" section in the example.

In fact, as i have explored std::visit, i am wondering if using it is even necessary at all here. I could just use parse_value and return it if it is found in the config_map

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: I simply removed this else statement

config/tools/config_generator.cpp
ykaravas: Good point. I reversed the order.

DONE

config/tools/config_generator.cpp
ykaravas: Just seeing this now after exploring using visit for a while. I agree with your assessment here. I don't think using a type-matching visit is any better. I will leave both as they are but remove the unreachable else statement. I figured its good practice to always have an else just in case, say, the parse_value code changes at some point. 

Using __builtin_unreachable() in unreachable else statement for now. 

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: After seeing the second place where this was discussed, i decided to leave code as is. Not sure what to do in place of the else, though, since it creates branch where function ends without return value. Using __builtin_unreachable() in unreachable else statement for now. 

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: Used:

 else {
                __builtin_unreachable();
            }

here. Is that correct usage for it? Or do i also need to return something?

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: Fixed this assumption wherever it is made according to directions given above. Tested with different build dir names as well as with build dir nested several dirs deep.

DONE

docker-compose-2pc.yml
ykaravas: Updated all the config file paths in the README.md. Thought about adding a section in the README regarding this new config file generation capability, but i think that may be premature at this moment since there isn't exactly a way to utilize the new config files. So adding it may confuse people before there is an actual integration test that utilizes the more complicated generated config files. What do you think?

NEEDS FOLLOW UP

ykaravas:
@HalosGhost I believe i have addressed all your addressable comments made in this PR. The ones that were very straightforward i have marked "DONE". Others that i either have questions about or have implemented but should probably be reviewed again i have marked as "NEEDS FOLLOW UP".

I have consolidated all my changes in one commit but have not squashed that commit into the original commit for this PR yet so that you can more easily view the diff.

config/tools/config_generator.cpp
HalosGhost: Ahh, I see. It's depending on `rand()` behaving as if `srand(1)` had been called. That's reasonable (and specified by the standard), but escaped me in-review. I guess my next question is why call `srand()` ever then? There's only one other place this code uses `rand()` and it doesn't look like it has any real need to be particularly random (if it does, replacing the call to `rand()` with the more modern tooling is the better option in my opinion).

config/tools/config_generator.cpp
HalosGhost: No need to return anything. That intrinsic serves two basic purposes:

* Tells the compiler to avoid diagnostic messages that it might otherwise offer
* Enables the compiler to produce higher-quality assembly by having an assurance that it can elide code paths that would get to that statement.

config/tools/config_generator.cpp
ykaravas: Well there is a template configuration parameter, tmpl_randomize_values, that allows one to create either a repeatable config file (every time it is generated, if nothing changes), or one that is random. Now, at the moment, the only parameters that end up being random are the shard coverage of each shard and the key pairs but i could foresee more in the future. Of course, the point of having it be repeatable is for testing and debugging purposes. But the point of having it possibly be random, also is to be able to generate and run, possibly, 10's or 100's of similar, yet different, scenarios and observe differences in behavior. Eventually, i see this configuration generation expanding quite a bit to be able to create very intricate scenarios which could help with micro-benchmarking and benchmarking in general or simply for identifying corner cases by virtue of the fact that it can create randomized but similar configuration files. 

My next step is to make the src/util/common/config.hpp config related stuff a class so i can create another class that inherits from it. This will allow me to easily read in the extended and more intricate configuration files that this creates (and i have a few more parameters in mind as well that i am thinking of adding). Then i can begin creating an integration test that utilizes these new configuration files. The parameter that determines whether or not to randomize the configuration file can also be used for randomizing the execution of the test as well so that transactions and their senders and recipients are different per run. I think this could also be useful.

And to clarify, srand is called when we want to generate a randomized scenario. Take a look at lines 150-152:

        if(randomize) {
            srand(std::chrono::system_clock::now().time_since_epoch().count());
        }

However, you are proposing, instead of using rand() to use a more modern tool? What would you suggest using?

HalosGhost: I was thinking of [`uniform_int_distribution`](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution) and friends that you use elsewhere already (and we use liberally throughout the codebase).

Note that `rand()` is unacceptable to use for any case where cryptographic randomness is necessary (and, frankly, lots of other cases as well); but if we don't need any real guarantees and just want pseudo-random-ish numbers, `rand()` is fine. I'm fine with leaving this as-is for now.

ykaravas: Got it. That makes sense yes. But on the other hand, this is just for creating configuration files for tests. I can change it out though no problem if you want.

ykaravas:
@HalosGhost This is just a general question. I am not sure where to post something like that. Would it be Zulip? Anyways, my question is, do you see, in your environment, any integration tests sporadically failing? I never used to see that occurring but recently, atomizer double_spend or atomizer invalid transaction sporadically fail. This occurs on all branches. I had not seen this until this week. I am just curious if this is a common occurrence or not.

HalosGhost:
> @HalosGhost This is just a general question. I am not sure where to post something like that. Would it be Zulip?

Zulip is a great catch-all for if you're not sure where to post something. However, there is a `feedback/question` tag for issues if it feels technically-relevant enough (as this does to me).

> […] do you see, in your environment, any integration tests sporadically failing? I never used to see that occurring but recently, atomizer double_spend or atomizer invalid transaction sporadically fail. This occurs on all branches. I had not seen this until this week. I am just curious if this is a common occurrence or not.

We are aware of several cases where they are more likely to fail now than they used to be (cf. #119). However, even if this feels more pronounced now, it's an extension of the fact that the test suite is a bit fragile at the moment (e.g., relying on `sleep()` calls hoping that the task has finished rather than a less racy solution).

I've also seen this a little more frequently in the GitHub CI of-late, but I think the real solution is a general overhaul of how some of the test suite is written/architected rather than any change (e.g., increasing the sleep times).

config/tools/config_generator.hpp
HalosGhost: cf. #147

ykaravas:
@HalosGhost Hey Sam, are there any blockers on merging this? I am beginning work on a subsequent branch to create a test to use these files as well as further organize what i have done.

HalosGhost:
There is a conflict to resolve, but it should be minor. Also, I must apologize for the delay in-review. My attention has been elsewhere; I will try to take a look at it this weekend!

Also, please feel free to stack PRs if you're going to work on a follow-up branch (just base the new branch on this one, and reference this PR in the PR text to make clear this needs to be merged first).

ykaravas:
@HalosGhost Completely understand about the delays! I am sure you have a very full plate so no worries! I just wanted to make sure you werent waiting on me thats all. I did not notice the conflict. I will resolve that soon. I have branched off this for my continuation yes. Thanks for the tips!

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+                template_file_is_valid = true;
+                std::cout << "Successfully copied " << template_filename
+                          << " from " << temp_config_tools_dir.string()
+                          << " to " << temp_build_config_tools_dir.string()
+                          << ". Using "
+                          << full_template_path_and_filename.string()
+                          << " as template file." << std::endl;
+            }
+        }
+        generator.seed(100);
```
HalosGhost: `100` feels a bit magical to me here. I assume this `seed()` call is made to default to a reproducible-config?

If so, instead of threading through `randomize` as an argument to many of the methods included in this PR, can `config_generator` have a `m_randomize` member, and then conditionally `generator.seed()` either a random-value or `1` (what `rand()` behaves as-if it had been seeded with if no explicit seeding was done)?

It probably also makes sense to move the `srand()` call to be in the same place.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+    }
+
+    // Get random value based on mean and standard deviation
+    [[nodiscard]] auto
+    config_generator::calculate_normal_distribution_point(const size_t mean,
+                                                          const double std_dev,
+                                                          const bool randomize)
+        -> double {
+        if(randomize) {
+            generator.seed(
```
HalosGhost: This is probably not a good idea, if called multiple times (this method is currently called in a `for`-loop), this will re-seed the generator. If that's done within the same second, that might cause some unintended repetition of values.

When not aiming for cryptographic-security, it is generally good practice to seed an RNG only once (at the beginning of its use, or at the beginning of the execution).[^1] See my comment above for a recommendation for how to handle this elegantly and hopefully simplify a few other places in this PR.

[^1]: I am not a cryptographer and this isn't relevant for this particular case, so I will keep this quite general: when aiming for cryptographic security, there *might* be times re-seeding a generator *is* useful (I defer to cryptographers on when/where/why), but it should always be done very carefully (and probably never using a reasonably-predictable seed like UNIX epoch time).

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+            temp_config_tools_dir.append("config").append("tools");
+            std::filesystem::path temp_build_config_tools_dir = m_build_dir;
+            temp_build_config_tools_dir.append("config").append("tools");
+            std::cout << "Warning: File provided, " << m_template_config_file
+                      << ", does not exist. Attempting to copy it from its "
+                         "original location, "
+                      << temp_config_tools_dir.string() << " to "
+                      << temp_build_config_tools_dir.string() << std::endl;
+            copy_templates_to_build_dir();
+            // Try to use newly copied tmplate files
```
HalosGhost:
```suggestion
            // Try to use newly copied template files
```
Might make sense to run `aspell` or something else to catch any simple typos.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+        outFile.open(_config_file);
+        outFile << m_new_config.str();
+        outFile.close();
+    }
+
+    [[nodiscard]] auto
+    config_generator::copy_to_build_dir(const std::string filename) -> bool {
+        std::filesystem::path cwd = std::filesystem::current_path();
+        cwd.append(filename);
+        if(std::filesystem::exists(filename)) {
```
HalosGhost: stylistic point: `if (cond) { return true; } else { return false; }` is usually a sign something can be simplified. In this case, the clue is that one branch doesn't take any action (meaning it's an avenue for an early-return).

I'd recommend negating the condition and rephrasing this as follows:

```c++
if(!std::filesystem::exists(filename)) {
    return false;
}

// no need for an `else` here, the rest of the code can avoid the extra level of indentation
[snip]
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+                                      copyOptions);
+                std::cout << "Copying " << dir_entry.path().string() << " to "
+                          << build_config_dir.string() << std::endl;
+            }
+        }
+    }
+
+    template<typename T>
+    [[maybe_unused]] auto config_generator::find_value(const std::string key,
+                                                       T& output) -> bool {
```
HalosGhost: Why do you prefer an out-parameter rather than a `std::optional<T>` return value?

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+    // Method to create all the Atomizer related components for generated
+    // config file
+    void config_generator::create_atomizer_component(const char* type,
+                                                     const size_t number) {
+        size_t randomize_int = 0;
+        find_value(tmpl_randomize_values, randomize_int);
+        bool randomize = randomize_int == 1;
+        auto _default_debug_level = static_cast<std::string>("INFO");
+        find_value(tmpl_default_log_level, _default_debug_level);
+        if(type == shard_count_key) {
```
HalosGhost: This type of structure is repeated often. Is it possible to refactor this?

My immediate suggestion for how would be to have a method for each component type allowing the body of the for-loop to be a single method call. At that point, it may be possible to refactor a bit further and share the for-loop (dispatching to the routine in the loop body). This also has the added benefit that, should a new type of component be added to the system, ostensibly only the component-specific routine may need to be added to include support.

That will create a few additional helper methods which aren't likely to be called anywhere else, but this is currently pretty hard for me to parse.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,935 @@
...
+        const auto sentinel_count
+            = get_param_from_template_file(sentinel_count_key, config_map);
+
+        if(std::holds_alternative<size_t>(is_two_phase_mode)
+           && std::get<size_t>(is_two_phase_mode) == 1) {
+            output_filename = "2pc_generated_config.cfg";
+            const auto coordinator_count
+                = get_param_from_template_file(coordinator_count_key,
+                                               config_map);
+            if(std::holds_alternative<size_t>(sentinel_count)) {
```
HalosGhost: This is another process that's repeated a lot. Does it seem reasonable to refactor this (if only for maintainability)?

config/tools/config_generator.cpp
HalosGhost: I'm still not in-love with this block, but everything I can think of to make it a little more accessible comes with more down-sides than I'd like. Let's call this one fine for now.

HalosGhost (changes requested):
Sorry this review took quite a while, but I tried to be quite thorough. I do think there are a few things that need to change before merge; most of them are hopefully simplifications and not making anything more complex.

config/tools/config_generator.cpp
ykaravas: Ok. I have implemented this as you have described, however, whether or not to randomize has to be determined in the generate_configuration_file after the template file has been read in. The above was simply setting a default seed.

DONE

config/tools/config_generator.cpp
ykaravas: This was removed as a result of the above comment.

DONE

config/tools/config_generator.cpp
ykaravas: Fixed this. Added Spellcheck plugin to my IDE.

DONE

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: std::optional is a new concept for me with this project. Will try and change it the way i think you are suggesting though. However, i suppose, i am not sure what i would return in the instances where it was not found. 

Having a little trouble understanding why the return value should be std::optional<T> as opposed to just T?

UPDATE: I think i updated it the way you have suggested here. Please check to make sure

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: I understand the first part of the refactor you are suggesting i believe, however, i am not sure what you mean by the second part (how to share the for loop). I will try and think about it more and see if i can understand though. 

UPDATE: Implemented first part of your refactor comment. Still unsure what you mean with regards to second part though after thinking about it for a bit

NEEDS FOLLOW UP

config/tools/config_generator.cpp
HalosGhost: No worries.

The benefit of `optional` is that it allows you to signal the potential absence of a value. In this case, then, using `optional` lets you signal the equivalent of the current `return false` cases (e.g., unknown template parameter type).

If we wanted to be more type-safe/efficiently-serializable (though, the benefits of doing so in this case might be negligible), we might define an error-status type for this function (or related functions), and return a `std::variant<T, error_status_type>` to let us specifically report what went wrong (allowing much more efficient serialization and dispatch); that would offer something a little closer to what exceptions might offer while still avoiding a lot of the drawbacks/overhead. We do something similar in the validation code so that we can serialize only an error code in messages rather than a full string which has to be parsed on the other side.

For this case though, just using an `optional` is probably enough. Hope that's helpful context! If you need more clarity on my question (note: I'm not necessarily against the out-param, though usually my default is to avoid side-effects), just let me know.

config/tools/config_generator.cpp
HalosGhost: Let me clarify: right now, this code is structured like this:

```c++
if (cond) {
    for(...)
} else if (cond) {
    for(...)
}...
```

The `for`-bodies look like they share a lot of the same structure, which makes it feel like there's potentially a more widely-applicable pattern that would make it easier to maintain this code (by serving as a more direct abstraction). After that simplification, it might make sense to make another reversal:

```c++
for(...) {
    if (cond) {
    } else if (cond) {
    }
}
```

Depending on the type of conditions, this might be equally as or more performant. However, in our case, it's actually likely to be a performance regression (because our conditions are often string-comparison, which isn't likely to have nice low-level performance). But, this isn't the system (where high-performance is a focus), this is just surrounding tooling. All else held equal, specifically in the tooling, I'd prefer to see the code be easier-to-maintain.

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: hmmm. I am still having trouble understanding how the for loop can be shared though since there can be different component numbers for each type of component. You could have a different number of shards or sentinels etc... How can you loop through different numbers for different components in the same for loop? Each for loop in the first block may have different number of iterations. I'm just having trouble seeing how you can successfully reverse the order if that is the case.

config/tools/config_generator.cpp
ykaravas: Ended up figuring this out i believe.

DONE

config/tools/config_generator.cpp
HalosGhost: After your refactor, the code looks like this:

```c++
helper1 (count) {
    for (i=0,count)
}

helper2 (count) {
    for (i=0,count)
}

make_components (type, count) {
    if (type == shard) {
        helper1(count)
    } else if ...
}
```

My thought was to have this structure:

```c++
helper1 (index) {
    // routine to generate configs for a single component of a single role
}

helper2 (index) {
    // routine to generate configs for a single component of a single role
}

make_components (type, count) {
    for (i=0,count) {
        if (type == shard) {
            helper1(i)
        } else if ...
    }
}
```

I think that's reasonable (`make_components()` is playing the same role as `create_atomizer_component()` and `create_2pc_component()` in the actual code—which may also be unifiable).

I hope this doesn't feel like busy-work; I think this would be a very helpful tool to have, I'm just hoping to have it be easy for us to maintain.

I'm not entirely convinced that the above refactor is particularly worth it, so think on it a bit before you spend time on implementing it (and send me a PM on Zulip if you want to think through it live or want further clarification).

ykaravas: Ok. I think i have implemented what you suggest above (actually slightly differently actually but very similar).  However, in doing so, i came across some confusion i had had and now i remembered. For the 2pc config, this double count for shards and coordinators in the original integration test config still confuses me:

image


I just dont understand why there is an overall count, and then seemingly, a count per shard and coordinator for 2pc. What does the count per shard and coordinator represent? Is it the same count as the general one?

HalosGhost: Those are raft-replica configs. So, `shard_count` is the total number of logical shards, and `shard0_count` is the number of replicas that will be created in the raft consensus pool backing the first logical shard. Same goes for coordinators.

ykaravas: Ok, I am pretty sure i had tested making those numbers different from the ones without a "0" or number in the name and the tests seemed to fail. I will test again. However, that being said, should not the atomizer have such counts too for the atomizer? I do not see such a count in the original atomizer configuration files? There is a total "atomizer_count" but in that case is that the raft replicated number? 

Also, for the 2pc integration tests, making the shard0_count or coordinator0_count does indeed cause the test to cash FYI. 

Whatever the case, i have now changed to config generation to take this replication count into account as its own thing for those items which require it.

HalosGhost: Where there are multiple logical shards or coordinators, there can only be one logical atomizer, so the count can only refer to the raft replicas. Worth mentioning that this isn't particularly intuitive, and, at some point, some cleaning up of the config format may be warranted to clarify some of these questions. As for increasing the replica count, did you correctly include the new settings necessary for the new replicas to be created appropriately?

ykaravas: There is a small issue with the raft replication params i added that i am currently working through FYI.

ykaravas: Ok. I think i am finished with this comment (and all the comments in this pull request for that matter). Let me know if there is anything else you think needs to change.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+        std::string raft_endpoint_val
+            = cbdc::network::localhost + ":" + std::to_string(get_open_port());
+        set_param_to_config_file(raft_endpoint_key, raft_endpoint_val);
+        std::string log_level_key = atomizer_name + "_loglevel";
+        std::string log_level_val = default_log_level;
+        set_log_level(tmpl_atomizer_log_level, log_level_val);
+        set_param_to_config_file(log_level_key, log_level_val);
+    }
+
+    void config_generator::create_atomizer_whatchtowers(
```
HalosGhost:
```suggestion
    void config_generator::create_atomizer_watchtowers(
```
another small typo.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+            auto _default_log_level
+                = find_value<std::string>(tmpl_default_log_level);
+            if(std::get<size_t>(component_count) == 0) {
+                return_msg += "Warning: 0 count for at least one component. "
+                              "Fix configuration template and rerun.\n";
+            } else if(create_2pc == true) {
+                for(size_t i = 0; i < std::get<size_t>(component_count); i++) {
+                    if(type == shard_count_key) {
+                        // Create Shards
+                        create_2pc_shards(_default_log_level.value(), i);
```
HalosGhost: following this refactor, it looks like most of these method names should no longer be pluralized.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+                      .value();
+        }
+    }
+
+    [[maybe_unused]] auto
+    config_generator::create_component(const char* type,
+                                       const value_t component_count,
+                                       bool create_2pc) -> std::string {
+        std::string return_msg = "";
+        if(std::holds_alternative<size_t>(component_count)) {
```
HalosGhost: I think the point of this is solely to ensure that `component_count` is a `size_t`. If it's not, it returns an empty string as a message. Given that, I'd suggest changing the `component_count` argument to a `size_t` and doing `std::get<>` at the call-site.

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+            auto _default_log_level
+                = find_value<std::string>(tmpl_default_log_level);
+            if(std::get<size_t>(component_count) == 0) {
+                return_msg += "Warning: 0 count for at least one component. "
+                              "Fix configuration template and rerun.\n";
+            } else if(create_2pc == true) {
```
HalosGhost:
```suggestion
            } else if(create_2pc) {
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+            shard_info.at(i).shard_id = i;
+            auto overlap_percentage
+                = find_value<double>(tmpl_avg_shard_start_end_overlap_percent);
+            shard_info.at(i).overlap_percentage_allowed
+                = std::abs(calculate_normal_distribution_point(
+                    0,
+                    overlap_percentage.value()));
+            shard_index_sum_total.at(randNum)++;
+        }
+        bool still_expanding = true;
```
HalosGhost:
```suggestion
        auto still_expanding = true;
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+                    .current_coverage_expansion_limits.second
+                - shard_info.at(shard_id)
+                      .current_coverage_expansion_limits.first)
+               == (array_total.size() - 1))) {
+            // This shard has reached its allowed overlap limit
+            shard_info.at(shard_id).still_expanding = false;
+        }
+        // Check if all indices have been visited at least once from 0 to
+        // shard_size
+        bool all_visited = true;
```
HalosGhost:
```suggestion
        auto all_visited = true;
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+        for(size_t i = 0; i < array_total.size(); i++) {
+            if(array_total.at(i) == 0) {
+                all_visited = false;
+                break;
+            }
+        }
+        // If all indices have been visited at least once from 0 to
+        // shard_size, allow shards to start overlapping in coverage
+        if(all_visited == true) {
```
HalosGhost:
```suggestion
        if(all_visited) {
```

config/tools/config_generator.cpp
```diff
@@ -0,0 +1,896 @@
...
+        if(value[0] != '\"' && value[value.size() - 1] != '\"') {
+            if(value.find('.') == std::string::npos) {
+                const auto as_int = static_cast<size_t>(std::stoull(value));
+                return as_int;
+            }
+            const auto as_dbl = std::stod(value);
+            return as_dbl;
+        }
+
+        if(keep_quotes == false) {
```
HalosGhost:
```suggestion
        if(!keep_quotes) {
```

config/tools/config_generator.hpp
```diff
@@ -0,0 +1,179 @@
...
+        // Helper to set proper log level for either Two-Phase Commit or
+        // Atomizer components
+        void set_log_level(std::string key, std::string& log_level);
+        // Method to create all the related components for generated config
+        // file
+        [[maybe_unused]] auto create_component(const char* type,
+                                               value_t component_count,
+                                               bool create_2pc) -> std::string;
+
+        // Create 2PC shards
```
HalosGhost: Much of these comments are probably not necessary (they mirror the method name exactly), but when adding documentation-comments, adding an additional slash (`///`) will make them visible to doxygen.

config/tools/config_generator.cpp
HalosGhost: There are a lot of places in this code where the variable type is unambiguous and clear based on the rhs of an assignment/initialization. Is there any reason to prefer spelling out the type over `auto`?

CMakeLists.txt
```diff
@@ -75,5 +75,6 @@ endif()
 
 add_subdirectory(src)
 add_subdirectory(tests)
+add_subdirectory(config/tools)
```
HalosGhost: I don't know how I missed this, but why is the configuration-generator in its own directory rather than in `tools/config`? It makes perfect sense (to me) to have the generated configuration files (as well as the ones we ship pre-made) in a unified directory, but we have a `tools` directory specifically to handle ancillary tooling (just like this).

HalosGhost (changes requested):
I do have a few suggested changes (most of them are cosmetic or small questions and so aren't particularly important), but I feel like this PR was to a point where it made sense for me to test it.

I did the following, and waited multiple minutes, but the tool never completed:

```console
$ /build/config/tools/generate_config ./config/tools/2pc_config_template.tmpl 5000 ./build
```

Is that an incorrect invocation? If so, the help output doesn't make it clear to me what I did wrong; thoughts?

ykaravas:
> 

So i usually run it from the build dir but i tried the command you supplied above and it seemed to run just fine for me for both template files.

ykaravas:
@HalosGhost So i usually run it from the build dir but i tried the command you supplied above and it seemed to run just fine for me for both template files. Not sure what the issue could be when you are running it. Will try and look into it somehow but since it is not occurring for me it is hard to debug.

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: Changed all these. Generally i like to explicitly do the == true or == false since it is a lot more legible than the absence or presence of "!" which can be easily missed. But i have changed all these.

DONE

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.cpp
ykaravas: DONE

config/tools/config_generator.hpp
ykaravas: Removed them.

DONE

config/tools/config_generator.cpp
ykaravas: Just for legibility. I usually try not to use auto since it also allows IDEs to more easily parse it and use autocompletion etc... 

Is there any advantage to using auto?

CMakeLists.txt
ykaravas: Well, i created the Config directory to organize the configuration files around the repo. It just wouldn't have made sense, i thought, to have the config files in tools/Config. But to have them in Config and to have the config generator in config/tools seemed to make more sense. Would you rather change it? And if so, do you want the config files in tools/config as well? That would seem a bit odd to me.

NEEDS FOLLOW UP

config/tools/config_generator.cpp
ykaravas: Changed the bools and strings that could be to autos.

DONE

config/tools/config_generator.cpp
ykaravas: So, what happens if it is an empty string though and it gets passed into the method? Wont an exception be hit? 

DONE

config/tools/config_generator.cpp
HalosGhost: > Is there any advantage to using auto?

It's fair to say this is largely a cosmetic stance. But, in my experience, fewer glyphs to parse = fewer places for mistakes to creep in (either in transcription or in comprehension). I also have spent a lot of time in much more heavy-powered type-inferring languages (e.g., Haskell), and the tendency to let the compiler do the book-keeping has definitely worked its way into my practice.

CMakeLists.txt
HalosGhost: Sorry, to clarify, I meant that having `/config` for the configuration files makes perfect sense. But having the configuration-generator live outside `/tools` (which is the specific place for ancillary tooling) doesn't. I'd expect `/tools/config` to have the generator and `/config` to house the pre-made configurations.

Does that feel like a reasonable breakdown?

HalosGhost:
> @HalosGhost So i usually run it from the build dir but i tried the command you supplied above and it seemed to run just fine for me for both template files. Not sure what the issue could be when you are running it. Will try and look into it somehow but since it is not occurring for me it is hard to debug.

I'll debug it a little more thoroughly and see what's causing the issue.

config/tools/config_generator.cpp
HalosGhost: This is also a cosmetic view, but I hold it a lot more strongly.

Personally, I don't find the additional information to be more legible, and the choice to include more doesn't include an obvious stopping point to me (`create_2pc` ≡ `create_2pc == true` ≡ `create_2pc == true == true == true`, etc.). To that end, the only obvious bright-line to draw for me is “comparison operators should be used when you need to convert something into a `bool` to make a decision.” Because `create_2pc` is already a `bool`, there's no need for an additional comparison operator.

config/tools/config_generator.cpp
HalosGhost: Sorry, to clarify, I meant that the signature of the function should be changed to have `component_count` be a `size_t`. At that point, it's impossible to call the function with an empty string as the `component_count`, and it's the call-site's job to decide what to pass and to handle the error locally if the thing it wants to use is not valid.

CMakeLists.txt
ykaravas: That was my original plan. But i remember, i think i had asked you on a call, or somewhere, if it should go in tools and you seemed apprehensive. I will put it there though

HalosGhost: Mmm, my apologies for steering you wrong. But, having the tool sit somewhere under `/tools` and the pre-made configs under `/config` sounds ideal to me.

config/tools/config_generator.cpp
ykaravas: Ok. I did change the function signature to take that as size_t.

DONE

CMakeLists.txt
ykaravas: DONE

ykaravas:
I am suddenly getting a bunch of linting errors that were not there before after moving the files. Not sure why this is happening.

One of them is complaining that i should use a range based loop instead of going over iterators from begin to end. Not sure why that should be done. 

There are a lot of these and if they are valid, i am not sure i will be able to get to them before i go on vacation for a bit.

HalosGhost:
I know why this is happening actually (it's actually a good reason for why the config-generator should be in `tools`); it's because the linter is run on all things in `tools` (but wasn't being run on `/config`). They do look valid to me, and though getting this merged in sooner rather than later is great, we don't have a strict timeline.

ykaravas:
That seems a bit odd to me because I have been having to fix lint issues for all the files while it was in config directory as well. It seemed to be running on the files there too. I would, in fact, get linting errors there also. Why have they suddenly expanded?

HalosGhost:
@ykaravas, hmm, that's interesting. There is a `.clang-tidy` defined at the root of the repo, but I'm unsure of which checks the linter would run underneath `/config` if it didn't define its own set. Technically, `/tools` actually removes one of the checks that's defined at the root level, so I wouldn't expect it to be generating *more* errors if `/config` was indeed getting all the checks from the root level.

ykaravas:
@HalosGhost Sorry for the delay. Got back from vacation September 6. I have finished fixing all the new linting errors and i believe all the comments in this request have been addressed.

HalosGhost:
Seems `./scripts/test.sh` needs to be updated to be made aware of the directory change: https://github.com/mit-dci/opencbdc-tx/actions/runs/3048785772/jobs/4914212231#step:5:1532

ykaravas:
@HalosGhost I received email with your reply regarding scripts.sh but i cannot seem to find where that is here on github anywhere. I could have sworn i had fixed that. Will look into it again.

HalosGhost:
> @HalosGhost I received email with your reply regarding scripts.sh but i cannot seem to find where that is here on github anywhere. I could have sworn i had fixed that. Will look into it again.

That's my fault, I deleted the comment after taking a brief look at your push 16 hours ago (thinking that I was wrong and you had indeed fixed it). I triggered a re-run of the tests CI pipeline to just eliminate all doubt. Either it will fail (and probably in the same place as last time, suggesting the `.tmpl` files were not found in the specified location), or it'll pass. And we'll move forward based on that!

ykaravas:
@HalosGhost  I dont understand why it is failing though. The test.sh script works on my VM at least.

HalosGhost:
The specific error is [here](https://github.com/mit-dci/opencbdc-tx/actions/runs/3048785772/jobs/4924688567#step:5:1532):

```
cp: cannot stat '/home/runner/work/opencbdc-tx/opencbdc-tx/scripts/../config/tools/*.tmpl': No such file or directory
```

In particular, it looks like the culprit is [this line](https://github.com/mit-dci/opencbdc-tx/pull/138/files?diff=unified&w=0#diff-52afcc694c0ff8ff253441f42ff44d6b4734fe30ae9070f2ba07a668a10e7baaR179). Looking at the changed-files list in the PR, `<repo_root>/config/tools/*.tmpl` does not appear to exist.

As a side-note: `cp <wildcard-which-matches-multiple-files> target` is rather dangerous (this predated your PR), but we should probably switch to using the `find…` construction at [line 167](https://github.com/mit-dci/opencbdc-tx/pull/138/files?diff=unified&w=0#diff-52afcc694c0ff8ff253441f42ff44d6b4734fe30ae9070f2ba07a668a10e7baaR167)

HalosGhost:
@ykaravas looks like you have it handled. utACK. Will test locally ASAP (heads-up that the next few days are rather busy on my end so it might be a little while before I can test fully).

HalosGhost:
@ykaravas I've started taking another look at this and I'm getting stumped earlier in the process. Namely, the test suite seems to hang indefinitely for me on the config-test validation. I reran the CI to see if the same issue would occur, but it ran fine. The only change I made was to rebase on `trunk`; can you do a `git pull --rebase upstream trunk` (assumes you have `mit-dci/opencbdc-tx` setup as a remote named `upstream` in your local clone), and see if you run into issues?

As a side-note: please put two returns between your commit header and the body; right now, the full commit message is in the header (which is unreadable).

ykaravas:
@HalosGhost That is odd. I have not been running into any hanging issues. I will test more on my end to see if i do. I have been rebasing all along. I will do that again in case it needs it but i think it is up to date.

And what do you mean about two returns between commit header and body? I usually run the "git commit -s -m "commit message" command. Is there some other way i should do it?

HalosGhost:
@ykaravas I'm so sorry, I accidentally edited your comment rather than replying! I've copied the result of my edit below for clarity, and I returned your comment to its original state. My apologies!

> And what do you mean about two returns between commit header and body? I usually run the "git commit -s -m "commit message" command. Is there some other way i should do it?

That makes sense. It's a little more typical to omit `-m` (which will then open a full editor (by default, usually `vim`, but this is configurable)). In doing so, the first line you enter in the editor corresponds to what you'd pass to `-m` (by convention, kept to <50 characters); hitting return twice (leaving an empty line between the initial line and the rest of the message) then starts the body of the message.

If you look through the commit history, you can see that some committers will expand quite a bit in the message body about why a change was done, or the general motivation. But, when keeping the header to <50 characters, you almost have to because that header is going to be the simplest summary of the change.

ykaravas:
@HalosGhost Have you had a chance to test this out yet?

HalosGhost:
@ykaravas I'm still a little unsure of what's happening. But, the test suite is still hanging indefinitely for me. I'm going to stepwise debug it here in the next couple of days unless you have a thought of what might be going wrong.

ykaravas:
@HalosGhost Hey Sam. I am not sure what could be going on as i cannot recreate that issue. It seems that the unit test is passing on the server even though. So it doesn't seem to be hanging there either as far as i can tell, nor does it on my system. I am not sure how to go about debugging it since i cannot recreate it though. Any suggestions?

HalosGhost:
@ykaravas nothing immediately comes to mind. Stepwise debugging on my side is the next reasonable step; sorry I've not been able to do so earlier!

HalosGhost:
Ah, I've found what's wrong. It's a bit of fragility, but I'm not actually sure of a pleasant way to fix it. My clone is not named `opencbdc-tx`. As a result, the routine to find the root-directory of the repo fails because the parent-directory it's looking for doesn't exist. Thus, that loop looking to find the repo-root runs infinitely.

I'll disregard that for now and plan to use my local CI emulation until/unless we figure out a less-fragile method of detecting the repo root.

HalosGhost:
Note: the way you do this in-shell is actually quite simple:

```console
$ cd "$(git rev-parse --show-toplevel)"
```

HalosGhost:
Confirmed. I've changed my local clone's containing name to make testing easier. ☺

HalosGhost:
@ykaravas I think we should change one more thing. Right now, the config-generator finishes by just saying `SUCCESS` if all went well. It doesn't tell you the name of the config generated nor where that generated config is. I think a really easy fix to this is to instead output the generated configuration on `stdout` (note that `SUCCESS` as well as the other status messages like where the `build` directory has been determined to be should *not* be on `stdout` (which is where they're printed now), but rather `stderr` if we make this change). That way, users can simply redirect the output of the tool to their intended target (naming and location is then up to the user). This might also mean that we can avoid trying to find the project root or build directory (though I'll defer to you on that one).

HalosGhost (changes requested):
As I mentioned above, I think there are two choices for the final change:

1. Move all diagnostic/status messages to `stderr`, and output the generated config on `stdout` (instructing users to redirect the output to their intended name/location)
2. Output the final path of the generated config along with the `SUCCESS` message.

I'm personally a fan of the first, but the second is definitely less work and I'd find it acceptable. The only thing that might strongly tip me in-favor of the first option is if it removes the need to locate the root/build directories. If so, then it can remove a ton of fragility that we'd be better off without.

PR #142, opened 2022-07-08
Build raft cluster from state manager

metalicjames:
This PR simplifies the Raft cluster building logic. Previously we started the initial followers first and subsequently started the leader which joined the followers to the cluster. This method of building the cluster is brittle and prone to transient failures. It also requires starting replicated components in a specific order. Instead, we now start with all the cluster nodes already joined to the cluster. When each node starts, it reconnects to the other cluster nodes, eliminating the startup order requirement. This method of building the cluster is not brittle, but has the tradeoff that the intended initial leader (node ID 0) is not guaranteed to be the leader at the start. To alleviate this, we set node ID 0 to have higher priority in leader elections, increasing the probability it becomes the leader (the election is still randomized, so this cannot be 100% guaranteed).

### DIFF
--- a/src/uhs/atomizer/atomizer/atomizer_raft.cpp
+++ b/src/uhs/atomizer/atomizer/atomizer_raft.cpp
@@ -12,15 +12,15 @@
 #include "util/serialization/util.hpp"
 
 namespace cbdc::atomizer {
-    atomizer_raft::atomizer_raft(uint32_t atomizer_id,
-                                 const network::endpoint_t& raft_endpoint,
-                                 size_t stxo_cache_depth,
-                                 std::shared_ptr<logging::log> logger,
-                                 config::options opts,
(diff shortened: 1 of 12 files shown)


HalosGhost:
Big concept ACK. Will review in-depth ASAP.

HalosGhost:
utACK. Having now looked at the code, I'm really excited to get this merged in (net reduction of over 100 lines of code resulting in a simpler raft-cluster setup, count me in).

There are a couple of small conflicts to resolve, and then I'll give it a local test and we'll be set to merge.

HalosGhost:
cc @metalicjames for conflict resolution.

metalicjames:
@HalosGhost should be good to merge now

HalosGhost (approved):
Tested ACK. This change to simplify raft cluster building looks terrific.

HalosGhost:
Merging.

PR #144, opened 2022-07-08
Fix bug in reconnect logic causing an infinite send loop

metalicjames:
This PR fixes the reconnect logic for peers in the connection manager. Before this PR, a socket disconnect would trigger a reconnect, causing `do_send()` to drop into a tight loop, consuming all the CPU. This was because the `m_wake` flag was never reset, causing `pop()` to always return false after `clear()`. Since the loop calling `pop()` used `continue` rather than `break`, the loop never ended.

The PR provides a method to reset the wakeup flag, and properly breaks out of the send loop when `pop()` returns false.

### DIFF
--- a/src/util/common/blocking_queue.hpp
+++ b/src/util/common/blocking_queue.hpp
@@ -83,6 +83,14 @@ namespace cbdc {
             m_cv.notify_all();
         }
 
+        /// Removes the wakeup flag for consumers. Must be called after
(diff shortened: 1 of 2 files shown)


HalosGhost:
@metalicjames is there a reasonable unit test we could add for this?

HalosGhost (approved):
This is a very simple change to avoid an obvious failure condition.

Looks good to me.

PR #130, opened 2022-06-09
Closes #129. Clarifying the use of wallet string address

dkgaraujo:
Signed-off-by: Douglas Araujo <dkg.araujo@gmail.com>

PR #146, opened 2022-07-10
Clarifies the use of wallet string addresses

dkgaraujo:
Previously discussed in https://github.com/mit-dci/opencbdc-tx/pull/130

Signed-off-by: Douglas Araujo <dkg.araujo@gmail.com>


HalosGhost (approved):
Nice clarification of a couple of the README walkthrough steps. Looks good to me.

PR #149, opened 2022-07-11
Specify random source in connection manager

metalicjames:
This PR explicitly specifies the randomness source to use for seeding the RNG in the connection manager. Previously we used the implementation-defined default source. Now it's set to `cbdc::config::random_source` to be consistent with the rest of the codebase.


HalosGhost (approved):
This small change just keeps the connection-manager's random source consistent with everything else in the code base.

Does what it says on the tin, and looks good to me.

PR #151, opened 2022-07-12
Sentinel 2pc controller bug fixes

### LINKED ISSUE #140: Accessing nonexistent elements in sentinel_2pc::controller
### Affected Branch

trunk

### Basic Diagnostics

### Description

In sentinel_2pc::controller,  a nonexistent element in the vector `m_opts.m_sentinel_endpoints` can be accessed if `m_sentinel_id > m_opts.m_sentinel_endpoints.size() - 1`.  The access occurs in Line 51 of src/ush/twophase/sentinel_2pc/controller.cpp:
```c++
if(ep == m_opts.m_sentinel_endpoints[m_sentinel_id]) {
```
and in Line 70:
```c++
m_opts.m_sentinel_endpoints[m_sentinel_id]);
```
Accessing a nonexistent element via the [] operator results in undefined behavior.  In my tests, the 1st access (Line 51) doesn't throw an error, but the 2nd access sometimes throws an error.  To trigger the nonexistent-element access, you can add this unit test to controller_test.cpp:
```c++
TEST_F(sentinel_2pc_test, out_of_range_sentinel_id) {
    constexpr uint32_t bad_sentinel_id = 1; 

    // Add private key for the bad sentinel ID to avoid triggering the error 
    // "No private key specified".
    constexpr auto sentinel_private_key
        = "000000000000000100000000000000000000000000000000000000000000000"
          "1";
    m_opts.m_sentinel_private_keys[bad_sentinel_id]
        = cbdc::hash_from_hex(sentinel_private_key);

    auto ctl = std::make_unique<cbdc::sentinel_2pc::controller>(bad_sentinel_id,
                                                                m_opts,
                                                                m_logger);
    ASSERT_FALSE(ctl->init());
}
```

### Solution
A solution is to check whether `sentinel_id` is out of range of `m_opts.m_sentinel_endpoints` in sentinel_2pc::controller::init, and if it is, indicate that initialization failed by returning false.  Something like this should work:
```c++
if(m_sentinel_id > (m_opts.m_sentinel_endpoints.size()-1)) {
    m_logger->error(“Sentinel ID is out of range of the sentinel endpoints vector”);
    return false;
}
```

### Code of Conduct

### LINKED ISSUE #141: Non-existent element access and infinite loop in sentinel_2pc::controller
### Affected Branch

trunk

### Basic Diagnostics

### Description

# Problems
I’ve observed 2 issues with `sentinel_2pc::controller::gather_attestations`:  a non-existent element access and an infinite loop.

### 1. Non-existent element access
When `m_sentinel_endpoints.size() = 1` and `m_attestation_threshold > 1`, a non-existent element in the vector `m_sentinel_clients` is accessed.   The access results in undefined behavior, but in my tests, a seg. fault is usually thrown.  The access occurs on Line 166 in src/uhs/twophase/sentinel_2pc/controller.cpp (lines 165 - 176 shown below):
```c++
                success
                    = m_sentinel_clients[sentinel_id]->validate_transaction(
                        tx,
                        [=](validate_result v_res) {
                            auto r = requested;
                            r.insert(sentinel_id);
                            validate_result_handler(v_res,
                                                    tx,
                                                    result_callback,
                                                    ctx,
                                                    r);
                        });
```
To reproduce the error, add this test to controller_test.cpp:
```c++
TEST_F(sentinel_2pc_test, nonexistent_element_access_in_gather_attestations) {
    m_opts.m_attestation_threshold = 2;
    auto ctl = std::make_unique<cbdc::sentinel_2pc::controller>(0,
                                                                m_opts,
                                                                m_logger);
    ASSERT_TRUE(ctl->init());

    // The next line causes a non-existent element to be accessed when 
    // execute_transaction calls gather_attestations.
    auto res = ctl->execute_transaction(m_valid_tx, [](auto /* param */) {});
    ASSERT_TRUE(res);
}
```

## 2.  Infinite loop   

When the number of required sentinel attestations is larger than the number of sentinels, an infinite loop can be triggered in `sentinel_2pc::controller:;gather_attestations`. The infinite loop occurs because the condition on Line 158 in src/uhs/twophase/sentinel_2pc/controller.cpp can never be satisfied:
```
 if(ctx.m_attestations.size() < m_opts.m_attestation_threshold) {
```
Sometimes the loop will eventually throw a seg. fault, but I don’t know why yet.   

To reproduce the problem, add this test to controller_test.cpp:
```c++
TEST_F(sentinel_2pc_test, execute_transaction_triggers_infinite_loop) {
    // Require 3 sentinel attestations.  This requirement is impossible to
    // achieve since this test only launches 2 sentinels.
    m_opts.m_attestation_threshold = 3;

    // Define 2nd sentinel controller (the 1st is defined in the fixture).
    unsigned short sentinel_port = 32003;
    auto sentinel_ep
        = std::make_pair(cbdc::network::localhost, sentinel_port);
    m_opts.m_sentinel_endpoints.push_back(sentinel_ep);
    constexpr auto sentinel_private_key
        = "000000000000000100000000000000000000000000000000000000000000000"
            "1";
    constexpr auto sentinel_public_key
        = "eaa649f21f51bdbae7be4ae34ce6e5217a58fdce7f47f9aa7f3b58fa2120e2b"
            "4";
    m_opts.m_sentinel_private_keys[1]
        = cbdc::hash_from_hex(sentinel_private_key);
    m_opts.m_sentinel_public_keys.insert(
        cbdc::hash_from_hex(sentinel_public_key));

    // Initialize the 1st sentinel controller.
    auto ctl = std::make_unique<cbdc::sentinel_2pc::controller>(0,
                                                                m_opts,
                                                                m_logger);
    ASSERT_TRUE(ctl->init());


    // Initialize the 2nd sentinel controller.
    auto ctl2 = std::make_unique<cbdc::sentinel_2pc::controller>(1,
                                                            m_opts,
                                                            m_logger);
    ASSERT_TRUE(ctl2->init());

    // This line will trigger an infinite loop when execute transaction calls
    // gather_attestations.
    auto res = ctl->execute_transaction(m_valid_tx, [](auto /* param */) {});
    ASSERT_TRUE(res);
}

```

# Solutions
To prevent these errors, we could add two bits of code to `sentinel_2pc::controller::init`.

1.  We could prevent accidental casting that occurs when defining the distribution `m_dist` (Line 57 in src/uhs/twophase/sentinel_2pc/controller.cpp):
```c++
m_dist = decltype(m_dist)(0, m_sentinel_clients.size() - 1);
```
When `m_sentinel_clients.size() = 0`, the second argument to `m_dist` above is -1, which is then cast to an unsigned integer.  In my tests, it  becomes 18446744073709551615.  To prevent this, we could replace the current definition of `m_dist` with:
```c++
constexpr size_t dist_lower_bound = 0;
const size_t dist_upper_bound 
	= m_sentinel_clients.empty() ? 0 : m_sentinel_clients.size() - 1;
m_dist = decltype(m_dist)(dist_lower_bound, dist_upper_bound);
```
This ensures that the `sentinel_id` defined on Line 161 in `sentinel_2pc::controller::gather_attestations` remains a reasonable number:
```c++
auto sentinel_id = m_dist(m_rand);
```

2.  We could add the following to `sentinel_2pc::controller::init` to check that the required number of attestations never exceeds the number of sentinels:
```c++
        if(m_opts.m_attestation_threshold
            > m_opts.m_sentinel_endpoints.size()) {
            m_logger->error("The number of required attestations is larger \n \
                than the number of sentinels that can provide them.");
            return false;
        } 
```

### Code of Conduct

mszulcz-mitre:
This PR addresses the bugs identified in Issue #140 and Issue #141.  The bug fixes and associated unit tests are the same as those described in the issues.

### DIFF
--- a/src/uhs/twophase/sentinel_2pc/controller.cpp
+++ b/src/uhs/twophase/sentinel_2pc/controller.cpp
@@ -25,6 +25,12 @@ namespace cbdc::sentinel_2pc {
                                                    .size())]) {}
 
     auto controller::init() -> bool {
+        if(m_sentinel_id >= m_opts.m_sentinel_endpoints.size()) {
+            m_logger->error(
+                "The sentinel ID is too large for the number of sentinels.");
+            return false;
+        }
+
         auto skey = m_opts.m_sentinel_private_keys.find(m_sentinel_id);
         if(skey == m_opts.m_sentinel_private_keys.end()) {
             if(m_opts.m_attestation_threshold > 0) {
--- a/src/uhs/twophase/sentinel_2pc/controller.cpp
+++ b/src/uhs/twophase/sentinel_2pc/controller.cpp
@@ -55,7 +55,10 @@ namespace cbdc::sentinel_2pc {
             m_sentinel_clients.emplace_back(std::move(client));
         }
 
-        m_dist = decltype(m_dist)(0, m_sentinel_clients.size() - 1);
+        constexpr size_t dist_lower_bound = 0;
+        const size_t dist_upper_bound
+            = m_sentinel_clients.empty() ? 0 : m_sentinel_clients.size() - 1;
+        m_dist = decltype(m_dist)(dist_lower_bound, dist_upper_bound);
 
         auto rpc_server = std::make_unique<cbdc::rpc::tcp_server<
             cbdc::rpc::async_server<cbdc::sentinel::request,
--- a/src/uhs/twophase/sentinel_2pc/controller.cpp
+++ b/src/uhs/twophase/sentinel_2pc/controller.cpp
@@ -25,6 +25,11 @@ namespace cbdc::sentinel_2pc {
                                                    .size())]) {}
 
     auto controller::init() -> bool {
+        if(m_opts.m_sentinel_endpoints.empty()) {
+            m_logger->error("No sentinel endpoints are defined.");
+            return false;
+        }
+
         if(m_sentinel_id >= m_opts.m_sentinel_endpoints.size()) {
             m_logger->error(
                 "The sentinel ID is too large for the number of sentinels.");
--- a/src/util/common/config.cpp
+++ b/src/util/common/config.cpp
@@ -685,6 +685,11 @@ namespace cbdc::config {
                 return "Two-phase mode requires at least one configured "
                        "sentinel";
             }
+            if(opts.m_sentinel_endpoints.size()
+               < opts.m_attestation_threshold) {
+                return "The number of required attestations is larger \n"
+                       "than the number of sentinels that can provide them.";
+            }
             if(opts.m_locking_shard_endpoints.empty()) {
                 return "Two-phase mode requires at least one configured shard";
             }
--- a/tests/unit/sentinel_2pc/controller_test.cpp
+++ b/tests/unit/sentinel_2pc/controller_test.cpp
@@ -38,6 +38,19 @@ class sentinel_2pc_test : public ::testing::Test {
         m_opts.m_coordinator_endpoints.resize(1);
         m_opts.m_coordinator_endpoints[0].push_back(coordinator_ep);
 
+        // The locking shard endpoint defined below may not be used in tests,
+        // but it must be defined for the options struct to be valid.  Without
+        // it, the check_options function couldn't be used to validate the
+        // other options used in this test.
+        static constexpr unsigned short m_locking_shard_port = 42001;
+        const auto locking_shard_endpoint
+            = std::make_pair(cbdc::network::localhost, m_locking_shard_port);
+        m_opts.m_locking_shard_endpoints.resize(1);
+        m_opts.m_locking_shard_endpoints[0].push_back(locking_shard_endpoint);
+
+        auto opt_chk_result = cbdc::config::check_options(m_opts);
+        ASSERT_FALSE(opt_chk_result.has_value());
+
         m_dummy_coordinator_thread = m_dummy_coordinator_net->start_server(
             coordinator_ep,
             [&](cbdc::network::message_t&& pkt)
--- a/tests/unit/sentinel_2pc/controller_test.cpp
+++ b/tests/unit/sentinel_2pc/controller_test.cpp
@@ -226,3 +226,24 @@ TEST_F(sentinel_2pc_test, bad_rpc_server_endpoint) {
     // Check that the controller with the invalid endpoint fails to initialize.
     ASSERT_FALSE(ctl->init());
 }
+
+TEST_F(sentinel_2pc_test, out_of_range_sentinel_id) {
+    // Test that controller initialization fails when the sentinel ID is
+    // too large for the number of sentinels.  Here, since there's only
+    // 1 sentinel, the only allowable sentinel ID is 0.  However, it's
+    // deliberately set to 1 to trigger failure.
+    constexpr uint32_t bad_sentinel_id = 1;
+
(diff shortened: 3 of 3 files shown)


src/uhs/twophase/sentinel_2pc/controller.cpp
```diff
@@ -25,6 +25,20 @@ namespace cbdc::sentinel_2pc {
                                                    .size())]) {}
 
     auto controller::init() -> bool {
+        if(m_sentinel_id > (m_opts.m_sentinel_endpoints.size() - 1)) {
+            m_logger->error(
+                "The sentinel ID is too large for the number of sentinels.");
+            return false;
+        }
+
+        if(m_opts.m_attestation_threshold
```
metalicjames: I think this check more appropriately belongs in the configuration file parsing code where there are other sanity checks like this.

mszulcz-mitre: Good call.  In the latest commit, I put it with the other sanity checks on the `options` struct in the function `cbdc::config::check_options`.

HalosGhost: For my own clarity, which check got moved?

HalosGhost: Pardon the noise. The way github is displaying this is quite strange.

src/util/common/config.cpp
```diff
@@ -682,6 +682,11 @@ namespace cbdc::config {
                 return "Two-phase mode requires at least one configured "
                        "sentinel";
             }
+            if(opts.m_sentinel_endpoints.size()
```
HalosGhost, line 688: It looks like this commit added this check but didn't remove it from `::init()`; is it valuable having this check twice?

mszulcz-mitre: Nope--it's probably not valuable.  I just forgot about it.  I just refactored the commits to remove it and to hopefully improve the commit messages.

HalosGhost:
@mszulcz-mitre this has a very minimal merge conflict with `trunk` (just the addition of some tests). When you have a moment and can rebase it, I'll give it one final review and we'll get this merged as well!

HalosGhost:
@metalicjames when you have a chance to review this, I'd love a second set of eyes.

src/uhs/twophase/sentinel_2pc/controller.cpp
```diff
@@ -25,6 +25,12 @@ namespace cbdc::sentinel_2pc {
                                                    .size())]) {}
 
     auto controller::init() -> bool {
+        if(m_sentinel_id > (m_opts.m_sentinel_endpoints.size() - 1)) {
```
metalicjames: Does this check still work if `m_sentinel_endpoints.size()` is zero? Does `m_opts.m_sentinel_endpoints.size() - 1` result in underflow, bypassing the check?

metalicjames (commented):
I think we should add a unit test for the potential underflow case, then I think it's good to merge.

mszulcz-mitre:
I modified the code to avoid the possibility of underflow.  For clarity, I added another check to confirm that at least one sentinel endpoint is defined and I added a related unit test too.

metalicjames (approved):

mszulcz-mitre:
@HalosGhost It looks like James approved the changes I made in response to his review.  Would you like me to do anything more for this pull request?

HalosGhost:
Sorry for the delay; I've finally had a chance to walk through the code locally and test it. Tested ACK. All these changes eliminate several failure points and do so minimally. Looks good to me.

Issue #127, opened 2022-06-08
Refactor test suite to support starting components asynchronously


HalosGhost:
From [a discussion](https://github.com/mit-dci/opencbdc-tx/pull/106/files#r887288399) on #106, it became clear that our test suite has a pretty unfortunate reliance on being able to start components synchronously.

In particular, some components (e.g., shards) rely on connecting to another component (e.g., an atomizer) before they can function properly. But, because the components must be started synchronously (right now), `init()` does not actually signal whether or not the component is actually ready to begin normal operation.

This leads to several situations where, despite `init()` returning successfully, the component might silently enter a failed state resulting in the whole test failing.

If the test suite were re-architected to support launching components asynchronously (and suitable modification/addition of a query to check that a component is correctly-initialized), we could instead kick off the initialization of all the components, and sleep till each component responds to the query successfully (offering a someone long timeout so that the suite doesn't hang indefinitely if something goes wrong).

This is also probably better for test-health in the long-run since it more closely mirrors how the system would actually be run.

PR #155, opened 2022-08-04
Fix doc and double save in client

mszulcz-mitre:
This PR contains 2 commits:

### 1. The 1st commit fixes an inaccurate method description

Here's the description of 'client::init':
 https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/client/client.hpp#L37-L42

The implementation shows that `init` does not attempt to create data files if they don't exist:
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/client/client.cpp#L33-L48

The creation of a wallet data file occurs in  `transaction::wallet::save`:
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/transaction/wallet.cpp#L307-L327

The creation of a client data file occurs in `client::save_client_state`:
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/client/client.cpp#L287-L298

### 2. The 2nd commit removes a redundant call
`client::mint` appears to call `save` twice.  As a result, the files written in the 1st call are cleared and the same data is then re-written to them.  The first call to save occurs when `mint` calls `import_transaction`:
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/client/client.cpp#L57-L70
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/client/client.cpp#L198-L201
The 2nd call occurs in `mint` itself (line 67).  This commit deletes the 2nd call.

### DIFF
--- a/src/uhs/client/client.cpp
+++ b/src/uhs/client/client.cpp
@@ -64,8 +64,6 @@ namespace cbdc {
             m_logger->error("Failed to send mint tx");
         }
 
-        save();
-
         return mint_tx;
     }
 
--- a/src/uhs/client/client.hpp
+++ b/src/uhs/client/client.hpp
@@ -36,8 +36,8 @@ namespace cbdc {
 
         /// \brief Initializes the client.
         ///
-        /// Attempts to load the data files, and creates new ones if they do
-        /// not exist. Establishes connections to the system components.
+        /// Attempts to load the data files and
+        /// establish connections to the system components.
         /// \return true if initialization succeeded.
         auto init() -> bool;
 


mszulcz-mitre:
Line 312 in wallet.cpp has a TODO: 
https://github.com/mit-dci/opencbdc-tx/blob/2b21cc9f857cabaf129a0a413a3ed8825dea2f36/src/uhs/transaction/wallet.cpp#L307-L316

Is there any reason not to follow up on it?

mszulcz-mitre:
@HalosGhost For my last few PRs, I've seen tests pass on my machine but fail here, and I've wondered if it's due to known issues in the test suite (e.g. Issue #127) or problems with my commits.  Is there anything I can do to check?  For example, is there a list of tests susceptible to the issues in #127 that I should expect to fail sometimes?  Is there a way I can re-run the tests here (without adding new commits) to see if they pass on a 2nd or 3rd run?

HalosGhost:
> I've seen tests pass on my machine but fail here, and I've wondered if it's due to known issues in the test suite (e.g. Issue #127) or problems with my commits. 

That they pass on re-run (or have in other PRs at least, the current rerun of unit/integration tests for this PR hasn't finished) strongly suggests it's fragility in the test suite rather than in your work.

> Is there anything I can do to check? For example, is there a list of tests susceptible to the issues in #127 that I should expect to fail sometimes? Is there a way I can re-run the tests here (without adding new commits) to see if they pass on a 2nd or 3rd run?

You can certainly use a tool like [`act`](https://github.com/nektos/act) to run the CI locally a few times. We don't have specific documentation of which tests are prone to failure, but a good short-hand is “does the test include a call to `sleep()`?” as that's a significant part of why the test suite is a bit fragile.

As for re-running them, feel free to ping me on Zulip (or here); it's a single button-press for me to re-run one of the CI steps.

----

Yep, CI passed with flying colors on-rerun.

HalosGhost:
So, the second `save()` comes following `send_mint_tx()`. As far as I can tell, this should be safe because the calls under `send_mint_tx()` which make modifications should all call `save()` themselves.

@metalicjames is there a specific reason to keep this (potentially) redundant call to `save()`?

metalicjames (approved):
LGTM

PR #157, opened 2022-08-15
Refactor test suite to support starting components asynchronously

mszulcz-mitre:
This pull request addresses Issue #127.

### DIFF
--- a/src/uhs/atomizer/archiver/controller.cpp
+++ b/src/uhs/atomizer/archiver/controller.cpp
@@ -103,10 +103,12 @@ namespace cbdc::archiver {
 
     auto controller::init_atomizer_connection() -> bool {
         m_atomizer_network.cluster_connect(m_opts.m_atomizer_endpoints, false);
-        if(!m_atomizer_network.connected_to_one()) {
-            m_logger->error("Failed to connect to any atomizers.");
-            return false;
+        while(!m_atomizer_network.connected_to_one()) {
+            // TODO:  should we limit the number of attempts?
+            m_logger->warn("Waiting to connect to atomizers...");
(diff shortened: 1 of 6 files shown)


mszulcz-mitre:
@HalosGhost So far, the only commit in this PR addresses the specific issue we previously discussed:  the inability of the atomizer to wait to sync with watchtowers in `atomizer_end_to_end_test.cpp`.  On my machine, at least, this commit makes the previous call to sleep between initializing the watchtowers and atomizer in the test unnecessary (line 50):
https://github.com/mit-dci/opencbdc-tx/blob/9daab811e54e7a489b715caaaa46aff3734d2566/tests/integration/atomizer_end_to_end_test.cpp#L46-L52

Is this sufficient to fix this test?  I doubt it--but I don't have any way of knowing for certain right now.  Unfortunately, I still can't reproduce the  failures described by Issues #119 and #127.  I've followed the instructions in #119:
- The log level is already set to trace in `two_phase_end_to_end_test.cpp`.
- I launch the tests though a shell in emacs.

I've also tried the following: 
- Running the tests in a docker container to slow things down (launched via emacs).
- Running the tests in a docker container that's restricted to only use 1 CPU core to slow things down even more (launched via emacs).
- Commenting out all the calls to `std::this_thread::sleep_for(std::chrono::milliseconds(100));` (lines 38, 42, 43, 45):
https://github.com/mit-dci/opencbdc-tx/blob/9daab811e54e7a489b715caaaa46aff3734d2566/tests/integration/two_phase_end_to_end_test.cpp#L33-L50

I recall that you may have mentioned some other way to try to trigger the problems.  Could you please remind me what you said?

PR #161, opened 2022-08-16
Add docker multi stage builds

jallen-frb:
- Adds docker multi-stage builds
- Updates CI to utilize multi-stage builds
- Updates docker-compose files to utilize new building commands and CI


jallen-frb:
Failing tests:
- https://github.com/mit-dci/opencbdc-tx/runs/7867404034?check_suite_focus=true
- https://github.com/mit-dci/opencbdc-tx/runs/7867404231?check_suite_focus=true 

This is due to no base image being pushed yet. Merging this will push the base image via CI.

Dockerfile
```diff
@@ -45,6 +55,9 @@ RUN wget https://github.com/eBay/NuRaft/archive/v${NURAFT_VERSION}.tar.gz && \
 # Set working directory
 WORKDIR /opt/tx-processor
 
+# Create Build Image
+FROM $BASE_IMAGE AS builder
```
kylecrawshaw, line 25: Do we need to use the external image since the base is defined above? Could this just be `FROM base as builder` or `FROM $IMAGE_VERSION as builder`?

README.md
```diff
@@ -83,40 +83,77 @@ The easiest way to compile the code and run the system locally is using [Docker]
 
 Don't forget to run the docker daemon!
 
+## Build the containers
+
+Building with Docker utilizes multi-stage builds. In order to run an architecture you will need to build each architecture independently if building locally.
+
+**Note:** We have pre-built images available [here](#launch-the-system-with-a-pre-built-image) if you would rather pull the images from GitHub packages over building them locally.
+
+1. Build 2PC architecture:
```
kylecrawshaw, line 92: Love the separate instructions, but it may be a bit clearer if we have separate sections focused on each architecture instead of separate instructions for each architecture at every step.

README.md
```diff
@@ -83,40 +83,77 @@ The easiest way to compile the code and run the system locally is using [Docker]
...
+    ```terminal
+    $ cd opencbdc-tx                                              # change to the project directory
+    $ sudo -s                                                     # open a root shell (needed for docker)
+    $ docker build --target twophase  -t opencbdc-tx-twophase .   # build the container
```
HalosGhost, line 96:
```suggestion
    # docker build --target twophase  -t opencbdc-tx-twophase .   # build the container
```

HalosGhost: here, and below, use `#` when a command should be run as-root.

README.md
HalosGhost: I don't have strong opinions on this. I think it's reasonable as-is, but @jallen-frb, if you think it will be cleaner as @kylecrawshaw suggests, I think it would be fine to change it as well.

HalosGhost (commented):
utACK.

I really like the multi-stage builds plan; should dramatically reduce the image size necessary for actually running the system, and still keep it very simple to build and test locally.

Note though, this includes #162 (which I have an open review comment for), so this might need a rebase. And I know there are some CI changes to-be-made in order for everything to run as-expected.

Feel free to re-request my review when those are done!

README.md
jallen-frb: I don't think these need to be ran as root.

HalosGhost:
@jallen-frb and I discussed this offline, and the two `docker-build-and-publish-*` CI steps are failing because no base image has been published yet (which should/will resolve itself upon merge of this PR). I'll be considering both of those jobs as `PASS`.

HalosGhost (approved):
Breaks out our docker building into multiple stages (results in the final images being dramatically smaller). Looks fabulous!

PR #162, opened 2022-08-19
Unify `Dockerfile` and `configure.sh`, upgrade to LLVM 14, build LevelDB from source

metalicjames:
This PR:

1. Removes the custom dependency installation instructions from the `Dockerfile`, using `configure.sh` instead.
2. Ensures the docker image builds all dependencies in release mode.
3. Upgrades LLVM to 14, and fixes issues raised by the upgraded `clang-tidy`. Disables the new `readability-identifier-length` check for now, as large parts of the codebase are not compliant (perhaps this should be an issue?). This is needed to support `NOLINTBEGIN` and `NOLINTEND`, which will be used in a subsequent PR.
4. Simplifies `configure.sh` to remove the dependency cache, which I don't think is used in our current iteration of Github actions, or when building using docker.
5. Builds LevelDB from source on all platforms.
6. Fixes `lint.sh` regex to only include files with the `*pp` extension, rather than any file ending in `pp`.

This probably conflicts with #161.

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -26,7 +26,7 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
   CPUS=$(sysctl -n hw.ncpu)
   # ensure development environment is set correctly for clang
   $SUDO xcode-select -switch /Library/Developer/CommandLineTools
-  brew install leveldb llvm@14 googletest lcov make wget cmake
+  brew install llvm@14 googletest lcov make wget cmake
   CLANG_TIDY=/usr/local/bin/clang-tidy
   if [ ! -L "$CLANG_TIDY" ]; then
     $SUDO ln -s $(brew --prefix)/opt/llvm@14/bin/clang-tidy /usr/local/bin/clang-tidy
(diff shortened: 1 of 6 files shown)


HalosGhost:
I really like a lot of the simplifications present in this. @jallen-frb can some/all of these simplifications be folded into (e.g., by rebasing your work on top of these commits) #161?

scripts/configure.sh
```diff
@@ -41,63 +41,53 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
...
 
+LEVELDB_VERSION="1.23"
+echo -e "${green}Building LevelDB from sources...${end}"
+wget https://github.com/google/leveldb/archive/${LEVELDB_VERSION}.tar.gz
+rm -rf "leveldb-${LEVELDB_VERSION}-${CMAKE_BUILD_TYPE}"
+tar xzvf ${LEVELDB_VERSION}.tar.gz
+rm -rf ${LEVELDB_VERSION}.tar.gz
+mv leveldb-${LEVELDB_VERSION} "leveldb-${LEVELDB_VERSION}-${CMAKE_BUILD_TYPE}"
+cd "leveldb-${LEVELDB_VERSION}-${CMAKE_BUILD_TYPE}"
+eval "cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DLEVELDB_BUILD_TESTS=0 -DLEVELDB_BUILD_BENCHMARKS=0 -DBUILD_SHARED_LIBS=0 ."
```
HalosGhost: We had this line before and I suppose it just escaped me previously; why are we using `eval`? As far as I can tell, it shouldn't be needed at all, right?

HalosGhost (commented):
untested ACK.

I only have one question (which isn't specific to a change included in this PR, but rather a question if we can improve a single line of code this this PR touches) made inline.

HalosGhost:
Looks good to me. Will be testing locally shortly.

HalosGhost:
tested ACK. This looks good and behaves well for me. My only question is about whether or not some of the proposed changes from #171 should be included here to minimize breakage potential.

metalicjames:
@HalosGhost I think those changes probably make most sense to be included in #170. Since the snappy linkage issue doesn't seem to occur on Ubuntu or MacOS (which are the current supported platforms).

HalosGhost (approved):
Solid improvement and simplification to the docker instrumentation by leveraging tooling we already have. And, as a bonus, it unifies more dependencies across platforms and updates LLVM.

Does what it says on the tin. Looks good to me.

pr4u4t:
> @HalosGhost I think those changes probably make most sense to be included in #170. Since the snappy linkage issue doesn't seem to occur on Ubuntu or MacOS (which are the current supported platforms).

My build env is Arch Linux and snappy library package doesn't include static in opposite to ubuntu packages.
- [ubuntu file list](https://packages.ubuntu.com/bionic/amd64/libsnappy-dev/filelist)
- [arch linux don't panic](https://archlinux.org/packages/extra/x86_64/snappy/)
This is why linking error occurs.

PR #163, opened 2022-08-23
Make blocking queue generic over the underlying queue type

metalicjames:
This PR refactors `blocking_queue` to be generic over the underlying queue type used to implement it. Previously it always used `std::queue`. This PR introduces an alias for an alternative which instead uses `std::priority_queue`. This is useful when you need a queue shared between multiple consumers and producers, but need to prioritize certain elements.


HalosGhost (approved):
Solid refactor to support more flexibility down the road. Looks good to me.

PR #166, opened 2022-08-30
allow disabling sentinel attestations

jallen-frb:
- fixes ability to set ```attestation_threshold=0```

Signed-off-by: Jonathan Allen <jonathan.allen@bos.frb.org>


HalosGhost (approved):
Really simple change that makes the requirement for specifying a sentinel's signing key optional if no attestations are required. Looks good to me.

PR #168, opened 2022-09-06
Make connection failure fatal only when there is one endpoint

metalicjames:
PR #135 introduced a subtle bug where it was impossible to start a configuration where clusters have more than one endpoint. This is because failing to connect to any endpoint is now fatal. However, only the Raft leader of a cluster listens on its endpoint, so we expect the connections to fail in the normal case.

This PR makes connection failure only fatal when there is one endpoint, meaning the leader should always be listening.

In the future this connection logic should probably be more complex, and aware of whether we're connecting to a Raft cluster, checking that at least one endpoint connected. However, in practice it may be better to deal with this elsewhere to remove the constraint on the startup order of components. Maybe it's okay for failure to be a warning in most components rather than a fatal error.

### DIFF
--- a/src/util/rpc/tcp_client.hpp
+++ b/src/util/rpc/tcp_client.hpp
@@ -53,9 +53,11 @@ namespace cbdc::rpc {
 
         /// Initializes the client. Connects to the server endpoints and
         /// starts the response handler thread.
-        /// \return true.
+        /// \return false if there is only one endpoint and connecting failed.
+        ///         Otherwise true.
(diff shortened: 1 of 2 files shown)


HalosGhost:
cc @kylecrawshaw

wadagso-gertjaap:
Can confirm the issue, fix LGTM.

HalosGhost (approved):
Simple and self-explanatory fix for replicated startup.

PR #169, opened 2022-09-07
Reduce failures to connect from fatal errors to warnings

### LINKED ISSUE #167: Sentinels fail to start coordinator client
### Affected Branch

tunk

### Basic Diagnostics

### Description

When running opencbdc-tx 2pc architecture, if the sentinels are started before coordinators the sentinels fail to start with the error `[ERROR] Failed to start coordinator client`. This occurs whether or not the coordinators and shards are run in a replicated mode. When replication is enabled the coordinators will also fail to start with the error `[FATAL] Failed to initialize shard client`.

This issue is not found in commits before https://github.com/mit-dci/opencbdc-tx/pull/135 was merged.

In order to reproduce the issue, follow these steps:
1. Start a sentinel, coordinator and shard all at the same time or start sentinel first. Sentinels will generally initialize faster than coordinators and shards.
2. Sentinels will crash almost immediately leaving the system in an unrecoverable state.
3. If coordinators and shards are replicated then the coordinators will also crash.

### Code of Conduct

metalicjames:
Follow up from #168. This PR reduces all failures to connect from fatal errors to warnings. This should eliminate the startup order requirement of the system components.

I think I got them all, but it's possible I missed one.

### DIFF
--- a/tests/unit/archiver_test.cpp
+++ b/tests/unit/archiver_test.cpp
@@ -165,12 +165,11 @@ TEST_F(ArchiverTest, archiver_terminate) {
     ASSERT_FALSE(terminating_archiver->running());
 }
 
-// We only test the atomizer initialization in failure, given that
-// for a succesful init we need a running atomizer, which by definition
-// is an integration test, and as such needs to happen in the integration
-// test suite, not here.
+// TODO: To test for a succesful init we need a running atomizer, which by
+// definition is an integration test, and as such this test needs to either be
+// an integration test, or the atomizer needs to be mocked.
 TEST_F(ArchiverTest, archiver_atomizer_init_failure) {
-    ASSERT_FALSE(m_archiver->init_atomizer_connection());
+    ASSERT_TRUE(m_archiver->init_atomizer_connection());
 }
 
 // Test if the archiver properly initializes its server interface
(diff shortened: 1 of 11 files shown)


wadagso-gertjaap:
Without this the 2PC system cannot start up regardless of the order:

The 2PC sentinel creates separate network clients for each of the other sentinels (for the attestation feature) - meaning if connecting to it fails, the sentinel fails to start up.

Combined with the fact that the sentinel doesn't start listening before it has connected to other sentinels means the system can't ever start up.

tests/unit/archiver_test.cpp
```diff
@@ -170,7 +170,7 @@ TEST_F(ArchiverTest, archiver_terminate) {
 // is an integration test, and as such needs to happen in the integration
 // test suite, not here.
 TEST_F(ArchiverTest, archiver_atomizer_init_failure) {
```
HalosGhost: The comment text here is now misleading.

tests/unit/archiver_test.cpp
```diff
@@ -286,7 +286,7 @@ TEST_F(ArchiverTest, init) {
     // init should return false because we can't connect to an atomizer
```
HalosGhost: Same with this one.

HalosGhost (changes requested):
utACK. There are a couple of now-misleading comments that need to be updated.

HalosGhost (approved):
Thank you for fixing those comments. This looks like a really simple change that both closes out the issue introduced by #135 and has a nice improvement in eliminating the not-particularly-well-documented component startup-order requirement.

Looks great to me.

kylecrawshaw:
Thanks for changing this @metalicjames . This fixes the issues I observed in https://github.com/mit-dci/opencbdc-tx/issues/167

kylecrawshaw (approved):

PR #170, opened 2022-09-07
Added Arch Linux support for scripts/configure.sh

pr4u4t:
Added support for Arch Linux in configure script.


HalosGhost:
Concept ACK.

I think we are pretty happy to have more and better support for a variety of platforms (and that includes multiple distros). But, to make sure we can actually support them, I think we need to ensure the CI gets updated (so we can see if something breaks a specific platform right away). To that end, I'd ask this PR include an additional copy of the build-release pipeline using arch as the base rather than ubuntu.

PR #172, opened 2022-09-08
Remove disabling RTTI message from CMake

metalicjames:
The message `Disabling Run Time Type Information (RTTI) features.` was introduced by 3426b09 and is removed in this PR. It's the only custom message we emit from CMake, and gets printed every time you build. The message seems a little unnecessary (and perhaps confusing), so I propose removing it.


HalosGhost:
I don't have strong feelings on this. @dmtrinh, you added this message; is there a compelling reason to keep it around?

HalosGhost:
Hearing nothing and not having strong opinions on it. I'm comfortable merging. 👍

PR #174, opened 2022-09-09
Replace usage of std::cout with the logger

metalicjames:
This PR removes a couple of instances where we still used `std::cout` to directly emit log messages, using the logger instead.


HalosGhost:
untested ACK. This seems wholly reasonable.

HalosGhost (approved):
tested ACK.

These appear to be the only two uses of `std::cout` that aren't either in the logging infrastructure itself or in the command-line interfaces.

Looks great to me.

PR #176, opened 2022-09-10
Moved decode_address from client-cli to client

pr4u4t:
Moved address_decode from client-cli to client since this function may be reused in other units later.

### DIFF
--- a/src/uhs/client/CMakeLists.txt
+++ b/src/uhs/client/CMakeLists.txt
@@ -4,10 +4,10 @@ add_library(client atomizer_client.cpp
                    client.cpp
                    twophase_client.cpp)
 
+target_link_libraries(client bech32)
+
 add_executable(client-cli client-cli.cpp)
 target_link_libraries(client-cli client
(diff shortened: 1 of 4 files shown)


HalosGhost:
> since this function may be reused in other units later.

That's interesting. I have no reason to suggest otherwise, but do you have something in-mind?

pr4u4t:
I'm making a REST interface based on client-cli and decode address is needed there and I don't want to duplicate this code.

HalosGhost:
When you say that it's “based on client-cli,” you mean that your REST interface is shelling out and executing a command-line utility? I would wonder if it would make more sense long-term for us to move `decode_address` and other potentially-helpful methods out to `util` and create a shared library from the utilities so that there's a well-defined entry point for higher layers.

HalosGhost:
All-in-all, this looks largely ready to merge, though there is still a lint error that should be resolved first.

HalosGhost:
via #210

PR #177, opened 2022-09-13
Make cross-sentinel connection failures non-fatal

wadagso-gertjaap:
Follow-up on #135 , #167 and #168

This sequence of updates now causes the connection between sentinels, used for requesting attestations from other sentinels, to silently fail. Even though the `false` return value from `tcp_client::init` is now considered a warning and the sentinel continues running, the connection is not working because the handler thread never gets started. So now, RPC calls from sentinel to sentinel fail at runtime.

The first sentinel in the startup sequence could potentially have no working connections and get stuck in endless retry loops.

Sentinel to sentinel communication is actually a situation where `cluster_connect(endpoints, false)` should be used even with a single endpoint, since it's fine for one or more sentinels to be unreachable temporarily (we'll just use another). But because `m_server_endpoints.size() <= 1` is used as parameter to cluster_connect that behavior is impossible.

The reason sentinel->sentinel communication uses clusters of 1 is because the sentinel has to control which other sentinels are called to prevent getting more than one attestation from the same sentinel. So we can't just build a cluster of all other sentinels and use `send_to_one()`

My suggestion, through this PR, is to add an optional boolean overload to `init()` that allows you to set the `error_fatal` parameter of `cluster_connect` manually.

### DIFF
--- a/src/uhs/sentinel/client.hpp
+++ b/src/uhs/sentinel/client.hpp
@@ -31,8 +31,11 @@ namespace cbdc::sentinel::rpc {
         auto operator=(client&&) -> client& = delete;
 
         /// Initializes the client. Establishes a connection to the sentinel.
+        /// \param error_fatal treat connection errors as fatal. See
+        ///                    tcp_client::init for further explanation.
         /// \return true if initialization succeeded.
-        auto init() -> bool;
+        /// \see \ref cbdc::rpc::tcp_client::init(std::optional<bool>)
+        auto init(std::optional<bool> error_fatal = std::nullopt) -> bool;
 
         /// Result type from execute_transaction.
         using execute_result_type
--- a/src/uhs/sentinel/client.cpp
+++ b/src/uhs/sentinel/client.cpp
@@ -15,8 +15,8 @@ namespace cbdc::sentinel::rpc {
         : m_logger(std::move(logger)),
           m_client(std::move(endpoints)) {}
 
-    auto client::init() -> bool {
-        if(!m_client.init()) {
+    auto client::init(std::optional<bool> error_fatal) -> bool {
+        if(!m_client.init(error_fatal)) {
             m_logger->error("Failed to initialize sentinel RPC client");
             return false;
         }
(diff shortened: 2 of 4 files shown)


pr4u4t:
I've also noticed that +1

src/uhs/sentinel/client.hpp
```diff
@@ -34,6 +34,12 @@ namespace cbdc::sentinel::rpc {
         /// \return true if initialization succeeded.
         auto init() -> bool;
 
+        /// Initializes the client. Establishes a connection to the sentinel.
+        /// \param error_fatal if false, the client will succeed initialization even if
+        /// the connection to the sentinel fails
+        /// \return true if initialization succeeded.
+        auto init(std::optional<bool> error_fatal) -> bool;
```
metalicjames: Why is `error_fatal` in a `std::optional`? Is it just so you don't have to modify other parts of the code, or does it have a semantic meaning when it's `std::nullopt`?

metalicjames (commented):
Concept ACK. Just a question about the implementation.

src/uhs/sentinel/client.hpp
HalosGhost: As a follow-up: if there is semantic meaning for `std::nullopt`, is there a reason to have this split into two overloads rather than defaulting the argument to `std::nullopt`?

wadagso-gertjaap: The semantic meaning to `std::nullopt` is to default back to the current behavior, which is to only treat errors as fatal when you're connecting to a cluster with one member. Otherwise it falls on the caller to implement this default, which means changes everywhere this is used. I have made this explicit in the doc comment.

I agree that I could just make this a default argument  to remove the explicit overload - so I have addressed that.

HalosGhost (approved):
Following your clarification (and the inclusion of that clarification in the documentation—thank you for that!), this looks good to me. Extends `tcp_client::init` to take an optional boolean allowing for a little more fine-grained control over when connection errors should be considered fatal.

@metalicjames I'd love your sign-off before merging in case I missed anything.

metalicjames (approved):
LGTM!

PR #179, opened 2022-09-14
Unify transaction metrics

wadagso-gertjaap:
In two phase commit, throughput and latency samples are collected at the load generator level. This also gives point-in-time data about transaction latency.

In atomizer, throughput samples are collected on the archiver on a block-by-block basis, and latency samples are collected at the load generator, without preserving the point in time - resulting in a plain heap of latency samples.

This PR unifies that metric output by having `atomizer-cli-watchtower` write the same latency sample file and format as `twophase-gen`.

### DIFF
--- a/tools/bench/atomizer-cli-watchtower.cpp
+++ b/tools/bench/atomizer-cli-watchtower.cpp
@@ -161,8 +161,8 @@ auto main(int argc, char** argv) -> int {
     static std::atomic_bool running = true;
     uint64_t best_watchtower_height = 0;
 
-    std::ofstream latency_log("latency_samples_" + std::to_string(cli_id)
-                              + ".txt");
+    std::ofstream latency_log("tx_samples_" + std::to_string(cli_id) + ".txt");
+
     assert(latency_log.good());
 
     std::ofstream utxo_set_log("utxo_set_size_" + std::to_string(cli_id)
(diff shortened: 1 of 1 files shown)


tools/bench/atomizer-cli-watchtower.cpp
```diff
@@ -216,7 +216,7 @@ auto main(int argc, char** argv) -> int {
                             if(tx_it != txs.end()) {
                                 const auto tx_delay
                                     = now - tx_it->second.first;
-                                latency_log << tx_delay << "\n";
+                                latency_log << now << " " << tx_delay << "\n";
```
HalosGhost, line 219: Confirmed matching [`twophase-gen`'s line 263](https://github.com/mit-dci/opencbdc-tx/blob/trunk/tools/bench/twophase_gen.cpp#L263)

tools/bench/atomizer-cli-watchtower.cpp
```diff
@@ -161,8 +161,8 @@ auto main(int argc, char** argv) -> int {
     static std::atomic_bool running = true;
     uint64_t best_watchtower_height = 0;
 
-    std::ofstream latency_log("latency_samples_" + std::to_string(cli_id)
-                              + ".txt");
+    std::ofstream latency_log("tx_samples_" + std::to_string(cli_id) + ".txt");
```
HalosGhost, line 164: Confirmed matching [`twophase-gen`'s line 136](https://github.com/mit-dci/opencbdc-tx/blob/trunk/tools/bench/twophase_gen.cpp#L136)

HalosGhost (approved):
This looks like a super slim change to bring the atomizer load-generator's latency-sampling output in-line with 2PC's.

Looks great to me.

PR #181, opened 2022-09-15
Use logging::log and remove use of std::cerr in raft::node

### LINKED ISSUE #180: Raft node logging
### Affected Branch

trunk

### Basic Diagnostics

## Problem

Pull request #174 replaced the usage of std::cout with a logger in class `raft::node`.  Here's the changes:
image

In my tests, the logger calls don't write logs.  When I step through the calls in a debugger, they resolve to the `info` method in the class `nuraft::logger`, which is deprecated and has an empty body.  Here's the relevant parts of the class from the NuRaft repository (https://github.com/eBay/NuRaft/blob/0a084b65531a7004a48df445a0bf532b626c01db/include/libnuraft/logger.hxx#L24-L44):
image

## Potential Solution
Instead of calling the `info` method, call the `put_details` method.  This method wraps calls to logging methods in `cbdc::logging::log` such as `info`:
https://github.com/mit-dci/opencbdc-tx/blob/95c25e28ecb2dd05e209506337921e52af37cae3/src/util/raft/console_logger.cpp#L12-L58  
This method appears to not be used yet in the source code, but it is used in a unit test:
https://github.com/mit-dci/opencbdc-tx/blob/95c25e28ecb2dd05e209506337921e52af37cae3/tests/unit/raft_test.cpp#L674-L718

## Another consideration
Many parts of the code call the `error` method in class `logging::log` to log error messages (grep returns more than 50 instances):
https://github.com/mit-dci/opencbdc-tx/blob/95c25e28ecb2dd05e209506337921e52af37cae3/src/util/common/logging.hpp#L106-L111
`raft::node`, however, uses `std::cerr`:
https://github.com/mit-dci/opencbdc-tx/blob/95c25e28ecb2dd05e209506337921e52af37cae3/src/util/raft/node.cpp#L53-L56
It's not clear to me why `std::cerr` is used in some parts of the code (about 30 instances) and the logger is used in other parts, but in this case at least, the `put_details` method could be used to replace the call to `std::cerr`.

### Code of Conduct

metalicjames:
Fixes #180.

This PR removes the use of `std::cerr` in `raft::node`. It also replaces usage of the NuRaft console logger, instead using our own logger class.

### DIFF
--- a/src/util/raft/node.cpp
+++ b/src/util/raft/node.cpp
@@ -17,14 +17,15 @@ namespace cbdc::raft {
         : m_node_id(static_cast<uint32_t>(node_id)),
           m_blocking(blocking),
           m_port(raft_endpoints[m_node_id].second),
-          m_raft_logger(nuraft::cs_new<console_logger>(std::move(logger))),
+          m_raft_logger(nuraft::cs_new<console_logger>(logger)),
           m_smgr(nuraft::cs_new<state_manager>(
               static_cast<uint32_t>(m_node_id + 1),
               node_type + "_raft_log_" + std::to_string(m_node_id),
               node_type + "_raft_config_" + std::to_string(m_node_id) + ".dat",
               node_type + "_raft_state_" + std::to_string(m_node_id) + ".dat",
               std::move(raft_endpoints))),
-          m_sm(std::move(sm)) {
+          m_sm(std::move(sm)),
+          m_log(std::move(logger)) {
         m_asio_opt.thread_pool_size_ = asio_thread_pool_size;
         m_init_opts.raft_callback_ = std::move(raft_cb);
         if(m_node_id != 0) {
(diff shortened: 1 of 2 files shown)


HalosGhost:
@mszulcz-mitre would love a review from you on this to make sure the issue is solved on your end.

mszulcz-mitre:
@HalosGhost I'd be happy to review.  It looks good at first glance, but I'll build and test to make sure.

mszulcz-mitre:
@HalosGhost I wasn't able to test exhaustively, but the tests I did run looked good.

mszulcz-mitre (approved):

HalosGhost (approved):
Very simple removal of `std::cerr` in favor of the internal logger, and making it possible for nuraft to use our logger internally, so it is all unified.

tested ACK, and looks good to me.

PR #182, opened 2022-09-16
3PC architecture: EVM and Lua smart contract support

metalicjames:
## Intro

This PR contributes a new architecture to OpenCBDC currently named "3PC" (due to an additional phase over 2PC). Unlike the two existing architectures (atomizer and 2PC) which assumed a UTXO model compatible with a UHS, 3PC makes no assumptions about the underlying data model or transaction semantics. Instead, 3PC implements a strongly-consistent, fault-tolerant, concurrency controlled, generic key-value store. Specific transaction logic and data representation is implemented by pluggable modules called "runners" which only affect a single component of the overall system. This makes the 3PC architecture extremely flexible for rapid prototyping, supporting arbitrary programmability, transaction logic or data models without having to modify core system components. It can even support multiple transaction formats simultaneously, including a mixture of UTXO and account-based data models. It would be fairly trivial to implement a runner which executes the transaction format used in 2PC/Atomizer.

There are two runners provided in this PR. The first is a fairly generic function execution environment using the programming language Lua (called "3PC/Lua"). The transaction semantics and data model are implemented _as Lua contracts_ and thus can be modified while the system is running (and transactions are in-flight) without modifying any C++ code. An example contract is provided with implements a simple account-based transaction format and data model. A benchmarking tool and wallet is also provided as a usage example.

The second runner implements the Ethereum Virtual Machine (EVM) and can execute raw Ethereum transactions verbatim (called "3PC/EVM"). It provides an HTTP JSON-RPC server which mimics the Infura/Geth RPC interface, enabling Ethereum tooling to be used out-of-the-box in many cases with 3PC/EVM. We provide a benchmarking tool for testing base token transfers and transactions calling an ERC20 contract. Unlike in Ethereum, transactions in 3PC/EVM execute in parallel rather than being sequenced by blocks, but 3PC enforces concurrency control resulting in equivalent transaction semantics. Transactions are guaranteed to complete eventually if there are no system failures, even when multiple transactions conflict with each other.

## Design

Similarly to Atomizer and 2PC, data is partitioned into shards. In 3PC, shards support multiple rounds of locking for a single transaction using two-phase locking, and shards enforce concurrency control by making transactions queue for locks. Each transaction is assigned a monotonically increasing "ticket number" by a replicated service called the ticket machine. The shards use the ticket numbers to implement the "wound-wait" deadlock avoidance algorithm which prioritizes fairness over efficiency. Transactions are coordinated between shards by an "agent" which contains both a transaction coordinator (similar to 2PC) and the contract runner itself. Unlike in 2PC, the coordinator/agent is not replicated, only shards and the ticket machine are raft replicated services. Instead, we introduce an additional commit step to the shards to enable agent recovery after a failure. Similarly to 2PC and unlike Atomizer, 3PC can scale horizontally by adding more shards and agents. Tickets are issued by the ticket machine in batches, the size of which can be increased significantly to alleviate the bottleneck at the ticket machine.

## Performance

We benchmarked 3PC/EVM replicated in three AWS regions (Virginia, Ohio, Oregon). We used the example ERC20 contract from OpenZeppelin and our benchmark repeatedly sent between paired accounts in the contract. With 128 shards and 64 agents we were able to achieve a peak throughput of ~140K TX/s with 99.999% of transactions finalized in under 4 seconds.

## Future work

- Experiment with and understand how lock contention affects 3PC performance
- Address various TODOs in the code, some of these are related to things like Raft snapshots (which are not yet implemented), and others are just ideas for future improvements
- More tests (system, integration and unit)
- More documentation/usage examples
- More runner implementations (different contract languages, transaction formats etc)
- Interoperability between runners
- Shard data rebalancing (using a directory service managed by the 3PC protocol)


jallen-frb:
There is a bug in the pipeline right now @HalosGhost . Basically on PR, the PR uses the cached base image even if it has changed. We are working on a fix.

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+        Json::Value params,
+        const server_type::result_callback_type& callback,
+        evm_log_query& qry) -> bool {
+        auto ret = Json::Value();
+
+        if(params[0]["blockhash"].isString()) {
+            auto maybe_block_num
+                = uint256be_from_hex(params[0]["blockhash"].asString());
+            if(!maybe_block_num) {
+                m_log->warn("Invalid blockNumber / hash parameter");
```
metalicjames, line 633: Return error to user using `ret` as below

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+        } else if(params[0]["fromBlock"].isString()
+                  && params[0]["toBlock"].isString()) {
+            auto highest_ticket_number = m_broker->highest_ticket();
+            if(params[0]["fromBlock"].asString() == "latest") {
+                qry.m_from_block = highest_ticket_number;
+            } else {
+                auto maybe_block_num
+                    = uint256be_from_hex(params[0]["fromBlock"].asString());
+                if(!maybe_block_num) {
+                    m_log->warn("Invalid fromBlock parameter");
```
metalicjames, line 647: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
 
-        if(!params.isArray()) {
-            m_log->warn("Parameter is not an array");
+            if(params[0]["toBlock"].asString() == "latest") {
+                qry.m_to_block = highest_ticket_number;
+            } else {
+                auto maybe_block_num
+                    = uint256be_from_hex(params[0]["toBlock"].asString());
+                if(!maybe_block_num) {
+                    m_log->warn("Invalid toBlock parameter");
```
metalicjames, line 659: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+            }
+        }
+        callback(ret);
+    }
+
+    auto http_server::handle_get_logs(
+        Json::Value params,
+        const server_type::result_callback_type& callback) -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isObject()) {
+            m_log->warn("Invalid parameters to getLogs");
```
metalicjames, line 780: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+                                       qry,
+                                       std::move(res));
+            });
+    }
+
+    auto http_server::handle_get_transaction_receipt(
+        Json::Value params,
+        const server_type::result_callback_type& callback) -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isString()) {
+            m_log->warn("Invalid parameters to getTransactionReceipt");
```
metalicjames, line 809: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
                 callback(ret);
-                return;
+            });
+    }
+
+    auto http_server::handle_get_code(
+        Json::Value params,
+        const server_type::result_callback_type& callback) -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isString()) {
+            m_log->warn("Invalid parameters to getBalance");
```
metalicjames, line 858: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+            return false;
+        }
+
+        auto params_str = params[0].asString();
+        auto maybe_runner_params
+            = cbdc::buffer::from_hex(params_str.substr(2));
+        if(!maybe_runner_params.has_value()) {
+            m_log->warn("Unable to decode params", params_str);
```
metalicjames, line 866: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+        }
+
+        cbdc::buffer runner_params;
+        if(params[0].asString() == "latest") {
+            runner_params = cbdc::make_buffer(
+                evmc::uint256be(m_broker->highest_ticket()));
+        } else {
+            auto maybe_block_num = uint256be_from_hex(params[0].asString());
+            if(!maybe_block_num) {
+                m_log->warn("Invalid blockNumber / hash parameter");
```
metalicjames, line 927: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+    }
+
+    auto http_server::fetch_block(
+        Json::Value params,
+        const server_type::result_callback_type& callback,
+        const std::function<void(interface::exec_return_type, cbdc::buffer)>&
+            res_cb) -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isString()
+           || (params.size() > 1 && !params[1].isBool())) {
+            m_log->warn("Invalid parameters to getBlock", params.size());
```
metalicjames, line 916: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+        callback(ret);
+        return true;
+    }
+
+    auto http_server::handle_get_block_tx(
+        Json::Value params,
+        const server_type::result_callback_type& callback) -> bool {
+        if(!params.isArray() || params.size() < 2 || !params[0].isString()
+           || !params[1].isString()) {
+            m_log->warn("Invalid parameters to "
```
metalicjames, line 1120: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+        callback(ret);
+        return true;
+    }
+
+    auto
+    http_server::handle_call(Json::Value params,
+                             const server_type::result_callback_type& callback)
+        -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isObject()) {
+            m_log->warn("Parameter to call is invalid");
```
metalicjames, line 1228: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+            return false;
+        }
+
+        auto maybe_tx = dryrun_tx_from_json(params[0]);
+        if(!maybe_tx) {
+            m_log->warn("Parameter is not a valid transaction");
```
metalicjames, line 1234: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+                           ret["result"] = "0x" + buf.to_hex();
+                           callback(ret);
+                       });
+    }
+
+    auto http_server::handle_send_transaction(
+        Json::Value params,
+        const server_type::result_callback_type& callback) -> bool {
+        if(!params.isArray() || params.empty() || !params[0].isObject()) {
+            m_log->warn("Invalid parameters to sendTransaction");
```
metalicjames, line 1284: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+            return false;
+        }
+
+        auto maybe_tx = tx_from_json(params[0]);
+        if(!maybe_tx) {
+            m_log->warn("Parameter is not a valid transaction");
```
metalicjames, line 1290: Return error to user

src/3pc/agent/runners/evm/http_server.cpp
```diff
@@ -49,142 +50,1284 @@ namespace cbdc::threepc::agent::rpc {
...
+    auto http_server::handle_accounts(
+        const Json::Value& /*params*/,
+        const server_type::result_callback_type& callback) -> bool {
+        auto ret = Json::Value();
+        ret["result"] = Json::Value(Json::arrayValue);
+        callback(ret);
+        return true;
+    }
+
+    auto http_server::handle_estimate_gas(
```
metalicjames, line 1194: Move to static handler

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -108,10 +122,224 @@ namespace cbdc::threepc::agent::runner {
...
+                    ret[m_param] = make_buffer(blk);
+                    m_result_callback(ret);
+                    return;
+                }
+
+                auto tx_success = m_try_lock_callback(
+                    v,
+                    broker::lock_type::read,
+                    [this,
+                     tn](const broker::interface::try_lock_return_type& res2) {
```
metalicjames, line 184: Split these nested lambdas into their own methods

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -108,10 +122,224 @@ namespace cbdc::threepc::agent::runner {
...
+
+        auto v = std::get<broker::value_type>(res);
+        auto maybe_logs = cbdc::from_buffer<evm_log_index>(v);
+        if(maybe_logs) {
+            // Found potentially relevant logs, add
+            std::unique_lock<std::mutex> lck(*log_indexes_mut);
+            log_indexes->push_back(maybe_logs.value());
+        }
+        auto acq = (*acquired)++;
+        if(acq + 1 == key_count) {
```
metalicjames, line 294: `if(++(*acquired) == key_count)`

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -108,10 +122,224 @@ namespace cbdc::threepc::agent::runner {
...
+        const evm_log_query& qry,
+        const std::shared_ptr<std::mutex>& log_indexes_mut,
+        const std::shared_ptr<std::vector<evm_log_index>>& log_indexes) {
+        m_log->info(m_ticket_number,
+                    "completed all queries, filtering",
+                    log_indexes->size(),
+                    "logs");
+
+        // Scanned them all - finish
+        std::unique_lock<std::mutex> lck(*log_indexes_mut);
```
metalicjames, line 309: Scope this lock to the protected data only

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -347,15 +575,44 @@ namespace cbdc::threepc::agent::runner {
...
+        }
+        auto acquired = std::make_shared<std::atomic<size_t>>();
+        for(auto& key : keys) {
+            auto success = m_try_lock_callback(
+                key,
+                broker::lock_type::write,
+                [acquired, callback, key_count = keys.size()](
+                    const broker::interface::try_lock_return_type&) {
+                    auto acq = (*acquired)++;
+                    if(acq + 1 == key_count) {
```
metalicjames, line 607: `if(++(*acquired) == key_count)`

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -368,10 +625,14 @@ namespace cbdc::threepc::agent::runner {
...
         auto v = std::get<broker::value_type>(res);
         auto from_acc = evm_account();
+
+        // TODO: Start at zero?
+        // auto exp_nonce = evmc::uint256be(0);
         if(v.size() > 0) {
             auto maybe_from_acc = cbdc::from_buffer<evm_account>(v);
             if(maybe_from_acc.has_value()) {
                 from_acc = maybe_from_acc.value();
+                // exp_nonce = from_acc.m_nonce + evmc::uint256be(1);
```
metalicjames, line 635: Remove comments

metalicjames (commented):
We should fix these issues before merging

AlexRamRam:
The term _dry_run_ is prevalent throughout the EVM runner modules as variables and as parts of interfaces. Can you clarify what the intent is - ie. are these to facilitate testing or is there a deeper functional purpose? In any case, I would benefit from more clarity on this design choice. Thanks.

src/3pc/agent/runners/evm/http_server.cpp
AlexRamRam: Done. See [b51f20c](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/b51f20c41f1e869bbc4397334d6e84999dc4232e)

src/3pc/agent/runners/evm/http_server.cpp
maurermi: Fixed. All below return error to user comments addressed as well in [b68e944](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/b68e94424a49e49fdce9a46a25af5357aa757714)

src/3pc/agent/runners/evm/impl.cpp
maurermi: Addressed in [3a8aef9](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/3a8aef924e0ecbb786523e193a67b7bc61f50fe1)

src/3pc/agent/runners/evm/impl.cpp
maurermi: Addressed in [3a8aef9](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/3a8aef924e0ecbb786523e193a67b7bc61f50fe1)

src/3pc/agent/runners/evm/impl.cpp
maurermi: Addressed in [3a8aef9](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/3a8aef924e0ecbb786523e193a67b7bc61f50fe1)

maurermi:
@HalosGhost feels like this is safe to close now with parsec being merged AFAIK

PR #188, opened 2022-09-28
Fix casting errors in client-cli

### LINKED ISSUE #187: Casting errors in client-cli
### Affected Branch

trunk

### Basic Diagnostics

### Description

The executable compiled from `client-cli.cpp` is used to interact with the transaction processor.  In the "Launch System" section in README.md, it's called to mint new coins, print the balance of a wallet, make a new wallet, and send coins between wallets.  For example, to mint new coins, the command is:
```terminal
# ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat mint 10 5
```

When calling `client-cli` with the commands "mint", "send", or "fan", the code may exhibit a casting error.  For example, if the mint command is accidentally called with a negative number, such as in
```terminal
# ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat mint -1 5
```
the code would cast -1 to 18446744073709551615 and would create 18446744073709551615 new utxos without warning.  If the mint command is invoked with -18446744073709551615 outputs, it actually only makes one:
```terminal
root@102611d59e8f:/opt/tx-processor# ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat mint -18446744073709551615 5
[2022-09-28 05:18:56.541] [WARN ] Existing wallet file not found
[2022-09-28 05:18:56.541] [WARN ] Existing client file not found
34162c6120b9ddb3d1dd6f69b4898ba2af4e4e6868e3b099d39316c133ab54ae
root@102611d59e8f:/opt/tx-processor# ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat info
Balance: $0.05, UTXOs: 1, pending TXs: 0
```

This is caused by the use of `std::stoull` and `std::stoul`, which are used to convert strings to unsigned integers.  For example, here's the function `mint_command`: 
```C++
auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
    -> bool {
    static constexpr auto min_mint_arg_count = 7;
    if(args.size() < min_mint_arg_count) {
        std::cerr << "Mint requires args <n outputs> <output value>"
                  << std::endl;
        return false;
    }

    const auto n_outputs = std::stoull(args[5]);
    const auto output_val = std::stoul(args[6]);
```

### Code of Conduct

mszulcz-mitre:
client-cli can produce casting errors if the numbers given as command-line arguments are negative. With this commit, the code now checks if the arguments are negative, and if so, prints an error message and exits.

Signed-off-by: Michael L. Szulczewski <mszulczewski@mitre.org>

### DIFF
--- a/src/uhs/client/client-cli.cpp
+++ b/src/uhs/client/client-cli.cpp
@@ -28,8 +28,24 @@ auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
         return false;
     }
 
-    const auto n_outputs = std::stoull(args[5]);
-    const auto output_val = std::stoul(args[6]);
+    static constexpr auto n_outputs_arg_idx = 5;
+    const auto& n_outputs_arg = args[n_outputs_arg_idx];
+    auto idx_first_non_whitespace = n_outputs_arg.find_first_not_of(' ');
+    if(n_outputs_arg[idx_first_non_whitespace] == '-') {
+        std::cerr << "The requsted number of UTXOs cannot be negative."
+                  << std::endl;
+        return false;
+    }
+    const auto n_outputs = std::stoull(n_outputs_arg);
+
+    static constexpr auto output_val_arg_idx = 6;
+    const auto& output_val_arg = args[output_val_arg_idx];
+    idx_first_non_whitespace = output_val_arg.find_first_not_of(' ');
+    if(output_val_arg[idx_first_non_whitespace] == '-') {
+        std::cerr << "The UTXO values cannot be negative." << std::endl;
+        return false;
+    }
+    const auto output_val = std::stoul(output_val_arg);
 
     const auto mint_tx
         = client.mint(n_outputs, static_cast<uint32_t>(output_val));
@@ -103,11 +119,20 @@ auto send_command(cbdc::client& client, const std::vector<std::string>& args)
         return false;
     }
 
-    const auto value = std::stoul(args[5]);
+    static constexpr auto value_arg_idx = 5;
+    const auto& value_arg = args[value_arg_idx];
+    const auto idx_first_non_whitespace = value_arg.find_first_not_of(' ');
+    if(value_arg[idx_first_non_whitespace] == '-') {
+        std::cerr << "The amount of currency to send cannot be negative."
+                  << std::endl;
+        return false;
+    }
+    const auto value = std::stoul(value_arg);
+
     static constexpr auto address_arg_idx = 6;
     auto pubkey = decode_address(args[address_arg_idx]);
     if(!pubkey.has_value()) {
-        std::cout << "Could not decode address" << std::endl;
+        std::cout << "Could not decode address." << std::endl;
         return false;
     }
 
@@ -130,8 +155,23 @@ auto fan_command(cbdc::client& client, const std::vector<std::string>& args)
         return false;
     }
 
-    const auto value = std::stoul(args[6]);
-    const auto count = std::stoul(args[5]);
+    static constexpr auto value_arg_idx = 6;
+    const auto& value_arg = args[value_arg_idx];
+    auto idx_first_non_whitespace = value_arg.find_first_not_of(' ');
+    if(value_arg[idx_first_non_whitespace] == '-') {
+        std::cerr << "The value cannot be negative." << std::endl;
+        return false;
+    }
+    const auto value = std::stoul(value_arg);
+
+    static constexpr auto count_arg_idx = 5;
+    const auto& count_arg = args[count_arg_idx];
+    idx_first_non_whitespace = count_arg.find_first_not_of(' ');
+    if(count_arg[idx_first_non_whitespace] == '-') {
+        std::cerr << "The count cannot be negative." << std::endl;
+        return false;
+    }
+    const auto count = std::stoul(count_arg);
 
     static constexpr auto address_arg_idx = 7;
     auto pubkey = decode_address(args[address_arg_idx]);


src/uhs/client/client-cli.cpp
```diff
@@ -28,8 +28,25 @@ auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
         return false;
     }
 
-    const auto n_outputs = std::stoull(args[5]);
-    const auto output_val = std::stoul(args[6]);
+    const auto n_outputs_arg_idx = 5;
```
metalicjames: Use `constexpr` for constants

src/uhs/client/client-cli.cpp
```diff
@@ -28,8 +28,25 @@ auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
...
+    const std::string& n_outputs_arg = args[n_outputs_arg_idx];
+    std::wstring::size_type idx_first_non_whitespace
```
metalicjames: Use `auto` instead of explicitly specifying `size_type`

src/uhs/client/client-cli.cpp
```diff
@@ -28,8 +28,25 @@ auto mint_command(cbdc::client& client, const std::vector<std::string>& args)
         return false;
     }
 
-    const auto n_outputs = std::stoull(args[5]);
-    const auto output_val = std::stoul(args[6]);
+    const auto n_outputs_arg_idx = 5;
+    const std::string& n_outputs_arg = args[n_outputs_arg_idx];
```
metalicjames: Use `auto&`

metalicjames (changes requested):
This looks good, but you should use `auto` and `constexpr`

HalosGhost:
Just a food-for-thought question: would it make sense to try to parse the argument as a `signed long long` first instead of trying to do some manual parsing tricks? Doing so means the test is a little more straight-forward:

```
sll ← as_sll(args₅)
if sll is E2BIG:
    ull ← as_ull(args₅)
elif sll ≤ 0 or sll is Error(_):
    error out
else:
    ull ← cast<ull>(sll)
```

Perhaps that's not really any cleaner than manually finding the first non-space character and manually matching `-`. Don't take this as a suggested change, just something to consider.

metalicjames:
>  would it make sense to try to parse the argument as a signed long long first instead of trying to do some manual parsing tricks?

+1

I agree that's a better approach

mszulcz-mitre:
@HalosGhost Interesting suggestion!  What does the first check in your pseudo-code do?
```terminal
sll ← as_sll(args₅)
if sll is E2BIG:
    ull ← as_ull(args₅)
```
I'm used to seeing E2BIG as an error code that means "Argument list too long" ([C++ ref](https://en.cppreference.com/w/cpp/error/errno_macros), [GNU ref](https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html)).

I'm happy to do the test either way, but the advantages of first converting to a `signed long long` aren't clear to me.  It seems like the main difference is whether you check if a signed int is less than zero or if a string starts with a minus character.

HalosGhost:
> What does the first check in your pseudo-code do?

Sorry for the confusion, the actual error-code I should have referenced was `ERANGE` (that the conversion failed because it fell outside the representable range of `signed long long`). And, in reality, we probably need to use [`std::strtoll()`](https://en.cppreference.com/w/cpp/string/byte/strtol) because exceptions are disabled (so a failed parse with `std::sto*()` will effectively `std::abort()`).

The primary benefit of this would essentially be that it offloads the actual parsing (e.g., skipping whitespace) to the standard library.

Again, I'm not actually sure this is dramatically cleaner…

mszulcz-mitre:
> Sorry for the confusion, the actual error-code I should have referenced was ERANGE

Great--thanks for clarifying.

> And, in reality, we probably need to use [std::strtoll()](https://en.cppreference.com/w/cpp/string/byte/strtol) because exceptions are disabled (so a failed parse with std::sto*() will effectively std::abort()).

Is it a problem to terminate the program for out-of-range errors?  I guess this is what you're saying above, but just to be certain, this is what happens with the current use of `std::stoll` and `std::stol`:
```terminal
root@3fd9e63ade0c:/opt/tx-processor# ./build/src/uhs/client/client-cli 2pc-compose.cfg mempool0.dat wallet0.dat mint 1000000000000000000000000000000 5
terminate called after throwing an instance of 'std::out_of_range'
  what():  stoull
Aborted (core dumped)
```
Maybe the error message could be a little clearer, but it doesn't seem too bad to me.

HalosGhost:
In the general case, no it's not a problem to terminate on those cases. However, with the alternative I offered, it's possible to require parsing twice (when the argument fits into the range of `unsigned long long`, but is too large to fit into a `signed long long`). When that happens, instead of appropriately double-parsing, it'll just fail out because of the out-of-range exception from the first parse.

Maybe it doesn't actually matter (this is research-level code after all), and restricting the client to work within `[0, LLONG_MAX]` is fine (even though the range available to the underlying API is `[0, ULLONG_MAX]`). In that case, there's no need for the double-parse (or error-handling) at all, and the pseudo-code becomes:

```
sll ← as_sll(args₅)
if sll ≤ 0:
    error out

ull ← cast<ull>(sll)
```

mszulcz-mitre:
Here's an implementation that I think does everything that @HalosGhost's pseudo-code does:
```C++
const auto* p = n_outputs_arg.data();
char* p_end;
const auto n_outputs_sll = std::strtoll(p, &p_end, 10);
if (n_outputs_sll < 0){
    std::cerr << "The requsted number of UTXOs cannot be negative."
              << std::endl;
    return false;
}
const auto n_outputs = std::stoull(n_outputs_arg);
```
I ended up modifying the pseudo-code because the first part would fail if the input argument was both out of range and negative:
```terminal
sll ← as_sll(args₅)
if sll is E2BIG:
    ull ← as_ull(args₅) \\ This will have a casting error if args₅ is negative.
```

For comparison, here's the version that checks for a leading '`-`'  in the argument string:
```C++
auto idx_first_non_whitespace
    = n_outputs_arg.find_first_not_of(' ');
if(n_outputs_arg[idx_first_non_whitespace] == '-') {
    std::cerr << "The requsted number of UTXOs cannot be negative."
              << std::endl;
    return false;
}
const auto n_outputs = std::stoull(n_outputs_arg);
```

Both methods look good to me.  However, I guess I favor the latter a little because it seems simpler and more readable.  I'll plan to push a cleaned version of it, but if someone feels strongly about the `strtoll` version, I'm happy to change it.

HalosGhost:
I've gone back and forth on this a couple of times, and I think the only reason I'm coming down in-favor of parsing rather than hand-matching is that there's another error case that's not handled in the code you have so far: asking for 0 UTXOs (or for UTXOs of 0-value). If we already have it as an integer, we can catch both error cases using `<=` instead of `<` (and changing the error appropriately). this is actually also why preferring to parse is, in-general, safer: it's more likely to catch malformed input earlier, and gives you structured data as output so you can refine more immediately.

PR #189, opened 2022-10-04
Performance fixes for sentinel attestations

wadagso-gertjaap:
This PR adds performance improvements to the sentinel attestations:

- Validation and attestation happen in parallel threads in the sentinel
- Attestation requests are prioritized by sentinels such that the end-to-end latency of transactions reduces under high(er) load: incoming transactions will wait until existing transactions pending attestation are completed.
- Validation of attestations happens in parallel in the coordinator
- Validation of attestations happens in parallel and outside of the state machine (before replication) in the shard.

### DIFF
--- a/src/uhs/twophase/coordinator/controller.cpp
+++ b/src/uhs/twophase/coordinator/controller.cpp
@@ -106,7 +106,21 @@ namespace cbdc::coordinator {
         // Initialize NuRaft with the state machine we just created. Register
         // our callback function to notify us when we become a leader or
         // follower.
-        return m_raft_serv->init(m_raft_params);
+        if(!m_raft_serv->init(m_raft_params)) {
+            return false;
+        }
+
+        auto n_threads = std::thread::hardware_concurrency();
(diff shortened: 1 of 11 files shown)


src/uhs/twophase/coordinator/controller.cpp
```diff
@@ -106,7 +106,21 @@ namespace cbdc::coordinator {
         // Initialize NuRaft with the state machine we just created. Register
         // our callback function to notify us when we become a leader or
         // follower.
-        return m_raft_serv->init(m_raft_params);
+        if(!m_raft_serv->init(m_raft_params)) {
+            return false;
+        }
+
+        auto n_threads = std::thread::hardware_concurrency();
```
anders94, line 113: Have you considered launching one less thread than `hardware_concurrency()` returns? We've found in benchmarking signature checking that using one less performs markedly better presumably because there is less swapping. Might be worth benchmarking within the context of this workload.

PR #190, opened 2022-10-11
Fix missing dependency

HalosGhost:
* Enabled cmake to check for and locate snappy
* Link snappy in everywhere it needs to be
* Ensure snappy is installed by ./scripts/configure.sh

Closes #171; closes #175

Signed-off-by: paparuch <paparuch@0day.im>
Signed-off-by: Sam Stuewe <stuewe@mit.edu>
Co-authored-by: Sam Stuewe <stuewe@mit.edu>

PR #192, opened 2022-10-26
Use explicit path to specify snappy library target

AlexRamRam:
By using the $SNAPPY_LIBRARY env var defined in the base CMakeLists.txt. Note that in macOS the brew snappy library installation path may not be in the LD_LIBRARY_PATH, hence this is required

Signed-off-by: Alexander Jung <104335080+AlexRamRam@users.noreply.github.com>

### DIFF
--- a/src/uhs/atomizer/archiver/CMakeLists.txt
+++ b/src/uhs/atomizer/archiver/CMakeLists.txt
@@ -16,4 +16,4 @@ target_link_libraries(archiverd archiver
                                 ${NURAFT_LIBRARY}
                                 secp256k1
                                 ${CMAKE_THREAD_LIBS_INIT}
-                                snappy)
+                                ${SNAPPY_LIBRARY})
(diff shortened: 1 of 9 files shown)


metalicjames (approved):
This is a necessary change after #190 to use the `SNAPPY_LIBRARY` variable set by CMake's `find_library`. 

However, I would prefer to remove the snappy dependency altogether as we don't use it (at least intentionally).

HalosGhost:
@metalicjames I've tested 2129fba46ad5b6ddca98a355de938423033dd1f2 in a clean environment, and everything went swimmingly. All else held equal, I'd prefer to have fewer dependencies than more. @AlexRamRam can you test James's commit above to confirm that it also solves a clean-build in macOS?

AlexRamRam:
> @metalicjames I've tested [2129fba](https://github.com/mit-dci/opencbdc-tx/commit/2129fba46ad5b6ddca98a355de938423033dd1f2) in a clean environment, and everything went swimmingly. All else held equal, I'd prefer to have fewer dependencies than more. @AlexRamRam can you test James's commit above to confirm that it also solves a clean-build in macOS?

Yes - commit https://github.com/mit-dci/opencbdc-tx/commit/2129fba46ad5b6ddca98a355de938423033dd1f2 builds fine in MacOS (tested after removing snappy & leveldb).

HalosGhost:
@AlexRamRam Thank you for testing. I'd actually like to close this then. @metalicjames would you like to open a PR with just 2129fba, and I can merge it ASAP?

PR #207, opened 2023-01-19
Added microbenchmarks using google test

maurermi:
This commit contains benchmarks for monitoring performance of basic features in this repository. This commit includes benchmarks for low_level primitives, as well as leveldb storage and std::unordered_set storage.

Signed-off-by: Michael Maurer <maurer.mi@northeastern.edu>

### DIFF
--- a/benchmarks/low_level.cpp
+++ b/benchmarks/low_level.cpp
@@ -1,0 +1,180 @@
+// Copyright (c) 2021 MIT Digital Currency Initiative,
+//                    Federal Reserve Bank of Boston
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// Note: Contains call to BENCHMARK_MAIN
+
+#include "uhs/transaction/messages.hpp"
+#include "uhs/transaction/transaction.hpp"
+#include "uhs/transaction/validation.hpp"
+#include "uhs/transaction/wallet.hpp"
+#include "util/serialization/istream_serializer.hpp"
+#include "util/serialization/ostream_serializer.hpp"
+
+#include <benchmark/benchmark.h>
+#include <filesystem>
+#include <gtest/gtest.h>
+
+class low_level : public ::benchmark::Fixture {
+  protected:
+    void SetUp(const ::benchmark::State&) override {
+        auto mint_tx1 = wallet1.mint_new_coins(100, 2);
+        wallet1.confirm_transaction(mint_tx1);
+
+        m_of.open(m_BENCHMARK_File);
+
+        counter = 1;
+    }
+
+    void TearDown(const ::benchmark::State&) override {
+        m_if.close();
+        m_of.close();
+        std::filesystem::remove_all(m_BENCHMARK_File);
+    };
+
+    cbdc::transaction::wallet wallet1;
+    cbdc::transaction::wallet wallet2;
+
+    cbdc::transaction::full_tx m_valid_tx{};
+    cbdc::transaction::full_tx m_valid_tx_multi_inp{};
+
+    // for serialization
+    std::ifstream m_if;
+    std::ofstream m_of;
+
+    cbdc::istream_serializer m_is{m_if};
+    cbdc::ostream_serializer m_os{m_of};
+
+    static constexpr auto m_BENCHMARK_File = "serial_BENCHMARK_File.dat";
+
+    // used to ensure that transaction validation timing
+    // uses unique N values for N->1 transactions
+    int counter = 1;
+};
+
+// serialize full tx
+BENCHMARK_F(low_level, serialize_tx)(benchmark::State& state) {
+    m_valid_tx = wallet1.send_to(2, wallet2.generate_key(), true).value();
+    for(auto _ : state) {
+        m_os << m_valid_tx;
+    }
+    m_of.close();
+}
+
+// serialize compact tx
+BENCHMARK_F(low_level, serialize_compact_tx)(benchmark::State& state) {
+    // serialize 1-1 tx
+    m_valid_tx = wallet1.send_to(2, wallet2.generate_key(), true).value();
+    auto cp_tx = cbdc::transaction::compact_tx(m_valid_tx);
+    for(auto _ : state) {
+        m_os << cp_tx;
+    }
+    m_of.close();
+}
+
+// deserialize full tx
+BENCHMARK_F(low_level, deserialize_tx)(benchmark::State& state) {
+    // serialize 1-1 tx
+    auto read_tx = cbdc::transaction::full_tx();
+    m_valid_tx = wallet1.send_to(2, wallet2.generate_key(), true).value();
+    for(auto _ : state) {
+        state.PauseTiming();
+        m_os.reset();
+        m_os << m_valid_tx;
+        m_of.close();
+        m_if.open(m_BENCHMARK_File);
+
+        read_tx = cbdc::transaction::full_tx();
+        m_is.reset();
+        state.ResumeTiming();
+        m_is >> read_tx;
+        state.PauseTiming();
+        ASSERT_EQ(read_tx, m_valid_tx);
+    }
+
+    m_if.close();
+}
+
+// deserialize compact tx
+BENCHMARK_F(low_level, deserialize_compact_tx)(benchmark::State& state) {
+    // serialize 1-1 tx
+    auto read_tx = cbdc::transaction::full_tx();
+    auto read_cp = cbdc::transaction::compact_tx(read_tx);
+
+    m_valid_tx = wallet1.send_to(2, wallet2.generate_key(), true).value();
+    auto cp_tx = cbdc::transaction::compact_tx(m_valid_tx);
+    for(auto _ : state) {
+        state.PauseTiming();
+        m_os.reset();
+        m_os << cp_tx;
+        m_of.close();
+        m_if.open(m_BENCHMARK_File);
+
+        read_cp = cbdc::transaction::compact_tx(read_tx);
+        m_is.reset();
+        state.ResumeTiming();
+        m_is >> read_cp;
+        state.PauseTiming();
+        ASSERT_EQ(cp_tx, read_cp);
+    }
+
+    m_if.close();
+}
+
+// wallet sign tx
+BENCHMARK_F(low_level, sign_tx)(benchmark::State& state) {
+    // sign 1-1 tx
+    for(auto _ : state) {
+        state.PauseTiming();
+        m_valid_tx = wallet1.send_to(2, wallet1.generate_key(), true).value();
+
+        state.ResumeTiming();
+        wallet1.sign(m_valid_tx);
+        state.PauseTiming();
+
+        wallet1.confirm_transaction(m_valid_tx);
+    }
+}
+
+// tx validiation
+BENCHMARK_F(low_level, valid_tx)(benchmark::State& state) {
+    // validate n-1 tx
+    auto err = cbdc::transaction::validation::check_tx(m_valid_tx);
+    for(auto _ : state) {
+        m_valid_tx = wallet1.mint_new_coins(counter, 1);
+        wallet1.confirm_transaction(m_valid_tx);
+        m_valid_tx
+            = wallet1.send_to(counter, wallet2.generate_key(), true).value();
+        state.ResumeTiming();
+        err = cbdc::transaction::validation::check_tx(m_valid_tx);
+        state.PauseTiming();
+        counter++;
+    }
+}
+
+// test quick-failing validation
+BENCHMARK_F(low_level, no_inputs)(benchmark::State& state) {
+    m_valid_tx.m_inputs.clear();
+    auto err = cbdc::transaction::validation::check_tx(m_valid_tx);
+    for(auto _ : state) {
+        err = cbdc::transaction::validation::check_tx(m_valid_tx);
+    }
+}
+
+// calculate uhs id
+BENCHMARK_F(low_level, calculate_uhs_id)(benchmark::State& state) {
+    m_valid_tx = wallet1.send_to(2, wallet2.generate_key(), true).value();
+    auto cp_tx = cbdc::transaction::compact_tx(m_valid_tx);
+    auto engine = std::default_random_engine();
+    for(auto _ : state) {
+        state.PauseTiming();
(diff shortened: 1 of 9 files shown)


HalosGhost:
Strong concept-ACK; being able to use something like this to build an intuition for how the system performs will be excellent.

~Note: the CI docker `pull_request / docker-build` workflow will not pass because this PR modifies the build.~ I'll be testing locally to make sure the image builds function as-expected before merge.

HalosGhost:
Sorry for the delay; local testing is running into a couple issues which I hope to have resolved shortly.

HalosGhost (changes requested):
Okay, with the help of some folks in `#docker`, I finally figured out what's going wrong; and, in retrospect, it's obvious. `benchmarks` is listed in `.dockerignore`. This PR needs to be modified to remove that line.

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
...
+        std::filesystem::remove_all(m_BENCHMARK_File);
+    }
+
+    cbdc::transaction::wallet wallet1;
+    cbdc::transaction::wallet wallet2;
+
+    cbdc::transaction::full_tx m_valid_tx{};
+    cbdc::transaction::full_tx m_valid_tx_multi_inp{};
+
+    // std::unique_ptr<secp256k1_context, decltype(&secp256k1_context_destroy)>
```
HalosGhost: Here, and elsewhere, let's avoid leaving dangling comments.

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
+// Copyright (c) 2021 MIT Digital Currency Initiative,
+//                    Federal Reserve Bank of Boston
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+/*
+    File contains benchmarks for
+    -   transactions size [1, 4] -> [1, 2]
```
HalosGhost: This list is pretty likely to fall out-of-date, and the resulting benchmarks executable is able to list all benchmarks grouped by their fixture. I'd suggest removing it.

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
...
+};
+
+// create_tx 1-1
+BENCHMARK_F(low_level, create_1to1_tx)(benchmark::State& state) {
+    // if(state.thread_index() == 0) {
+    //     GTEST_LOG_(INFO) << "Starting transaction benchmarks ... ";
+    // }
+    for(auto _ : state) {
+        m_valid_tx
+            = wallet1.send_to(100, wallet2.generate_key(), true).value();
```
HalosGhost: `wallet::send_to()` just generates a transaction (might be a good candidate for a better name in the future), the result needs to be sent somewhere for it to matter. As a result, there's no need for the confirm-and-send-back pattern.

The one thing we would want to double-check is to ensure that the tx-creation/assignment isn't elided due to it not being used. But, that could just be an assert checking that the transaction is structured as you expect (1:1).

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+// create_tx 1-2
+BENCHMARK_F(low_level, create_1to2_tx)(benchmark::State& state) {
+    for(auto _ : state) {
+        state.PauseTiming();
+        cbdc::transaction::wallet wallet_a;
+        cbdc::transaction::wallet wallet_b;
+        auto mint_tx_a = wallet_a.mint_new_coins(1, 50);
```
HalosGhost: The `SetUp` already mints coins for the wallets at the beginning of each benchmark; are more needed?

What we're actually timing here will be clearer if each individual benchmark can minimize any other ceremony (there will probably always be some small amount, but I think less here is possible).

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
...
+        m_is >> read_cp;
+        state.PauseTiming();
+        ASSERT_EQ(cp_tx, read_cp);
+    }
+
+    m_if.close();
+}
+
+// roundtrip benchmarking full tx
+BENCHMARK_F(low_level, roundtrip_tx)(benchmark::State& state) {
```
HalosGhost: I love roundtrip tests, but since we already benchmark serialize and deserialize (because the individual performance of those is actually more helpful for us in most cases), I don't know that round-tripping here offers us much help beyond the two more granular cases.

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,455 @@
...
+            = wallet2.send_to(100, wallet1.generate_key(), true).value();
+        wallet1.confirm_transaction(m_valid_tx);
+        wallet2.confirm_transaction(m_valid_tx);
+    }
+}
+
+// tx validiation
+BENCHMARK_F(low_level, valid_tx)(benchmark::State& state) {
+    // validate 1-1 tx
+    auto err = cbdc::transaction::validation::check_tx(m_valid_tx);
```
HalosGhost, line 143: So, this is validating (and below, failing validation for) the same tx over and over again. Is there a significant difference in-numbers if we validate a different transaction every time? If so, would the varied benchmark be more fair?

scripts/benchmarks.sh
```diff
@@ -0,0 +1,91 @@
...
+# If the build folder is a relative path, convert it to an absolute path
+# to avoid potential relative path errors and to improve readability
+# if the path is written to stdout.
+export BUILD_DIR=$(cd "$BUILD_DIR"; pwd)
+echo "Build folder: '${BUILD_DIR}'"
+echo
+
+run_test_suite () {
+    cd "$BUILD_DIR"
+    #find . -name '*.gcda' -exec rm {} \;
```
HalosGhost: Another dangling comment.

scripts/test.sh
```diff
@@ -65,6 +65,7 @@ EXAMPLES:
 # Parse command-line arguments.
 RUN_UNIT_TESTS="true"
 RUN_INTEGRATION_TESTS="true"
+RUN_BENCHMARKS="true"
```
HalosGhost: It looks like this is left-over from an older version of this PR where the benchmarks were run as a part of the test suite. I think it probably makes sense to remove all changes to `./scripts/test.sh`.

benchmarks/uhs_leveldb.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include "util/serialization/istream_serializer.hpp"
+#include "util/serialization/ostream_serializer.hpp"
+
+#include <benchmark/benchmark.h>
+#include <filesystem>
+#include <gtest/gtest.h>
+#include <leveldb/db.h>
+#include <leveldb/write_batch.h>
+
+#include <vector>
```
HalosGhost, line 15: I know you ran into some issues with the Fixture and threading for levelDB; is there another way of grouping benchmarks in leveldb? It would be nice if these could be grouped similarly to how the Fixture groups them for low-level.

HalosGhost (changes requested):
There are a few things I think we should probably clean-up before merge (comments made inline), but I think it's not much work before this is in a great position.

I do think we should take some time to simplify the low-level benchmarks; mint coins in the `SetUp` with enough versatility that you can use them for each benchmark instead of having to set them up additional times. I'd also suggest a helper function here; if you have $N$ coins available to spend, you can always generate an $N-in/1-out$ transaction (and extending to $N-in/2-out$ is usually trivial). You can make this even easier to reason about by minting `2`-valued coins. $N-in/1-out$ is accomplished by sending $2×N$ in-value (e.g., $4-in/1-out$ is sending `8`), $N-in/2-out$ by sending $2×N-1$ (e.g., $4-in/2-out$ by sending `7`). Let me know if that's not clear!

benchmarks/low_level.cpp
maurermi: Good note -- mistake on my part, I thought I had removed these interleaved comments before sending this commit

benchmarks/low_level.cpp
maurermi: To exclusively rely on the `SetUp` method would require the wallets to be initialized to hold a huge amount of value. This is because since the number of times that the benchmark is run is dynamically determined (that is, the number of times the code within the `for(auto _ : state)` block). `SetUp` does not get called before each loop, so the possible workarounds I've identified are to instantiate new wallets each loop and discard the setup time, mint new coins each loop (not sure if I should reset the wallets in this scenario), or force a maximum number of loops. 

Not sure which method would be most beneficial here, happy to hear any thoughts.

benchmarks/low_level.cpp
maurermi: This is a good question, and something I was wondering myself -- I'd agree that it's likely better to generate new transactions rather than reusing the same one, will make that change

benchmarks/uhs_leveldb.cpp
maurermi: Good question, I will need to look into this more, but one idea I'll start with is uniting them with a class -- in the output I would assume they will be listed in class/subclass format just as the Fixtures do

benchmarks/low_level.cpp
HalosGhost: Mmm, that's interesting, and kind of an odd limitation. In that case, I'd suggest making an additional helper function that you call on each iteration (outside timing) to setup the wallets the same way each time. That will still reduce the noise in each benchmark definition.

benchmarks/low_level.cpp
maurermi: Quick note -- for the failed validation, I would say that the usage of the same transaction in this particular case is sensible, since it will continuously fail on an early step (before much processing occurs). However, this would suggest that different types of verification failures should be benchmarked as well.

benchmarks/low_level.cpp
maurermi: Resolving because of modifications made

benchmarks/low_level.cpp
HalosGhost: That's a reasonable point. I'd suggest marking it as an early-failout then, and maybe adding a todo (or we can open an issue after this is merged) to expand the failing benchmarks in the future.

benchmarks/uhs_leveldb.cpp
```diff
@@ -0,0 +1,203 @@
...
+        std::memcpy(out_arr.data(),
+                    db.m_cp_tx.m_uhs_outputs.data(),
+                    db.m_cp_tx.m_uhs_outputs.size());
+        leveldb::Slice OutPointKey(out_arr.data(),
+                                   db.m_cp_tx.m_uhs_outputs.size());
+
+        // actual storage
+        state.ResumeTiming();
+        db.res = db.m_db->Put(db.write_opt, OutPointKey, leveldb::Slice());
+        state.PauseTiming();
```
HalosGhost, line 113: Here, and below: Are these end-of-loop `::PauseTiming()` calls necessary? You have one at the beginning of each loop as well. Given that the call comes off `state` which is what's being looped over (and is thus persisting across iterations), I'd imagine this (at most) might skip timing the loop iteration mechanics. If so, should we do that everywhere?

benchmarks/low_level.cpp
```diff
@@ -0,0 +1,180 @@
...
+        m_valid_tx
+            = wallet1.send_to(counter, wallet2.generate_key(), true).value();
+        state.ResumeTiming();
+        err = cbdc::transaction::validation::check_tx(m_valid_tx);
+        state.PauseTiming();
+        counter++;
+    }
+}
+
+// test a failed validation
```
HalosGhost:
```suggestion
// test quick-failing validation
```

scripts/test.sh
```diff
@@ -181,3 +181,5 @@ then
 else
     echo "Skipping integration tests."
 fi
+
+echo
```
HalosGhost: I don't think this is necessary?

benchmarks/transactions.cpp
```diff
@@ -0,0 +1,138 @@
...
+    cbdc::transaction::wallet wallet_b;
+    bool valid;
+    for(auto _ : state) {
+        reset_wallets(wallet_a, wallet_b, SWEEP_MAX);
+        valid = generate_Nto1_tx(wallet_a, wallet_b, state.range(0), state);
+        if(!valid) {
+            GTEST_LOG_(ERROR) << state.range(0) << "-2 transaction invalid";
+            return;
+        }
+        ASSERT_EQ(wallet_a.balance(), SWEEP_MAX * 2 - 2 * state.range(0));
```
HalosGhost, line 100: It looks like we're timing everything in these few functions, including the `ASSERT`s; is that something the `BENCHMARK()` pattern is smart enough to ignore?

HalosGhost (changes requested):
I think there are just a few small changes left and this will be ready for merge.

benchmarks/uhs_leveldb.cpp
maurermi: Ah this is an oversight, I intended to remove the `::PauseTiming() `calls at the start of the loop

benchmarks/transactions.cpp
maurermi: In these functions, starting/stopping the timer is handled by `generate_Nto1_tx()` and `generate_Nto2_tx()`, so the `ASSERT`s are ignored (all timing occurs within these functions, and does not occur within `Nto1_tx(benchmark::State&)` or `Nto2_tx(benchmark::State&)`.

scripts/test.sh
maurermi: Agreed

HalosGhost:
tACK, this looks good to me, and runs solidly on my local. I have one last question: given that we have a benchmark for deletion from leveldb, should we also have a benchmark for deletion from the set?

maurermi:
@HalosGhost good suggestion, I have added unordered_set removal in the most recent commit

HalosGhost:
@maurermi this looks solid. Squash-when-ready, I'll do a final double-check and we'll get this merged!

HalosGhost (approved):
This PR adds a minimal basis for microbenchmarks to help anyone working with a code develop an intuition for how the system behaves on a variety of tasks. This is both a huge tool to help a general intuition for the performance of the system, and also a very powerful tool for regression testing (at some point, it might make sense to add a CI step that runs this script and compares it to the performance of the last commit to help establish a performance floor, but that's a ways off).

Looks good to me! Merging.

PR #209, opened 2023-02-15
Updates to Config file Auto-generation (includes #138)

maurermi:
This PR looks to help complete #138 by making the changes requested in the most recent review. Namely, this change redirects the config file output to `stdout`, and confines all other other messages to `stderr`. In addition, adds changes which reflect current naming conventions and design choices.


tools/config_generator/config_generator.cpp
```diff
@@ -0,0 +1,860 @@
...
+                                           false);
+            return_msg += create_component(watchtower_count_key,
+                                           std::get<size_t>(watchtower_count),
+                                           false);
+        }
+        write_generated_config_to_file();
+        return_msg += "Completed...";
+        return return_msg;
+    }
+}
```
maurermi, line 806: This should be updated to provide support for loadgens, and seeding to appropriately reflect current config file requirements for local testing

HalosGhost:
@maurermi note the merge conflict in `CMakeLists.txt`

maurermi:
Merge resolved, updates required to unit tests to reflect new functionality as well (coming soon)

tools/config_generator/config_generator.cpp
```diff
@@ -0,0 +1,811 @@
...
+        } else {
+            const auto atomizer_count
+                = get_param_from_template_file(atomizer_count_key, config_map);
+            const auto archiver_count
+                = get_param_from_template_file(archiver_count_key, config_map);
+            const auto watchtower_count
+                = get_param_from_template_file(watchtower_count_key,
+                                               config_map);
+            return_msg += create_component(shard_count_key,
+                                           std::get<size_t>(shard_count),
```
maurermi: If `shard_count` is not specified in the template file, or is initialized to anything besides a legal `size_t` value, `std::get<size_t>(shard_count)` will fail and cause an abort. This also applies to all below calls of `create_component` with their corresponding component names.

Since this is a tool external to the system, this is probably not a big deal, but I am putting this here for consideration.

maurermi:
@HalosGhost this is ready for a review -- in particular it would be good to know whether (and if so, what) additional functionality would be helpful to be included.

tools/config_generator/config_generator.cpp
HalosGhost: Yeah, I'm also not too concerned about this. Though, anything that's a bit error-prone like this underscores that everything here would be less error-prone if we had better error-handling for config file parsing.

tools/config_generator/config_generator.cpp
```diff
@@ -0,0 +1,811 @@
...
+        const auto seed_begin
+            = get_param_from_template_file(seed_from, config_map);
+        const auto seed_end
+            = get_param_from_template_file(seed_to, config_map);
+        if(!std::holds_alternative<std::string>(seed_begin)
+           && !std::holds_alternative<std::string>(seed_end)
+           && std::get<size_t>(seed_begin) != std::get<size_t>(seed_end)) {
+            std::pair<std::string, std::string> key_pair = create_key_pair();
+            set_param_to_config_file(seed_privkey, key_pair.first);
+            write_generated_config_to_file();
```
HalosGhost: Looks like this exhibits undesirable behavior; namely, it still writes the config to-file, and it seems to be triggered if `seed_{begin,end}` aren't specified in the template. Here's my local test session reproducing this:

```console
$ ./scripts/build.sh
<snip>
$ cd ./build/tools/config_generator
$ ./generate_config ../../../config/unit/2pc_template_unit_test.tmpl 5000
Warning: Could not find param, seed_from.
Warning: Could not find param, seed_to.
SUCCESS

```

Since the aim is for the generator to only produce configs on `stdout`, maybe we should completely delete `write_generated_config_to_file()`?

2pc-compose.cfg
HalosGhost, line 1: So, the `README` has been updated to instruct people to use the `config/general` versions of this file (and `atomizer-compose.cfg`). Is there any reason to leave these around?

HalosGhost (changes requested):
Looks like there are a couple of things to address (a few of them minor clean-ups or considerations, but one of them is definitely more pressing).

2pc-compose.cfg
maurermi: Realistically, no -- however I am curious whether you think it's preferable to contain config files in the `config/general` folder or whether we should keep them in the top level folder. I have no real preference

HalosGhost: Personally, I kind of love the idea of cleaning up the root of the repo (if anyone can figure out how to get docker-compose to be happy about having its configuration files siloed off instead of polluting the root directory, I would definitely accept that PR).

However, my only strong preference is consistency. ☺

tools/config_generator/config_generator.cpp
maurermi: This is a good note -- I neglected that this tries to print to stdout whereas the config file is printed to stderr when other paths are taken

2pc-compose.cfg
maurermi: Sounds good to me! I think that I have fixed the docker issue as well in the most recent commit by modifying the Dockerfile

tools/config_generator/config_generator.cpp
HalosGhost: Sorry, just to clarify: the config should be written to `stdout`, the rest of the messages (like, `SUCCESS`) should be written to `stderr`; but the bigger thing is that in the path above, the config actually was never written to any visible stream.

maurermi: Got it, I will make that adjustment. It looked to me like previously `write_config_to_file()` was trying to write the value of `m_new_config` to stdout (possibly was uninitialized which is why nothing was printed, or that print was trapped).

PR #210, opened 2023-02-16
Updates to Moved decode_address from client-cli to client (includes #176)

maurermi:
Fixes linting error from PR #176 , should enable merging

### DIFF
--- a/src/uhs/client/client.cpp
+++ b/src/uhs/client/client.cpp
@@ -5,6 +5,8 @@
 
 #include "client.hpp"
 
+#include "bech32/bech32.h"
+#include "bech32/util/strencodings.h"
 #include "crypto/sha256.h"
 #include "uhs/sentinel/format.hpp"
 #include "uhs/transaction/messages.hpp"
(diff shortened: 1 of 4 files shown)


src/uhs/client/client.cpp
```diff
@@ -20,6 +22,43 @@
 #include <utility>
 
 namespace cbdc {
+
+    namespace address {
+        auto decode(const std::string& addr_str)
+            -> std::optional<cbdc::hash_t> {
+            (void)addr_str;
```
HalosGhost: This snuck past me in #176, but this `void`-cast is non-sensical. The primary reason you'd ever `void`-cast something is to avoid a compiler-complaint for a variable not being used. In C++, it's far preferred to use ``[[maybe_unused]]``, but in this case, that's not even relevant (this variable *is* used just a few lines beneath this cast). At least, this should be removed.

HalosGhost (changes requested):
There's literally a single-line change that I think we should make before merging. Otherwise, this looks good to me.

Will pull for local-testing once more before merge.

src/uhs/client/client.cpp
maurermi: Good catch, I'm not sure why that was included at all -- removed in latest push

HalosGhost (approved):
tACK. Exposes address-decoding in the relevant linkable library. Merging.

PR #211, opened 2023-02-17
Adds plot.py for local performance analysis

maurermi:
Graphs latency and throughput of a locally run system by interpreting tx_samples_*.txt


HalosGhost (approved):
tACK. This is a simple python script to analyze `tx_samples` files you've collected (either generated locally, or copied locally from a larger test).

Looks nice!

PR #212, opened 2023-03-20
Generate hashes with random keys to avoid collisions

maurermi:
Address TODO in src/util/common/hashmap.hpp which calls for randomly generated hash keys. Requires use of libsodium for crypto-safe random number generation.

NOTE: This causes certain tests to fail currently because of the added randomness. This will be fixed shortly.

### DIFF
--- a/src/util/common/hashmap.hpp
+++ b/src/util/common/hashmap.hpp
@@ -9,6 +9,7 @@
 #include "buffer.hpp"
 #include "crypto/siphash.h"
 #include "hash.hpp"
+#include "random_source.hpp"
 
 #include <array>
 #include <cstring>
@@ -16,20 +17,14 @@
 namespace cbdc::hashing {
     /// \brief SipHash function to generate STL data structure hash keys for
     ///        system IDs.
-    ///
-    /// [TODO: Uses randomly generated initialization keys to insure that
-    /// colliding values will differ from run to run.] For transaction data
-    /// that comes from users, it is important to key data in sets and hashmaps
-    /// non-deterministically. Otherwise, malicious actors might pre-calculate
-    /// colliding values and spam the system with them to cripple performance.
-    /// \tparam T hash-like type to hash.
     template<class T>
     struct const_sip_hash {
         static_assert(std::is_trivial_v<T>);
         auto operator()(T const& tx) const noexcept -> size_t {
             // TODO: pick the initialization key at random.
-            static constexpr std::array<uint64_t, 2> siphash_key{0x1337,
-                                                                 0x1337};
+            auto generator = random_source("/dev/urandom");
+            std::array<uint64_t, 2> siphash_key{generator(),
+                                                generator()};
             CSipHasher hasher(siphash_key[0], siphash_key[1]);
             std::array<unsigned char, sizeof(tx)> tx_arr{};
             std::memcpy(tx_arr.data(), &tx, sizeof(tx));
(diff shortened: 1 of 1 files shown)


maurermi:
@HalosGhost some tests currently fail with this update because of the added randomness, but wanted to push it here first to get initial thoughts. My main question is whether libsodium is the most appropriate choice here, so if you have thoughts, I'd appreciate hearing them.

src/util/common/hashmap.hpp
```diff
@@ -41,9 +33,7 @@ namespace cbdc::hashing {
     template<>
     struct const_sip_hash<buffer> {
         auto operator()(const buffer& buf) const noexcept -> size_t {
-            // TODO: pick the initialization key at random.
-            static constexpr std::array<uint64_t, 2> siphash_key{0x1337,
-                                                                 0x1337};
```
maurermi, line 46: Also curious whether this seems appropriate. Since the random values can't be determined at compile time, this update required making this value non-constant. Dynamic RNG seems like a good choice to me, but curious if static would be more appropriate here.

src/util/common/hashmap.hpp
```diff
@@ -41,9 +33,7 @@ namespace cbdc::hashing {
...
+            std::array<uint64_t, 2> siphash_key{randombytes_random(), randombytes_random()};
```
maurermi: Also curious whether this seems appropriate. Since the random values can't be determined at compile time, this update required making this value non-constant. Dynamic RNG seems like a good choice to me, but curious if static would be more appropriate here.

HalosGhost: > My main question is whether libsodium is the most appropriate choice here, so if you have thoughts, I'd appreciate hearing them.

When in-doubt, libsodium (or [monocypher](https://monocypher.org/), in my opinion) is a very good fallback.

However, we already have our own [`random_source`](https://github.com/mit-dci/opencbdc-tx/blob/trunk/src/util/common/random_source.hpp#L27) which we use throughout the codebase. Since the purpose of choosing this key at-random is to avoid collision (rather than specifically a given security property with specific requirements), it probably makes more sense to just leverage our utility directly.

> Also curious whether this seems appropriate. Since the random values can't be determined at compile time, this update required making this value non-constant. Dynamic RNG seems like a good choice to me, but curious if static would be more appropriate here.

I suppose my immediate wonder is if it would be possible to inject values into that process during the test harness. That way, we could keep our deterministic tests, but still leverage the randomness at runtime. However, we would have then created a divergence between what we're testing and what we're running—and it might be more prudent to modify the tests to handle the randomness (to remain inline).

Mull on that possibility for a little while, and I will attempt to take a look at the affected tests in reasonably short order to have a better handle on the lift.

maurermi: Ah I completely neglected `random_source` -- jumped the gun on that one. One potential solution would be in testing, to seed `random_source` from a fixed file, rather than /dev/urandom as random_source.hpp suggests. However, this does still create divergence. Another, likely more difficult strategy would be when a random value is generated in a test, save it (either in a temporary variable or in a file) for evaluating whether the operation was performed correctly after the fact. 

Also upon further inspection, some of the unit tests that are failing are not ones which I'd expect to fail on first glance, so I will need to delve into that more as well.

HalosGhost: Potential food-for-thought: `random_source` takes a file source as its seed. We could, for the test suite, choose a stable, unchanging file to seed from (the most obvious option would be `/dev/zero`). This then leverages all the same code paths as would be used everywhere already, but because the seed is known, would allow the randomness to be predictable throughout the test suite.

This is only a very minimal divergence (and, at least in part, what seeding is for). Keep it in mind as a possibility as you explore the options.

maurermi:
Pushed an update to use random_source() based on /dev/urandom, but this is still very much a work in progress. 

@HalosGhost here's some of my thoughts on what to do here -- it might be simplest to specify the generator file in a config file (not to add more dependence on config files), and we could create a predictable RNG for a more authentic testing environment -- then in an actual system outside of the tests we specify /dev/random as the generator

Issue #173, opened 2022-09-08
Devise a more flexible alternative to configuration files


HalosGhost:
Our current configuration file setup is optimized for ease of testing. This is not at all a problem, but it does occasionally have its drawbacks and it has major negatives for anyone wanting to run a copy of this system in a more realistic environment. Namely, it makes it impossible to increase the number of components in the system dynamically (because the configuration file change would need to be disseminated to all components).

What form this would take is not necessarily clear; perhaps the obvious incremental step is to make configurations specific to a given component, and while that increases flexibility some, it doesn't wholly solve the issue.

Related reading: #81

cc @kylecrawshaw and @metalicjames

Issue #81, opened 2022-03-25
Adding Feature Flags


HalosGhost:
# Background

Right now, we have two architectures (2PC and Atomizer), so when a new feature gets proposed, there is a high-probability that implementation will require touching two separate locations in the code.
As we explore more architectures, or solutions which involve changes to the two we have, this problem is likely going to be more pronounced.
It results in the following draw-backs:

* The possibility for architectures to get out-of-sync in feature-support.
    * Accepting this leads to an extra documentation burden (documenting the matrix of which architecture supports which features).
    * Avoiding this leads to an ever-increasing implementation burden (a new feature will need to intentionally interoperate will all previously-merged features).
* Testing with a new feature (say, B) but without an old one (say, A) is tedious.
    * Accepting this would require the following overhead: `checkout` a new branch, `reset` to a commit before A, `cherry-pick` the relevant commits for B, and fix conflicts.
    * Avoiding this leads to treating any sufficiently-fundamental change to an architecture being treated as a new architecture.
* The mental burden necessary to get involved grows dramatically with every new feature (as even knowing which architecture you want to use becomes much less intuitive).
    * Accepting this leads to fewer contributions.
    * (Even partially) avoiding this leads to dramatically increased documentation burden (trying to explain not just which architectures support which features, but *why* and which you might want to use for a given experiment).
* Supplemental tooling (e.g., docker compose configurations) is subject to more and more significant duplication/complexity/churn.
    * Accepting this may mean that we end up with a very large library of docker compose configurations (one for each of the now-many architectures), which also implies frequent additions of new tooling solutions for each new architecture/feature.
    * Avoiding this either results in a much more complex setup for the supporting tooling such that it can handle the complexity above, or simply not supporting some architectures with tooling as well as others (additional maintenance or documentation burden, or both)

In short, we know there are a lot more architectures and features worth exploring; to ensure project sustainability, we need to develop patterns and conventions for adding new features in such a way that we can minimize the potential combinatorial expansion shown above.

# A Solution

[Feature flags](https://en.wikipedia.org/wiki/Feature_toggle) offer a potential avenue to refactor the code such that new features are gated behind a flag to change the system's behavior at run-time.
This would allow, for example, many different implementations of a sentinel to coexist and for which implementation gets used to be selected at run-time (rather than build- or deploy-time).

[...]

PR #213, opened 2023-04-05
Programmability - prep for merge

AlexRamRam:


src/3pc/agent/runners/evm/http_server.cpp
HalosGhost, line 1: @AlexRamRam so, I understand that this patch removes the callback parameter from `exec_tx`, but I'm not clear on why. Is this because it was unused? It looks like an error-reporting path was taken through the callback and is no longer present. Is that readded/replaced in a different patch?

HalosGhost: @metalicjames happy to have your insight here as well!

AlexRamRam: Thanks @HalosGhost. Please see 
[In http_server::exec_tx() fix error propagation via json value object](https://github.com/mit-dci/opencbdc-tx/pull/213/commits/887ccbee0d7261ce9274b006e7f82f41de11fe8c)

src/3pc/agent/runners/evm/impl.cpp
```diff
@@ -155,10 +155,35 @@ namespace cbdc::threepc::agent::runner {
         }
         auto tn = to_uint64(maybe_tn.value());
         auto tn_key = m_host->ticket_number_key(tn);
+        auto cb
```
maurermi: Note: I am not fully sure if this is what was meant by the comment to "split the nested lambdas into separate functions" -- would be good to have a second opinion

AlexRamRam: Thanks @maurermi . May I suggest we replace the new `cb()` with a separate private member function as follows:

`evm_runner::lock_tx_receipt( const cbdc::buffer& valid_txid ) { ... }`

Let me now what you think.

maurermi: @AlexRamRam I think that is a fair change, would help with readability

src/3pc/agent/runners/evm/http_server.cpp
HalosGhost: This looks much better. Is there a specific motivation for `res_success_cb` to be passed as `const&`?

AlexRamRam: If I recall, the clang linter inspired this change.

src/3pc/agent/runners/evm/impl.cpp
maurermi: @AlexRamRam I realized that this may not be necessarily feasible because `cb` requires `tn` to be in its scope. That would indicate that if it were made a separate method, it would require tn to be passed as an argument or for `tn` to be stored at class or global scope. It cannot be passed as an argument because `m_try_lock_callback` expects a function that takes one argument of type `const broker::interface::try_lock_return_type&`. I think rather than storing `tn` in a broader scope, perhaps it's simpler to just rename `cb` to a more informative name, (ex. `handle_tx_lock_receipt_callback` ?)

README.md
```diff
@@ -18,8 +18,9 @@ The design decisions we made to achieve these goals will help inform policy make
 **NOTE:** In cases where there are significant changes to the repository that might need manual intervention down-stream (or other important updates), we will [make a NEWS post](NEWS.md).
 
 # Architecture
-
-We explored two system architectures for transaction settlement, both based on an [unspent transaction output (UTXO)](https://en.wikipedia.org/wiki/Unspent_transaction_output) data model and transaction format.
+There are two main architectures as follows:
+## UTXO & UHS-Based Transaction Processor
```
```suggestion
We have explored several architectures as follows:

## UHS-Based
```

README.md
```diff
@@ -36,8 +37,19 @@ Both architectures handle multiple geo-distributed datacenter outages with a [re
     - Maximum demonstrated throughput ~1.7M transactions per second.
     - Geo-replicated latency <1 second.
 
-Read the [architecture guide](docs/architecture.md) for a detailed description of the system components and implementation of each architecture.
+Read the [2PC & Atomizer architecture guide](docs/architecture.md) for a detailed description of the system components and implementation of each architecture.
```
```suggestion
Read the [UHS architecture guide](docs/architecture.md) for a detailed description of the system components and implementation of each architecture.
```

*Note*: this is probably a good time to move `docs/architecture.md` to something like `docs/uhs-architecture.md` to be less confusing.

README.md
```diff
@@ -36,8 +37,19 @@ Both architectures handle multiple geo-distributed datacenter outages with a [re
...
+
+## Generic Smart Contract Platform ("3PC")
```
HalosGhost: Here and, in all other docs, please keep to the same style used in the rest of the document (e.g., unify the numbering/bulleting style with the section above), and ensure there is only one logical sentence/thought-per-line. In particular, if there's a `.`, you should probably add a new-line after it.

README.md
```diff
@@ -117,16 +130,16 @@ To run the system with our pre-built image proceed to the [next section](#launch
         ```terminal
         $ docker compose --file docker-compose-2pc.yml up --build
         ```
-    1. Atomizer architecture:
+    2. Atomizer architecture:
```
HalosGhost: This is not necessary; markdown almost always automatically numbers lists; it's a *really* good idea to get in the habit of numbering all lists as `1.` so that you don't need to make changes like this. (This is what is done in lists above.)

README.md
```diff
@@ -73,8 +85,9 @@ Note that if you have not already installed the xcode cli tools you will need to
 ```terminal
 # xcode-select --install
 ```
-
-# Run the Code
+# Run the Code: Programmability architecture
```
HalosGhost: This is jarring. Either the basics of getting started with 3PC should belong in the same place as the UHS architectures here in the readme, or the UHS architecture getting-started routine should also be separated to mirror the programmability stuff.

(non-blocking)

benchmarks/transactions.cpp
```diff
@@ -97,10 +97,14 @@ static void Nto1_tx(benchmark::State& state) {
             GTEST_LOG_(ERROR) << state.range(0) << "-2 transaction invalid";
             return;
         }
-        ASSERT_EQ(wallet_a.balance(), SWEEP_MAX * 2 - 2 * state.range(0));
-        ASSERT_EQ(wallet_a.count(), SWEEP_MAX - state.range(0));
-        ASSERT_EQ(wallet_b.balance(), state.range(0) * 2);
-        ASSERT_EQ(wallet_b.count(), 1);
+        ASSERT_EQ(
+            wallet_a.balance(),
+            static_cast<unsigned long>(SWEEP_MAX * 2 - 2 * state.range(0)));
```
HalosGhost, line 102: Am I right in thinking that all these with an explicit cast were not fixable by adding the literal suffixes? If not, feel free to immediately mark this as resolved.

If, however, these can be silenced with the suffix addition, that's definitely a little less noisy.

docs/programmability_architecture.md
HalosGhost, line 1: Reformat. In particular, follow standard document structure (use progressive headings, rather than skipping ones), and keep to a single-line per-sentence.

docs/programmability_architecture_diagram.jpg
HalosGhost, line 1: Here, and below, do not check-in binaries, only source-documents. If source-documents are not available or are in proprietary formats, they need to be rebuilt. Git is not for binaries.

docs/programmability_architecture.md
```diff
@@ -0,0 +1,33 @@
...
+
+The higher level computation layer executes programs (i.e. smart contracts) and uses the lower-level key-value data store to maintain its state. The _Agent_ is the main interface for this layer and serves the role of a transaction coordinator. This computation layer contains implementations of the _Runner_ interface which is an abstraction for retrieving and executing context-specific bytecode. The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua)](https://www.lua.org/) and demonstrates the flexibility of the architecture to support different program execution environments.
```
HalosGhost:
```suggestion
The higher level computation layer executes programs (i.e. smart contracts) and uses the lower-level key-value data store to maintain its state. The _Agent_ is the main interface for this layer and serves the role of a transaction coordinator. This computation layer contains implementations of the _Runner_ interface which is an abstraction for retrieving and executing context-specific bytecode. The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua](https://www.lua.org/) and demonstrates the flexibility of the architecture to support different program execution environments.
```

Note the removal of the extra paren (making such diffs more clear is the specific reason why it is good practice to keep to a single sentence-per-line).

docs/programmability_architecture.md
```diff
@@ -0,0 +1,33 @@
...
+### Directory
+The _Directory_ identifies the target _Shard_ unit for each and every key-value data pair. This is a detail managed by the _Broker_, allowing the end-client (the _Agent_) to interact with the distributed _Shards_ as a single database unit. The _Directory_ is where the sharding logic is defined and its implementation can be designed to optimize database resources for particular use cases.
+### JSON-RPC Server
+The _JSON-RPC Server_ implements the [Ethereum JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) over HTTP. This is the external interface compatible for external Ethereum clients such as wallets (e.g. Metamask), and development tools (e.g. Hardat, Truffle, ethers.js, web3.js)
+### Runner
+The _Runner_ is an abstraction for retrieving and executing context-specific bytecode. The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua)](https://www.lua.org/) and demonstrates the flexibility of the architecture to support different program execution environments.
```
HalosGhost:
```suggestion
The _Runner_ is an abstraction for retrieving and executing context-specific bytecode. The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua](https://www.lua.org/) and demonstrates the flexibility of the architecture to support different program execution environments.
```

Another additional parenthesis. Maybe it's worth re-reading the docs for copy-editing?

docs/programmability_architecture.md
```diff
@@ -0,0 +1,33 @@
...
+### Broker
+The _Broker_ is the interface component for both the external _Shards_ and the _Ticket Machine_. The _Broker_'s primary role is to abstract the external set of distributed _Shards_ as a single logical database unit, providing functions such as _begin()_, _commit()_, _finish()_ and _rollback()_. The _Broker_ is also the interface for the _Ticket Machine_ for the _Agent_.
+### Directory
+The _Directory_ identifies the target _Shard_ unit for each and every key-value data pair. This is a detail managed by the _Broker_, allowing the end-client (the _Agent_) to interact with the distributed _Shards_ as a single database unit. The _Directory_ is where the sharding logic is defined and its implementation can be designed to optimize database resources for particular use cases.
+### JSON-RPC Server
+The _JSON-RPC Server_ implements the [Ethereum JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) over HTTP. This is the external interface compatible for external Ethereum clients such as wallets (e.g. Metamask), and development tools (e.g. Hardat, Truffle, ethers.js, web3.js)
```
HalosGhost: > Hardat

Hardhat typo?

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
+# Programmability User Guide
+
+This README runs you through setting up the 3PC system with the EVM runner, and then interacting with it using Hardhat and MetaMask.
+
+## Prerequisites
+
+Make sure you have [Docker installed](https://docs.docker.com/engine/install/ubuntu/).
+
+Make sure you have [NodeJS v14+ installed](https://github.com/nodesource/distributions/blob/master/README.md).
```
HalosGhost: I got a warning for using Node v19; it all seemed to work fine, but if we have specific version requirements, we should specify them as clearly as we can.

In the ideal case, we'd include a `.nvmrc` (cf. https://github.com/mit-dci/opencbdc-ui/commit/e9bbd8f0d12de42bd04c829bbf950949f07057c5).

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Clone the repository
+
+First, clone OpenCBDC locally, and pull the programmability branch:
```
HalosGhost: This is repeated documentation from the README, I think it makes more sense to have the README include cloning directions, and leave it out of this guide. Additionally, note that this is written with the understanding that the code sits at the `programmability` branch, which won't be the case when this lands. Please rephrase so that everything references the correct locations for where everything will live *post-merge*.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Usage pre-merge (current state)
+The Dockerfile uses a remote image as the base image (`ghcr.io/mit-dci/opencbdc-tx-base:latest`). This will be updated when this branch is merged into trunk.
+Until then, we use a workaround which is to rebuild the base image, and
+tag it with the name of the remote repository, as the remote base image does
+not have 3PC dependencies.
+
+```
+docker build --target base . -t ghcr.io/mit-dci/opencbdc-tx-base:latest
+```
```
HalosGhost:
````suggestion
```console
$ docker build --target base . -t ghcr.io/mit-dci/opencbdc-tx-base:latest
```
````

Where possible, use appropriate syntax highlighting, and make clear when commands are to be run at the terminal (by convention, this is done by showing the `$` as the prompt for non-root commands (some flavorful folks who prefer zsh will use `%` instead here), and `#` for root-required commands).

File: docs/programmability_user_guide.md [hunk 14]
```diff
@@ -0,0 +1,231 @@
...
+### Build 3PC containers
+
+```
+docker compose -f docker-compose-3pc.yml build
+```
+
+### Start it up
+
+```
+# will build containers if not already built
```
HalosGhost: If this builds the images if they don't already exist in all cases, then do we need the build step? If it doesn't build in all cases, should we have the comment?

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+Finish with initializing hardhat using:
+
+```
+npx hardhat
+```
+
+Select to create a javascript project.
+
+If you are using Docker, edit your hardhat.config.js to look like this:
```
HalosGhost: Can we ship these two config files and replace this documentation with a copy command? If not, why not? Is there more explanation needed here?

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+    opencbdc: {
+      url: `http://*YOUR SPECIFIED IP*:*YOUR SPECIFIED PORT*`,
+      accounts: ["32a49a8408806e7a2862bca482c7aabd27e846f673edc8fb0000000000000000"]
+    }
+  }
+};
+```
+* Note: the value in the `accounts` array is the private key of one of the hard-coded, pre-minted accounts in 3PC/EVM.
+
+You should now be able to use hardhat through for instance
```
HalosGhost: > for

?

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+```
+npx hardhat console
+```
+
+Look at more fun stuff with hardhat and compiling and deploying contracts [here](https://hardhat.org/hardhat-runner/docs/getting-started#quick-start)
```
HalosGhost: In general, avoid using `here` or similar things as link-text (terrible for accessibility), consider rephrasing this to be something like "See [the upstream documentation](thelink) for more fun..."

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+> const signers = await ethers.getSigners()
+undefined
+> await signers[0].sendTransaction({to:"0x08293b196E8F1c5552e455CFD10B642EC7a809A7", value:ethers.utils.parseUnits("500.99").toHexString()});
+{
+    ...
+}
+```
+
+Note that the `>` symbol is the hardhat prompt. The two commands to issue are `const signers = await ethers.getSigners()` and then `await signers[0].sendTransaction(...)`
```
HalosGhost: This probably should preface the codeblock in-question so it serves as instruction; that'd also serve as an opportunity to remove the repetition without any loss of clarity.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+Replace `0x08293b196E8F1c5552e455CFD10B642EC7a809A7` by your MetaMask address.
+
+You should see the balance appear after a few moments.
+
+### Try deploying en ERC20 token!
+
+Modify the provided contracts/MITCoin.sol file to deploy your own ERC20 token using the below directions, or create one [from scratch](https://dev.to/yakult/a-concise-hardhat-tutorial-part-2-writing-erc20-2jpm).
+
+First, update the provided deploy script (`scripts/deploy.js`) to reflect the new ERC20 token.
```
HalosGhost: This is probably clear for people who are quite familiar with solidity, but probably isn't enough information otherwise; either, we should link to upstream documentation strategically to onboard people as-necessary, or we should be far more specific.

Note also that, in following these instructions, I received a `contracts` and `scripts` directory underneath the playground, but these instructions seem to refer to those directories as they exist in the root of the repo; either way is fine, but we should ensure the right files exist in the right places for all the commands to work.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+The token is now deployed on the network! Issue some to your metamask account:
+
+*The below example assumes that the ERC20 has a function `mint(address, amount)`*
+
+```
+npx hardhat console
+> const yourToken = await (await ethers.getContractFactory("*yourTokenName*")).attach("*yourTokenAddress noted above*");
+undefined
+> await cbdcToken.mint("*your metamask address*", ethers.utils.parseUnits("5000"))
```
HalosGhost:
```suggestion
> const yourToken = await (await ethers.getContractFactory("*yourTokenName*")).attach("*yourTokenAddress noted above*");
undefined
> await yourToken.mint("*your metamask address*", ethers.utils.parseUnits("5000"))
```

just renamed the variable in the second line to match the declaration from the first.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+
+*Take note of the token address*
+
+The token is now deployed on the network! Issue some to your metamask account:
+
+*The below example assumes that the ERC20 has a function `mint(address, amount)`*
+
+```
+npx hardhat console
+> const yourToken = await (await ethers.getContractFactory("*yourTokenName*")).attach("*yourTokenAddress noted above*");
```
HalosGhost: Note that "token address" doesn't appear anywhere in the output from the commands. Let's make sure the specific properties are referred to correctly (ideally with example output, but as long as the names are correct, examples are a stretch goal).

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+## Running the Server
+### Run with Docker
+
+Spawns an agent running the EVM runner by default. Lua runner may be used by modifying the --runner_type
+flag in the agent0 command argument, in the file `docker-compose-3pc.yml`.
+
+### Usage pre-merge (current state)
+The Dockerfile uses a remote image as the base image (`ghcr.io/mit-dci/opencbdc-tx-base:latest`). This will be updated when this branch is merged into trunk.
```
HalosGhost: In the future, this should be avoided. Reviewers should be reviewing documentation as it will exist in the repo; exceptional instructions, if they need to exist for testing, should be provided in the PR body/comments.

scripts/3pc-run-local.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+IP="127.0.0.1"
+PORT="8888"
+RUNNER_TYPE="evm"
+LOGLEVEL="WARN"
+
+function print_help() {
+    echo "Usage: myscript.sh [OPTIONS]"
```
HalosGhost: > myscript.sh

scripts/lint.sh
```diff
@@ -4,7 +4,7 @@ set -e
 echo "Linting..."
 
 check_files=$(git ls-files \
-              | grep -v -E ".jpg|3rdparty" | cat)
+              | grep -v -E ".jpg|.pdf|3rdparty" | cat)
```
HalosGhost: This should be reverted at the same time as all binaries are removed.

scripts/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
...
+    usage
+fi
+
+WAITFORIT_TIMEOUT=${WAITFORIT_TIMEOUT:-15}
+WAITFORIT_STRICT=${WAITFORIT_STRICT:-0}
+WAITFORIT_CHILD=${WAITFORIT_CHILD:-0}
+WAITFORIT_QUIET=${WAITFORIT_QUIET:-0}
+
+# Check to see if timeout is from busybox?
+WAITFORIT_TIMEOUT_PATH=$(type -p timeout)
```
HalosGhost, line 145: Do we want this variable definition that gets immediately overwritten?

src/3pc/agent/impl.cpp
```diff
@@ -0,0 +1,679 @@
...
+            [this](broker::interface::commit_return_type commit_res) {
+                if(commit_res.has_value()) {
+                    if(std::holds_alternative<
+                           runtime_locking_shard::shard_error>(
+                           commit_res.value())) {
+                        auto err
+                            = std::get<runtime_locking_shard::shard_error>(
+                                commit_res.value());
+                    }
+                }
```
HalosGhost: Is this block necessary? It seems like `err` is unused, and getting that value seems to be the only point of this block.

src/3pc/agent/impl.cpp
```diff
@@ -0,0 +1,679 @@
...
+            case state::rollback_sent:
+                m_log->fatal("Result reported in rollback_sent state");
+            case state::rollback_complete:
+                if(!std::holds_alternative<error_code>(m_result.value())
+                   || std::get<error_code>(m_result.value())
+                          != error_code::retry) {
+                    m_log->fatal("Result reported in rollback_complete state "
+                                 "when result is not retry");
+                }
+                [[fallthrough]];
```
HalosGhost: Since these cases should all represent fatal bugs, is there a reason we fall-through to `break`, seemingly to retry, instead of exiting?

Additionally, since all these cases fall-through, would hitting one of the earlier cases result in all these error messages being logged?

src/3pc/agent/impl.cpp
HalosGhost, line 1: Not actually a comment on specifically this file, though it stuck out particularly as I was reviewing this one: `[[fallthrough]]` should *always* [be explicit](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-break).

src/3pc/agent/runners/evm/host.hpp
```diff
@@ -0,0 +1,210 @@
...
+        /// \param is_readonly_run true if no state changes should be applied.
+        /// \param ticket_number ticket number for transaction.
+        evm_host(std::shared_ptr<logging::log> log,
+                 interface::try_lock_callback_type try_lock_callback,
+                 evmc_tx_context tx_context,
+                 evm_tx tx,
+                 bool is_readonly_run,
+                 interface::ticket_number_type ticket_number);
+
+        [[nodiscard]] auto
```
HalosGhost: Here, below, and elsewhere, public methods should all have doc comments.

(non-blocking, but strongly encouraged)

src/3pc/agent/runners/evm/init_addresses.hpp
HalosGhost, line 1: Is there any reasoning/clarification/documentation for why these addresses rather than others, or why any addresses need to have accounts minted for them in-advance?

I see that it's used by `mint_initial_accounts()`; is this for ease-of-testing/dive-in?

src/3pc/agent/runners/evm/serialization.hpp
```diff
@@ -0,0 +1,171 @@
...
+    auto tx_encode(const cbdc::threepc::agent::runner::evm_tx& tx,
+                   uint64_t chain_id = opencbdc_chain_id,
+                   bool for_sighash = false) -> cbdc::buffer;
+
+    /// Converts a given buffer to an evm_tx
+    /// \param buf buffer containing the transaction to decode
+    /// \param logger logger to output any parsing errors to
+    /// \param chain_id the expected chain ID for the transaction. If the
+    // transaction contains a different chain ID this method will return
+    // std::nullopt
```
HalosGhost: Here, and below, there are broken doc comment line continuations.

src/3pc/agent/runners/interface.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+        interface(const interface&) = delete;
+        auto operator=(const interface&) -> interface& = delete;
+        interface(interface&&) = delete;
+        auto operator=(interface&&) -> interface& = delete;
+
+        /// Begins function execution. Retrieves the function bytecode using a
+        /// read lock and executes it with the given parameter.
+        /// \return true.
+        virtual auto run() -> bool = 0;
```
HalosGhost: Is there a reason to have this return a value at all? The core interface, and both runner implementations define it to always return `true`.

src/3pc/directory/impl.hpp
```diff
@@ -0,0 +1,35 @@
...
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef OPENCBDC_TX_SRC_3PC_DIRECTORY_IMPL_H_
+#define OPENCBDC_TX_SRC_3PC_DIRECTORY_IMPL_H_
+
+#include "interface.hpp"
+
+namespace cbdc::threepc::directory {
+    /// Implementation of a directory. Uses siphash to uniformly map keys to
```
HalosGhost: Nit: this will only be uniformly-distributed if `m_n_shards` is a power-of-two; else, there will be bias introduced from `%`. In this case, I assume this is likely immaterial since it seems unlikely that anything in the system depends on key distribution being uniform so much as being generally load-balanced across the shards.

src/3pc/runtime_locking_shard/state_machine.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "util/rpc/format.hpp"
+#include "util/serialization/format.hpp"
+#include "util/serialization/util.hpp"
+
+namespace cbdc::threepc::runtime_locking_shard {
+    auto state_machine::commit(uint64_t log_idx, nuraft::buffer& data)
+        -> nuraft::ptr<nuraft::buffer> {
+        auto maybe_req = from_buffer<rpc::replicated_request>(data);
+        if(!maybe_req.has_value()) {
+            // TODO: This would only happen if there was a deserialization
```
HalosGhost: Have the `TODO`s sprinkled throughout the code been evaluated for whether or not they should be made into issues or addressed now rather than later?

src/3pc/ticket_machine/interface.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <optional>
+#include <variant>
+
+namespace cbdc::threepc::ticket_machine {
+    /// Type alias for a ticket number.
+    using ticket_number_type = uint64_t;
+
+    /// \brief Ticket machine interface.
+    ///
+    /// Returns batches of monotonically increasing ticket numbers for
```
HalosGhost: Nit: this is not monotonically increasing; it's expected to wrap (just not frequently).

src/3pc/util.cpp
```diff
@@ -0,0 +1,247 @@
...
+            auto token = s.substr(pos_start, pos_end - pos_start);
+            pos_start = pos_end + delim.size();
+            ret.emplace_back(token);
+        }
+
+        ret.emplace_back(s.substr(pos_start));
+        return ret;
+    }
+
+    auto parse_args(int argc, char** argv)
```
HalosGhost: This is a little error-prone and definitely non-standard. We need to reunify configuration between the new architecture and the old infrastructure; when we do, we should also settle on a standardized argument/option parser.

(non-blocking)

src/util/rpc/http/kqueue_event_handler.cpp
```diff
@@ -0,0 +1,113 @@
...
+namespace cbdc::rpc {
+    kqueue_event_handler::~kqueue_event_handler() {
+        close(m_kq);
+    }
+
+    auto kqueue_event_handler::init() -> bool {
+        if((m_kq = kqueue()) == -1) {
+            return false;
+        }
+        return true;
```
HalosGhost:
```suggestion
        m_kq = kqueue();
        return m_kq != -1;
```

Style nit; no reason to prefer inline-assignment inside the `if`, especially when the `if` is of the always-redundant form `if (condition) { return bool; } else { return bool; }`

src/util/rpc/http/kqueue_event_handler.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    void kqueue_event_handler::register_fd(int fd, event_type et) {
+        switch(et) {
+            case event_type::remove: {
+                struct kevent ev {};
+                EV_SET(&ev, fd, EVFILT_READ, EV_DELETE, 0, 0, 0);
+                m_evs.emplace_back(std::move(ev));
+            }
+                {
```
HalosGhost, line 40: If `clang-format` is responsible for this really odd formatting, we should figure out how to fix it.

(long-term todo/issue; non-blocking)

tools/bench/3pc/lua/lua_bench.cpp
```diff
@@ -0,0 +1,350 @@
...
+        std::numeric_limits<size_t>::max(),
+        shards,
+        ticketer,
+        directory,
+        log);
+
+    auto pay_contract = cbdc::buffer();
+    pay_contract
+        = cbdc::buffer::from_hex(
+              "1b4c7561540019930d0a1a0a040808785600000000000000000000002877400"
```
HalosGhost: Here, and elsewhere, can we avoid using particularly large magic hex strings? Ideally, we'd have the plaintext version in source-control, and compile/convert/serialize them appropriately at init.

tools/bench/3pc/evm/contracts.hpp
HalosGhost, line 1: This file, because it is generated, is partly exempt from the hope to have all public methods documented. However, given that it is generated, should we actually check it in at all, or should we add a build-step so that this file is *always* generated as-needed?

(non-blocking)

HalosGhost (changes requested):
Several parts of the documentation need a bit of attention before merge; however, the vast majority of the content looks solid.

I'll also add a reminder note that there are commits which must be squashed before merge as well.

src/3pc/util.cpp
AlexRamRam: Agreed. Also, keeping in mind that corresponding changes to test controller are likely required.

tools/bench/3pc/evm/contracts.hpp
AlexRamRam: Fully agree in principle (to have user build/generate this). This would add more dependencies to the build (node, hardhat, etc), but cleaner and worth considering.

docs/programmability_architecture_diagram.jpg
AlexRamRam: Git is not for derived elements - agreed.
In this case - due to time and other constraints - the architecture diagram was generated using a vendor software program. If the value of this diagram does not adequately justify its existence as is, we can remove it. Or, we can accept it as the source image at least temporarily until we redraw using another tool.

scripts/lint.sh
AlexRamRam: pdf is not necessary and will be removed. Discussions about new .jpg diagrams are elsewhere.

src/3pc/runtime_locking_shard/state_machine.cpp
AlexRamRam: Good point that some of the TODO's can be made more transparent as GH issues, though certainly not all. What has been the policy for prior commits?

src/3pc/ticket_machine/interface.hpp
AlexRamRam: I believe the existing comment is appropriate for a general purpose statement for the interface. Note that this detail (wrapping, etc) are documented in `get_ticket_number()`

src/3pc/directory/impl.hpp
AlexRamRam: `... uniformly map keys to shard IDs based on a fixed number of shards`: seems correct to me (perhaps out of ignorance). @HalosGhost: if a value-adding correction is needed please make the change, or close.

src/3pc/agent/runners/interface.hpp
AlexRamRam: This is an appropriate interface to return a system level failure (although not used). However, should be `[[nodiscard]]` and aided by comments.

src/3pc/agent/impl.cpp
AlexRamRam: Not needed - will be removed. Handled in `handle_commit()`

src/3pc/agent/impl.cpp
AlexRamRam: Note that `m_log->fatal(...)` calls system `exit()` (i.e. execution does not return to this stack, there is no retry).

The log object having the power to system exit is surprising - something we can discuss and potentially change. e.g. rename `fatal()` to `fatal_system_exit()`

It's curious that we're not getting the `fallthrough` warnings, some cleanup is needed here.

scripts/wait-for-it.sh
AlexRamRam: It's used as input in the next line - can be prettier, but ok with me.

docs/programmability_architecture.md
AlexRamRam: Paren is necessary - but it's currently part of Lua hyperlink (asymmetric). Subtle - hard to see in the viewable doc. Good catch.
Excellent point on keeping single sentence per line.

docs/programmability_architecture.md
AlexRamRam: Paren is necessary - but it's currently part of Lua hyperlink (asymmetric). Subtle - hard to see in the viewable doc. Good catch.

benchmarks/transactions.cpp
AlexRamRam: Cannot use UL suffix here for the 2nd value because it's not a literal.

src/3pc/util.cpp
HalosGhost: I mentioned this as an issue without offering a clear action to close review item; sorry about that!

In short, I mentioned this because I think the addition of this architecture exacerbates the existing DIY configuration system, which motivates our trying to solve the problem. I think a new issue is warranted, targeting a unifying configuration system for the three architectures we have, and with space for more. (Probably referencing #173 and/or #81.) At some point, before merge, that issue should be created (or establishing an alternative path toward handling this inconsistency; for example, an official sunset, or stable release plan---these are not the only options, of course).

Hopefully that clarifies the purpose of this task, as well as a clear completion target.

tools/bench/3pc/evm/contracts.hpp
HalosGhost: Mmm, very good point. I don't love adding new dependencies (and this PR includes quite a few additional build-time ones already). I suppose I hadn't considered they had been excluded for that reason because we already pull in several EVM-specific dependencies at configure-time. It's for this same reason I hadn't really considered there to be a separate lifecycle phase between building the full repository at the top-level and doing anything lower down.

Do you think we should make that explicit? If so, would the reasonable separation be between the C++ architecture at build-time, and then runner-specific build-time should happen later and on-demand? This would allow people who are interested only in, say, the EVM to avoid downloading lua-dependencies if they have no intention of using those (or, of course, vice versa). This would require a bit of build-system work (e.g., some `./configure.sh`-style infrastructure for each runner).

Alternatively, we could life all the dependencies from all committed runners to top-level, and unify the build-times. This is *much* easier, both to write, and to maintain. The downside is that, the officially-supported build infrastructure will assume all users will download all dependencies of all committed code. And, to be fair, that might not be a sustainable long-term view. (I know it doesn't immediately feel scalable or particularly frugal.) However, because not a lot of work has been done yet on planning for the long-term maintainability of multiple, potentially wildly-disparate architectures, I do not think this particular case needs to be solved right now, and I would definitely accept such a simplification as a valid solution for now.

src/util/rpc/http/kqueue_event_handler.cpp
AlexRamRam: Agreed on separating `m_kq = kqueue();`. However, I suggest we keep the if/else structure for clarity. At least for me reading `return m_kq != -1` (unlike, let's say `return x==y`), despite its brevity, requires some brain cycles.

tools/bench/3pc/evm/contracts.hpp
HalosGhost: Or, is the separation as simple as "all the C++ stuff" building at top-level and "all the other stuff" separate? I *think* that accurately captures the current separation. If so, and that's the intention, then the path to close this review task is far simpler: just make it explicit. Namely, add a step to the README's walk-through for the additional steps (or, if we separate the build guides out entirely, mention that sub-directories may have a required build-step). (This is the documentation task that, if one of the other splits makes sense to pursue, might still be required to complete.)

src/util/rpc/http/kqueue_event_handler.cpp
HalosGhost: (Sorry if this seemed silly; I realize it probably did. It's just one of those ones that bothers me for some reason. Take nothing personal about it, please!)

src/3pc/agent/impl.cpp
AlexRamRam: ... I can think of cases where I would prefer to not see `[[fallthrough]]`

tools/bench/3pc/evm/contracts.hpp
AlexRamRam: Currently the "other stuff" (i.e. creating the contract ABI for C++) has to be done first to build the "C++ stuff" like `evm_bench` and the 3pc integration test. Instead of performing that step we have committed the derived c++ element. That part is well documented and can easily be regenerated.

src/3pc/directory/impl.hpp
HalosGhost: Sorry, allow me to clarify: [`impl.cpp`:15](https://github.com/mit-dci/opencbdc-tx/pull/213/files#diff-77a7dfffad078695ca8ec3a68263414d82b8790b84ffdc33b307689e2b69eba9R15) divides the key generated by the number of shards running to select the shard for load-balancing. This constrains the random-number-generator of the hash function to the range of running shard ids (i.e., `[0, m_n_shards - 1]`). However, it introduces "modulo bias".[^1]

[^1]: https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator

I mentioned this because the comment is inaccurate. However, as I mentioned, I do not know if any security properties of the system depend on this property (i.e., I am not sure if this load-balancing is required to be perfectly uniformly distributed). If it does not matter, and this is either acceptably uniform for the purposes of the architecture, or the uniformity of the distribution is only desirable, not necessarily something that must be achieved, then that's not a problem! The comment should be corrected though.

However, if the load-balancing here is needed to be uniformly random, I see two valid directions: 1) we should document that the number of shards is required to be a multiple of the range of the hash-function. This is equivalent to taking the `n`-least-significant bits as the shard index. Or, 2) we modify `key_location()` cut the bias from this selection. This could be done either with the STL's templates for random-number generation (which we use extensively elsewhere in this codebase, for easy reference; [Edit]: or maybe one of our secure random sources would be even better), or, say with rejection sampling into a large range known to be a multiple of `m_n_shards`, and then taking the modulus just as it does now. That also should perfectly remove the bias since it creates an equivalent case to `m_n_shards` being a multiple of the hash-function's range.

If we don't know, or don't need to know at this level of depth, whether or not this property is important to the system's security properties, we could potentially close this task by modifying the comment to be accurate. That's not a hard fix or unreasonable, but I don't know if it's the right choice. Definitely happy for alternative actions too, of course.

src/util/rpc/http/kqueue_event_handler.cpp
HalosGhost: It bothers me, but as it is marked as a nit, you can safely read it as "I'm not particularly moved by this one; take it or leave it"; so, if you don't find this compelling, please do feel close it and I won't be mad in any fashion.

scripts/wait-for-it.sh
HalosGhost: Definitely a misread from me. Ignore this noise.

src/3pc/agent/impl.cpp
HalosGhost: Funny that you mention it is surprising `logger::fatal()` is able to fail. That surprises me right now, as a matter of fact. That's also on me for not having investigated whether or not it would terminate; clearly came from an untested response. Consider adding a closure task for this review item to open a bug for potentially reconsidering fatal-logging's immediate termination. Maybe it's not large enough a deal to worry about, but I'd like to not forget about it and avoid revisiting due to attrition.

src/3pc/agent/impl.cpp
HalosGhost: Not currently a hard-and-fast rule, we generally try to adhere to the [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines). That generally governs our choices for linting options, as it helps us define a reasonable code quality and formatting standard.

To butcher a saying from the Go programming community, "The CppCoreGuidelines are no one's favorite formatting/practice guide; but having a standard practice guide to follow is everyone's favorite." Consider this case to be closer to a nit than not.

src/3pc/runtime_locking_shard/state_machine.cpp
HalosGhost: This is also something we don't have a hard and fast rule for. I do want to, however, try to help give people easier ways to dive into the code, as well as to unify on as close to a single source of truth for people to discover a comprehensive list of known improvements available to be made to the code base.

I recognize that some things are going to be worthy of being added as items to #3, and nothing more, some probably aren't worth any additional documentation at all (in which case, maybe we should consider removing the TODO?).

This is both a question of "What is the list of TODOs which we think should be turned into issues to increase their accessibility?" (Which includes the possibility "there aren't any worth bringing over.") and a request to improve the documentation for those that are worth keeping.

The invovled part is actually determining which of the newly-added TODO lines are worth trying to document more sustainably/visibly. Once that's been determined, and recorded, even adding those items to an omnibus issue (similar to #3) would be a reasonably closure path.

src/3pc/util.cpp
AlexRamRam: #218

src/3pc/runtime_locking_shard/state_machine.cpp
hgyami: @AlexRamRam @maurermi can you put together a comprehensive list of the TODOs in one place that we can review in the eng meeting tomorrow? Will also raise with Bob - will be helpful as we build our research agenda/roadmap moving forward!

src/3pc/agent/runners/evm/init_addresses.hpp
AlexRamRam: It's convenient and useful to have fully deterministic execution for testing. I'm ok with this.

HalosGhost: I have no issue with using these for testing. If this is exclusively for testing, is there a reason it must be in `src`? Either way, can a doc comment be added explaining what these are for? (That's the easiest path to close this review task in my opinion.)

docs/programmability_architecture_diagram.jpg
AlexRamRam: Investigating options:
* use mermaid & commit source
* commit svg output created by [inkscape](https://inkscape.org/)

src/3pc/agent/impl.cpp
AlexRamRam: Will create an issue item for this and close.

src/3pc/agent/impl.cpp
AlexRamRam: Create issue item for this & close

src/3pc/agent/runners/evm/init_addresses.hpp
AlexRamRam: Will document the intent of this and close

src/util/rpc/http/kqueue_event_handler.cpp
AlexRamRam: will first run clang-format on this and see if it fixes it. otherwise will create an issue item and close

tools/bench/3pc/lua/lua_bench.cpp
maurermi: Currently working on a fix for this

src/3pc/agent/impl.cpp
AlexRamRam: #219

src/3pc/agent/impl.cpp
AlexRamRam: #220

src/util/rpc/http/kqueue_event_handler.cpp
AlexRamRam: #221

src/3pc/agent/impl.cpp
AlexRamRam: Following is the case I had in mind where I would prefer to not see `[[fallthrough]]`. [CppCoreGuidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) is OK with this and `clang-tidy` requires them to be removed (complains when `switch has 2 consecutive identical branches`):

```
Multiple case labels of a single statement is OK:

switch (x) {
case 'a':
case 'b':
case 'f':
    do_something(x);
    break;
}
```

tools/bench/3pc/lua/lua_bench.cpp
maurermi: As a follow up, is there any reason that comes to mind for store the bytecode in plaintext anywhere? It's fairly easy to generate the bytecode at runtime by leveraging Lua's embedded scripting capabilities and then we don't need to store this massive string anywhere persistently. 

So far I have two solutions: one where this bytecode string is generated during the build process and is stored persistently in a text file, which is then read in at runtime -- and another where the bytecode is generated at runtime (not requiring any external storage, or large static strings). I prefer the second solution but if there's any reason to store the bytecode persistently, I would be interested to hear it.

HalosGhost: Unless any of the generation would happen during performance measurements, I see no reason to keep around the hex-version if it's just as easy to keep around the source.

AlexRamRam: Unless there is a reason to re-compile every time, compile once and keep the derived (non-git element) persistent file.

maurermi: @AlexRamRam Sorry just now seeing this -- I would say in regards to the bench program it makes sense to interpret the Lua script at runtime because it allows easier interaction with different Lua scripts which could be useful, but in terms of the unit test set up I can see interpreting the script at runtime being unnecessary

maurermi: ^^ Also -- not sure if I did a good job explaining this part, but the bytecode is only generated once at runtime (in the setup for either the bench test or unit test) and then it exists statically for the remainder of the test, it is not being regenerated each time the bytecode is used

src/3pc/agent/runners/evm/host.hpp
maurermi: The non-documented public methods in host.hpp are all inherited methods from the class evmc::HostInterface in evmc.hpp, which then refers using @copydoc to the associated functions in evmc.h 

Would you recommend that in this case, we include new doc comments within the OpenCBDC codebase for these inherited functions? Curious what the protocol is here

docs/programmability_architecture_diagram.jpg
AlexRamRam: As an imperfect but acceptable solution, replaced .jpg binaries with formatted .svg files which can be loaded into an open source design tool such as [Inkscape](inkscape.org) and be modified.

tools/bench/3pc/evm/contracts.hpp
maurermi: Have we decided how to proceed in regard to this issue (or whether we wish to at all)?

HalosGhost: I think making the current “C++ gets built at top-level, with all deps as-needed for that goal / subdirs might need other deps” explicit is the easiest, acceptable solution (likely taking the form of an **N.B.,**-style comment (which we have in several places in our docs already) in the walkthrough mentioning that modifying the system may require other dependencies, documented where-relevant).

However, I think this separation is far from ideal; personally, the following three splits are the best ergonomically:

* no separation; all deps to run/modify all code are handled by `./scripts/configure.sh` (con: wasteful; pro: single step for everyone)
* separate at the runner-level (con: big lift, requires splitting tests/deps-handling, etc.; pro: super-clear why the split is where it is, and people can install only the things they care to use
* do both! Modularize each piece so that `./scripts/configure.sh` (or, more likely, its replacement) has feature-flags which determine what gets brought in and built (this is actually what `configure.sh` does more normally) (con: massive reworking of the build-system; pro: it's much closer to the standard build experience, and also reaps all the pros of both other separations)

I think, especially given our hope to merge this ASAP, the first solution (which is just adding a single line in the docs) is probably most reasonable. I would ask that an additional closure task be added for this thread: namely, opening an issue to revisit the build system for one of these three alternative separations.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
+# Programmability User Guide
+
+This guide runs through setting up the 3PC system with the EVM runner, and then interacting with it using Hardhat and MetaMask. This assumes that the system has been set up as described in the [README](https://github.com/mit-dci/opencbdc-tx#readme).
+
+## Prerequisites
+
+Make sure you have [NodeJS installed](https://github.com/nodesource/distributions/blob/master/README.md).
+We reccomend using `v18` or `v20`, as Hardhat [actively supports these versions](https://github.com/nodejs/release#release-schedule) currently.
```
HalosGhost:
```suggestion
We recommend using `v18` or `v20`, as Hardhat [actively supports these versions](https://github.com/nodejs/release#release-schedule) currently.
```
typo nit.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
...
+However, to use the Hardhat compiler, all Solidity files should be stored in the
+`contracts` subdirectory.
+
+Copy the example [hardhat.config.js](https://github.com/mit-dci/opencbdc-tx/blob/trunk/scripts/hardhat.config.js) into this directory.
+
+Edit the `url:` value in the `hardhat.config.js` file to correspond with the url of the agent RPC server.  
+Using Docker this will be `http://localhost:8080/`. If running outside of
+Docker, use the IP and port specified
+ when running `./scripts/3pc-run-local.sh`
+(default is `http://localhost:8888/`).
```
HalosGhost:
```suggestion
Using Docker this will be `http://localhost:8080/`.
If running outside of Docker, use the IP and port specified when running `./scripts/3pc-run-local.sh` (default is `http://localhost:8888/`).
```
nit to fix some really odd line-breaking.

docs/programmability_architecture_diagram.svg
HalosGhost, line 1: Thank you for the switch to SVG, this is dramatically better! At some point, I wouldn't mind revisiting our documentation image artifacts to go to a more unified source form we can leverage universally, but this is a fabulous step towards that. :+1:

docs/programmability_architecture.md
```diff
@@ -0,0 +1,54 @@
...
+### Directory
+The _Directory_ identifies the target _Shard_ unit for each and every key-value data pair.
+This is a detail managed by the _Broker_, allowing the end-client (the _Agent_) to interact with the distributed _Shards_ as a single database unit.
+The _Directory_ is where the sharding logic is defined and its implementation can be designed to optimize database resources for particular use cases.
+### JSON-RPC Server
+The _JSON-RPC Server_ implements the [Ethereum JSON-RPC API](https://ethereum.org/en/developers/docs/apis/json-rpc/) over HTTP.
+This is the external interface compatible for external Ethereum clients such as wallets (e.g. Metamask), and development tools (e.g. Hardhat, Truffle, ethers.js, web3.js)
+### Runner
+The _Runner_ is an abstraction for retrieving and executing context-specific bytecode.
+The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua)](https://www.lua.org/) and demonstrates the flexibility of the architecture to support different program execution environments.
```
HalosGhost:
```suggestion
The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/developers/docs/evm/) and [Lua](https://www.lua.org/)) and demonstrates the flexibility of the architecture to support different program execution environments.
```

style nit that wasn't corrected.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
+# Programmability User Guide
+
+This guide runs through setting up the 3PC system with the EVM runner, and then interacting with it using Hardhat and MetaMask. This assumes that the system has been set up as described in the [README](https://github.com/mit-dci/opencbdc-tx#readme).
+
+## Prerequisites
+
+Make sure you have [NodeJS installed](https://github.com/nodesource/distributions/blob/master/README.md).
+We reccomend using `v18` or `v20`, as Hardhat [actively supports these versions](https://github.com/nodejs/release#release-schedule) currently.
+We also provide a [.nvmrc](https://github.com/mit-dci/opencbdc-tx/blob/trunk/.nvmrc).
```
HalosGhost: Thank you!

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
...
+| --------------- | ------------------------------------------|
+| Network name    | `OpenCBDC`                                |
+| Chain ID        | `0xcbdc`                                  |
+| New RPC URL     | `http://<agent IP address>:<agent port>/` |
+| Currency symbol | `CBDC`                                    |
+
+The agent IP address and agent port are the IP address and port specified
+for the agent when spawning the system.  
+Using Docker, this defaults to {IP: localhost, port: 8080}  
+Running outside of Docker, this defaults to {IP: localhost, port:8888}.
```
HalosGhost:
```suggestion
The agent IP address and agent port are the IP address and port specified for the agent when spawning the system.  
Using Docker, this defaults to {IP: localhost, port: 8080}.
Running outside of Docker, this defaults to {IP: localhost, port:8888}.
```
Another really odd line-breaking pattern.

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
...
+Note: To compile the example contract, `@openzeppelin/contracts` is a required install.
+```console
+$ npm install @openzeppelin/contracts
+```
+Create a deploy script for the token. An example is provided in [opencbdc-tx/scripts/deploy.js](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/programmability_user_guide.md).  
+As written, this script requires the name of the ERC20 token to be changed to match the name of the token which is to be deployed.  
+
+Note: When the Hardhat project was initialized, a `scripts/deploy.js` file was created,
+which corresponds to the file `contracts/Lock.sol`   
+that was also created upon initialization.
```
HalosGhost:
```suggestion
Note: When the Hardhat project was initialized, a `scripts/deploy.js` file was created.
It corresponds to the file `contracts/Lock.sol` that was also created upon initialization.
```

docs/programmability_user_guide.md
```diff
@@ -0,0 +1,165 @@
...
+## Prerequisites
+
+Make sure you have [NodeJS installed](https://github.com/nodesource/distributions/blob/master/README.md).
+We reccomend using `v18` or `v20`, as Hardhat [actively supports these versions](https://github.com/nodejs/release#release-schedule) currently.
+We also provide a [.nvmrc](https://github.com/mit-dci/opencbdc-tx/blob/trunk/.nvmrc).
+## Running the Server
+### Run with Docker
+
+Spawns an agent running the EVM runner by default. Lua runner may be used by modifying the --runner_type
+flag in the command keyed by agent0, in the file [`docker-compose-3pc.yml`](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docker-compose-3pc.yml).
```
HalosGhost:
```suggestion
Spawns an agent running the EVM runner by default.
Lua runner may be used by modifying the `--runner_type` flag in the command keyed by agent0, in the file [`docker-compose-3pc.yml`](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docker-compose-3pc.yml).
```

scripts/3pc-run-local.sh
```diff
@@ -0,0 +1,48 @@
...
+mkdir -p logs
+echo Running agent on $IP:$PORT
+echo Log level = $LOGLEVEL
+echo Runner type = $RUNNER_TYPE
+
+./build/src/3pc/runtime_locking_shard/runtime_locking_shardd --shard_count=1 --shard0_count=1 --shard00_endpoint=$IP:5556 --shard00_raft_endpoint=$IP:5557 --node_id=0 --component_id=0 --agent_count=1 --agent0_endpoint=$IP:6666 --ticket_machine_count=1 --ticket_machine0_endpoint=$IP:7777 --loglevel=$LOGLEVEL > logs/shardd.log &
+sleep 1
+./scripts/wait-for-it.sh -s $IP:5556 -t 60 -- ./build/src/3pc/ticket_machine/ticket_machined --shard_count=1 --shard0_count=1 --shard00_endpoint=$IP:5556 --node_id=0 --component_id=0 --agent_count=1 --agent0_endpoint=$IP:6666 --ticket_machine_count=1 --ticket_machine0_endpoint=$IP:7777 --loglevel=$LOGLEVEL > logs/ticket_machined.log &
+sleep 1
+./scripts/wait-for-it.sh -s $IP:7777 -t 60 -- ./scripts/wait-for-it.sh -s $IP:5556 -t 60 -- ./build/src/3pc/agent/agentd --shard_count=1 --shard0_count=1 --shard00_endpoint=$IP:5556 --node_id=0 --component_id=0 --agent_count=1 --agent0_endpoint=$IP:$PORT --ticket_machine_count=1 --ticket_machine0_endpoint=$IP:7777 --loglevel=$LOGLEVEL --runner_type=$RUNNER_TYPE > logs/agentd.log &
```
HalosGhost: maintainability note (consider this slightly above a nit), it's dramatically harder to read really long source lines (which is why our linter forces line-breaks in the C++ code). For shell, this is traditionally improved by adding an escaped line-break after each option.

src/3pc/agent/runners/evm/host.hpp
HalosGhost: Discussed offline. For any method including `override`, it is assumed the core documentation is elsewhere, and only relevant deviations from the overridden method are documented inline.

In this case, these methods are overriding for the purpose of providing our own implementation of an external interface, so the deviations are all implementation-details. To that end, it's fair for these to remain undocumented here.

@AlexRamRam and @maurermi, feel free to chime in here with any thoughts or clarifications if I missed any context which should be here for traceability.

HalosGhost (approved):
I have a few very small style/typo nits, none of which are particularly high-priority.

Assuming the final bits of documentation follow all the same criteria current docs have gone through, that the renaming goes without a hitch (@maurermi and I chatted about strategy for this a bit offline, and I think there's a good plan in-place), and that the commits are squashed appropriately, I think this is set to merge. :+1:

maurermi (approved):
T-ACK. I think this is set to merge pending the name change.

HalosGhost:
Finalized reconciliation of rename, and updated docs. :+1: from me. Ready for merge.

Issue #218, opened 2023-06-05
Unify configuration systems among all architectures


AlexRamRam:
### Affected Branch

programmability

### Basic Diagnostics

- [X] I've pulled the latest changes on the affected branch and the issue is still present.

- [X] The issue is reproducible in docker

### Description

The Programmability architecture introduces a different way of configuring the system from the existing architectures (i.e. command line args vs. config files). Unify the configuration systems and consider the broader set of questions raised here: #173, #81 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #228, opened 2023-09-06
fix: ensure correct handling of mock endpoints

HalosGhost:
This ensures that, regardless of what has been specified for mocking, any module with no endpoints defined in the config will have a server started for it in a mock system test.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -10,7 +10,12 @@ Checks: '
     -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
     readability-*,
     google-explicit-constructor,
-    -readability-identifier-length'
+    -readability-identifier-length,
+    -cppcoreguidelines-avoid-const-or-ref-data-members,
+    -bugprone-unchecked-optional-access,
+    -cppcoreguidelines-avoid-do-while,
+    -modernize-use-emplace,
+    -readability-convert-member-functions-to-static'
 CheckOptions:
     - {key: cppcoreguidelines-explicit-virtual-functions.AllowOverrideAndFinal, value: true}
     - {key: modernize-use-override.AllowOverrideAndFinal, value: true}
--- a/tests/integration/mock_system.cpp
+++ b/tests/integration/mock_system.cpp
@@ -31,29 +31,42 @@ namespace cbdc::test {
         const std::unordered_set<mock_system_module>& disabled_modules,
         config::options opts)
         : m_opts(std::move(opts)) {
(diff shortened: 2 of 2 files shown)


HalosGhost:
awaiting fix for test breakage in 211bbc9 to rebase and rerun CI.

AlexRamRam (commented):
utACK

The intended changes look solid, as intended.

A comment that may be beyond the scope of this PR:
It may be valuable if the `mock_system` provides some level of self-integrity by ensuring that the input configuration is at least theoretically workable for one of the existing architectures. This would require some documentation and checks with error messages. I may be missing some context, but the self-integrity feature may make this module even more compelling, especially as a c++ module as opposed to a shell utility aimed to serve a similar purpose.

HalosGhost:
@AlexRamRam note that the mock_system is a testing utility which allows us to mock N components while running one or more “real” components so we can do integration/boundary tests.

However, a utility that parses a config and determines if it is even statically-valid for the specified architecture would be an excellent tool to have. (Though, it probably requires us to finalize an answer to #218 before we could reasonably implement one.)

.clang-tidy
```diff
@@ -10,7 +10,12 @@ Checks: '
     -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
     readability-*,
     google-explicit-constructor,
-    -readability-identifier-length'
+    -readability-identifier-length,
+    -cppcoreguidelines-avoid-const-or-ref-data-members,
+    -bugprone-unchecked-optional-access,
+    -cppcoreguidelines-avoid-do-while,
+    -modernize-use-emplace,
+    -readability-convert-member-functions-to-static'
```
maurermi, line 18: I don't feel particularly strongly about removing any of these, but since some time has passed since this was PRed maybe it's worth reconsidering whether we remove these or not.

Unchecked optional access and modernize use emplace seem like two we might want to keep around

maurermi (approved):
T-ACK. Tests run as expected, and this improves our test suite by allowing endpoints to be mocked properly. Left one comment about the changes to .clang-tidy to consider but think this is good to merge.

.clang-tidy
HalosGhost: It's been long enough since I opened this PR, that I don't remember my exact reasoning; but in general, I'd always want to avoid removing tidy plugins. If my hazy memory serves correctly, I removed these because they were flagging quite a lot of things in the codebase, so there'll be some churn when we enable them.

Not a problem at all, but should be done in a different PR.

PR #230, opened 2023-09-16
Fix typo on  uhs architecture url

hendrikttan:
Signed-off-by: Hendrik Tan <tanjaya@gmail.com>

PR #234, opened 2023-09-22
Fix errant hardhat console instruction in parsec_user_guide

maurermi:
ethers.toBeHex(ethers.parseUnits()) no longer appears to be supported as written, modified to parseEther() for correct functionality.

### DIFF
--- a/docs/parsec_architecture.md
+++ b/docs/parsec_architecture.md
@@ -15,7 +15,7 @@ The system provides two _Runner_ implementations ([EVM](https://ethereum.org/en/
 
 # Key Features
 ### Flexible Data Model and Transaction Semantics
-In contrast to the [UHS-based architectures (Atomizer & 2PC)](architecture.md), this design does not dictate a particular data model or transactional semantics.
+In contrast to the [UHS-based architectures (Atomizer & 2PC)](uhs-architectures.md), this design does not dictate a particular data model or transactional semantics.
 The generic computation layer allows flexibility to employ a wide range of designs with minimal changes to this underlying C++ system.
 Smart contracts can be deployed and re-defined, thus making it possible to make fundamental changes to the system during runtime.
 


maurermi:
May also loop in #230 if DCO not fixed for an omnibus doc improvement PR

AlexRamRam (commented):
Note that `ethers.utils.parseEther` works in [ethers v5](https://docs.ethers.org/v5/api/utils/display-logic/#utils-parseEther), but is not defined in [ethers v6](https://docs.ethers.org/v6) (rather it is just `ethers.parseEther` see: https://docs.ethers.org/v6/api/utils/#parseEther)

```
> ethers.version
'6.7.1'
> ethers.parseEther
[Function: parseEther]
> ethers.utils
undefined
> ethers.utils.parseEther
Uncaught TypeError: Cannot read properties of undefined (reading 'parseEther')
```

More generally, [ethers v6](https://docs.ethers.org/v6) is not backward compatible with [v5](https://docs.ethers.org/v5). The main gist of the migration issues are explained [here](https://docs.ethers.org/v6/migrating/) but the doc is not comprehensive.

I recommend that we explicitly require ethers v6 (as in the [uniswap demo doc](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/uniswap_demo/uniswap_demo.md)) and make the necessary changes in the doc.

maurermi:
> Note that `ethers.utils.parseEther` works in [ethers v5](https://docs.ethers.org/v5/api/utils/display-logic/#utils-parseEther), but is not defined in [ethers v6](https://docs.ethers.org/v6) (rather it is just `ethers.parseEther` see: https://docs.ethers.org/v6/api/utils/#parseEther)
> 
> ```
> > ethers.version
> '6.7.1'
> > ethers.parseEther
> [Function: parseEther]
> > ethers.utils
> undefined
> > ethers.utils.parseEther
> Uncaught TypeError: Cannot read properties of undefined (reading 'parseEther')
> ```
> 
> More generally, [ethers v6](https://docs.ethers.org/v6) is not backward compatible with [v5](https://docs.ethers.org/v5). The main gist of the migration issues are explained [here](https://docs.ethers.org/v6/migrating/) but the doc is not comprehensive.
> 
> I recommend that we explicitly require ethers v6 (as in the [uniswap demo doc](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/uniswap_demo/uniswap_demo.md)) and make the necessary changes in the doc.

@AlexRamRam I agree, this is an oversight on my part. Thanks for catching this. I will make the corresponding updates shortly

PR #238, opened 2023-10-06
Upgrade from c++17 to c++20

### LINKED ISSUE #237: Upgrade build to c++20

AlexRamRam:
Fix #237

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ include(CheckCXXCompilerFlag)
 
 project(opencbdc-tx)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED True)
 set(CMAKE_CXX_EXTENSIONS False)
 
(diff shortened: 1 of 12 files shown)


maurermi:
concept-ACK, I think this is a good idea -- will test soon

HalosGhost:
Concept ACK (we had initially avoided 20 till Bitcoin switches, but given we now have dependencies which will require switching, I think that ship has sailed); reminder we need to test basic performance impact locally and in test controller.

maurermi (approved):
T-ACK. Tested using Docker and all tests pass. Did not explicitly evaluate performance impact at this time relative to the C++17 build but changes are minimal and it would be quite surprising if this impacted performance as far as I can tell. 

The question remains of when exactly we want to make this switch. Sooner seems better than later since later developments may depend on C++20. 

Edit: Appended this commit to trunk for the test, can confirm that this change works with current state of trunk.

HalosGhost:
tACK. I've checked rebased against the latest trunk, and done some local performance comparisons (no performance degradation is visible; if anything, I see slightly higher sustained throughput for identical configurations).

LGTM. Merging.

HalosGhost (approved):

Issue #236, opened 2023-10-05
Investigate alternatives to deeply nested callbacks (particularly in PArSEC) to improve code readability


AlexRamRam:
In contrast to the Atomizer/2PC architectures, PArSEC requires a more standard form of the [two-phase locking protocol](https://en.wikipedia.org/wiki/Two-phase_locking) where locks are sequentially accumulated as execution proceeds. As a result, the reader encounters deeper levels of nesting of callbacks in the PArSEC code base as they follow the flow of execution.

Using nested callbacks is sensible and efficient (avoids waiting or polling for tasks synchronously). However, the code becomes more difficult to follow, particularly for those with less experience with asynchchronous programming. This issue item asks the openCBDC community to investigate alternatives that aim to improve code readability without sacrificing performance.

One path is to use [chain-able/continuable futures](https://en.cppreference.com/w/cpp/experimental/future), which is not yet part of the C++ standard. This would allow us to chain asynchronous tasks, and might look like the following (ignoring parameterization of return values , inputs and other important details):
```
auto runTransaction = async( lockAddress() )
.then( lockTransactionID() )
.then( lockTicketID() )
.then ...
```

Until the c++ standard catches up, are there implementations of concurrency frameworks available today that help improve readability (without adding performance burdens) that also pass the necessary set of standards for openCBDC?. For example, I've experimented with one implementation here: https://github.com/Amanieu/asyncplusplus

Are there other solutions (i.e. a different threading model) that improves code readability without adding performance burdens?

PR #239, opened 2023-10-18
Aiming to improve readability of PArSEC (EVM) transaction flow

AlexRamRam:
Human readability, ease of access and maintainability are subjective to a large degree. An iterative approach of modest-scale changes with potential for discussions and feedback may be the way to go, particularly in a large open source community such as ours.

Here is one iteration of proposed changes from the perspective of following a EVM transaction flow in PArSEC.

Relates to issue #236 (does not solve it)

### DIFF
--- a/src/parsec/agent/impl.cpp
+++ b/src/parsec/agent/impl.cpp
@@ -40,8 +40,8 @@ namespace cbdc::parsec::agent {
         switch(m_state) {
             // In these states we can start again from the beginning
             case state::init:
-            case state::begin_sent:
-            case state::begin_failed:
+            case state::ticket_number_request_sent:
+            case state::ticket_number_request_failed:
                 break;
 
             // We already have a ticket number but need to start again
@@ -92,15 +92,16 @@ namespace cbdc::parsec::agent {
         }
 
         m_result = std::nullopt;
-        m_state = state::begin_sent;
-        auto success = m_broker->begin(
+        m_state = state::ticket_number_request_sent;
+        auto success = m_broker->get_new_ticket_number(
             [&](broker::interface::ticketnum_or_errcode_type res) {
-                handle_begin(res);
+                handle_new_ticket_number(res);
             });
 
         if(!success) {
-            m_state = state::begin_failed;
-            m_log->error("Failed to contact broker to begin");
+            m_state = state::ticket_number_request_failed;
+            m_log->error(
+                "Failed to contact broker to get a new ticket number");
             m_result = error_code::broker_unreachable;
             do_result();
         }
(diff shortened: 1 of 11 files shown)


src/parsec/agent/impl.cpp
```diff
@@ -237,88 +240,85 @@ namespace cbdc::parsec::agent {
...
-                                 m_ticket_number.value());
-                    auto reacquired = std::make_shared<std::atomic<size_t>>();
-                    for(auto& it : *reacq_locks) {
-                        m_log->trace("Re-acquiring lock on",
-                                     it.first.to_hex(),
-                                     "type",
-                                     static_cast<int>(it.second),
-                                     "for",
+
+        if(std::holds_alternative<broker::value_type>(res_variant)) {
```
maurermi, line 252: I personally find this easier to read with std::visit, as it makes clear that all possible variant types are covered

src/parsec/agent/runners/evm/http_server.hpp
```diff
@@ -170,11 +170,14 @@ namespace cbdc::parsec::agent::rpc {
             -> bool;
 
         static auto
-        handle_error(const Json::Value& params,
+        handle_error(const server_type::result_callback_type& callback,
```
maurermi, line 176: This is helpful for readability, but would be good to add a comment specifying that this function is overloaded (and perhaps that it calls the other version)

src/parsec/agent/runners/evm/impl.cpp
```diff
@@ -620,7 +628,7 @@ namespace cbdc::parsec::agent::runner {
         }
     }
 
-    void evm_runner::handle_lock_from_account(
+    void evm_runner::handle_lockfromaccount_and_continue_exec(
```
maurermi, line 631: For some of these longer names, it might be easier to just specify in a descriptive comment that this function ex. handles lock and then continues execution. I am of the opinion that we should do this anyway even for private methods but curious what others think

src/parsec/broker/impl.hpp
```diff
@@ -34,7 +34,9 @@ namespace cbdc::parsec::broker {
         /// \param result_callback function to call with the begin result.
         /// \return true if the request to the ticket machine was initiated
         ///         successfully.
-        auto begin(begin_callback_type result_callback) -> bool override;
+        auto get_new_ticket_number(
+            std::function<void(ticketnum_or_errcode_type)> result_callback)
```
maurermi, line 38: The nice thing about using an alias like `begin_callback_type` here is that it's easy to redefine what the type is later. Not sure if there's any reason to pass a function with a different return or argument type besides (`void`, `ticketnum_or_errcode_type`) but just something to consider. 

That being said, it is much easier to determine what the argument actually is when presented this way.

maurermi (commented):
UT-ACK, I find these changes to be largely helpful for readability, but left some comments to consider. Overall I would say these are welcome updates.

I would also say that some of the value added by changing function names could be achieved by adding doc comments for our private methods like we do for public methods, so maybe we should consider that as well.

src/parsec/agent/runners/evm/impl.cpp
AlexRamRam: My bias is towards making the code itself readable and descriptive - with minimal comments if possible. Not sure if this is successful, but the hope is that when the reader looks at the block of code, the attention is drawn to this line which follows the main flow of execution.

PR #241, opened 2023-12-03
Guide to implementing a new runner + Python runner for PArSEC

maurermi:
This is a Python-based runner and a corresponding guide for adding a new runner. The goal of this extension is to make development of new runner environments more straightforward. 

Also embedded here are a script for running the Lua bench test, a ticket logging mechanism for reporting the status of tickets seen by a broker, and a get_row utility for reading directly from the shards. These could be split out into a separate PR if helpful.

Note (for reviewers): This introduces a dependency on Python developer tools. I think it's worth considering making this a separate branch or repository as a result.


maurermi:
Will address the linting errors shortly

src/parsec/agent/runners/py/impl.cpp
```diff
@@ -0,0 +1,398 @@
...
+            arg_string.erase(0, pos + 1);
+        }
+        functionString.erase(0, arg_delim + 1);
+
+        m_function = cbdc::buffer();
+        m_function.append(functionString.c_str(), functionString.size());
+        m_function.append("\0", 1);
+    }
+
+    auto py_runner::parse_params() -> std::vector<std::string> {
```
maurermi, line 155: Realized that the pointer arithmetic here is bad practice -- fixing this ASAP

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
+def pay(pay, balance1, balance2):
```
rockett-m, line 1: Type hints won't forcibly typecast or error out but helps for reading the code
```suggestion
def pay(pay: int, balance1: int, balance2: int) -> [int, int]:
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+    balance1 = int(balance1)
+    balance2 = int(balance2)
+    pay = int(pay)
+    if(pay > balance1):
+        pay = 0
+    balance1 = balance1 - pay
```
rockett-m, line 7:
```suggestion
    balance1 -= pay
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+    balance2 = balance2 + pay
```
rockett-m, line 8:
```suggestion
    balance2 += pay
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
+def pay(pay, balance1, balance2):
+    balance1 = int(balance1)
+    balance2 = int(balance2)
+    pay = int(pay)
```
rockett-m, line 4: [optional] could compact it by assigning on one line. Just style preference
```suggestion
    balance1, balance2, pay = int(balance1), int(balance2), int(pay)
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+    if(pay > balance1):
```
rockett-m, line 5: spacing after the if is preferable for readability
```suggestion
    if (pay > balance1):
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+def accrueInterest(rate, balance1):
+    balance1 = int(balance1)
+    rate = float(rate)
+    print("Balance 1:", balance1)
+    print("Rate:", rate*100, "%")
+    balance1 = int(balance1 * (2.71828**rate))
+    print("Balance 1:", balance1)
+    return [balance1]
+
+def pay2(pay1, pay2, balance0, balance1, balance2):
```
rockett-m, line 20:
```suggestion
def pay2(pay1, pay2, balance0, balance1, balance2) -> list[int, int, int]:

    balance0, balance1, balance2 = int(balance0), int(balance1), int(balance2)

    if (pay1 + pay2 > balance0) and (min(pay1, pay2) > 0):
        pay1, pay2 = 0, 0

    balance0 -= (pay1 + pay2)
    balance1 += pay1
    balance2 += pay2

    return [balance0, balance1, balance2]
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+    return [balance1, balance2]
+
+def accrueInterest(rate, balance1):
+    balance1 = int(balance1)
+    rate = float(rate)
+    print("Balance 1:", balance1)
+    print("Rate:", rate*100, "%")
+    balance1 = int(balance1 * (2.71828**rate))
+    print("Balance 1:", balance1)
+    return [balance1]
```
rockett-m, line 18: Would write it like this. Could add a "Balance 1 after rate:" print after the conversion has been applied.
```suggestion
def accrueInterest(rate: float, balance1: int) -> list[int]:

    rate, balance1 = float(rate), int(balance1)

    print(f"Balance 1: {balance1}")
    print(f"Rate: {rate*100}%")

    balance1 = int(balance1 * (2.71828**rate))
    print(f"Balance 1: {balance1}")

    return [balance1]
```

scripts/paycontract.py
```diff
@@ -0,0 +1,33 @@
...
+    balance2 = int(balance2)
+    pay1 = int(pay1)
+    pay2 = int(pay2)
+    if(pay1 + pay2 > balance0 and pay1 > 0 and pay2 > 0):
+        pay1 = 0
+        pay2 = 0
+    balance0 = balance0 - pay1 - pay2
+    balance1 = balance1 + pay1
+    balance2 = balance2 + pay2
+    return [balance0, balance1, balance2]
```
rockett-m, line 32:
```suggestion
def pay2(pay1, pay2, balance0, balance1, balance2) -> list[int, int, int]:

    balance0, balance1, balance2 = int(balance0), int(balance1), int(balance2)

    if (pay1 + pay2 > balance0) and (min(pay1, pay2) > 0):
        pay1, pay2 = 0, 0

    balance0 -= (pay1 + pay2)
    balance1 += pay1
    balance2 += pay2

    return [balance0, balance1, balance2]
```

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
+# assumes all return types are strings
+def parseContract(file, funcName):
+    import re
```
rockett-m, line 3: Best to put imports at the top before functions

scripts/lua_bench.sh
```diff
@@ -0,0 +1,14 @@
...
+IP="localhost"
+PORT="8889"
+N_WALLETS=2
+
+./build/tools/bench/parsec/lua/lua_bench --component_id=0 \
+    --ticket_machine0_endpoint=$IP:7777 --ticket_machine_count=1 \
+    --shard_count=1 --shard0_count=1 --shard00_endpoint=$IP:5556 \
+    --agent_count=1 --agent0_endpoint=$IP:$PORT \
+    --loglevel=TRACE scripts/gen_bytecode.lua $N_WALLETS
+echo done
```
rockett-m, line 14: Looks good to me

scripts/parsec-run-local.sh
```diff
@@ -24,8 +24,13 @@ for arg in "$@"; do
...
             RUNNER_TYPE="lua"
-        elif [[ "$arg" != "--runner_type=evm" ]]; then
-            echo "unknown runner type, using evm"
+        elif [[ "$arg" == "--runner_type=py" ]]; then
+            RUNNER_TYPE="py"
+        elif [[ "$arg" == "--runner_type=evm" ]]; then
+            RUNNER_TYPE="evm"
+        else
+            echo "unknown runner type, using "
+            echo $RUNNER_TYPE
```
rockett-m, line 33: Makes sense

scripts/configure.sh
```diff
@@ -41,7 +41,7 @@ fi
 
 if [[ "$OSTYPE" == "linux-gnu"* ]]; then
   apt update
-  apt install -y build-essential wget cmake libgtest-dev libbenchmark-dev lcov git software-properties-common rsync unzip
+  apt install -y build-essential wget cmake libgtest-dev libbenchmark-dev lcov git software-properties-common rsync unzip python3-dev
```
rockett-m, line 44: Verified that this works. It installed python 3.10 in my Ubuntu environment, but in `Dockerfile` at root dir, it is specified that we use python 3.8 `ENV PY_VER=python3.8` (line 30)

Dockerfile
```diff
@@ -27,6 +27,8 @@ FROM $BASE_IMAGE AS builder
 # Copy source
 COPY . .
 
+ENV PY_VER=python3.8
```
rockett-m, line 30: Why 3.8 instead of newer such as 3.10+ ?

Dockerfile
```diff
@@ -90,6 +92,11 @@ COPY --from=builder /opt/tx-processor/build/src/parsec/ticket_machine/ticket_mac
...
 COPY --from=builder /opt/tx-processor/build/tools/bench/parsec/evm/evm_bench ./build/tools/bench/parsec/evm/evm_bench
 COPY --from=builder /opt/tx-processor/build/tools/bench/parsec/lua/lua_bench ./build/tools/bench/parsec/lua/lua_bench
+COPY --from=builder /opt/tx-processor/build/tools/bench/parsec/py/py_bench ./build/tools/bench/parsec/py/py_bench
 
 # Copy wait script
 COPY --from=builder /opt/tx-processor/scripts/wait-for-it.sh ./scripts/wait-for-it.sh
+
+# Copy python scripts
+COPY --from=builder /opt/tx-processor/scripts/pythonContractConverter.py ./scripts/pythonContractConverter.py
+COPY --from=builder /opt/tx-processor/scripts/paycontract.py ./scripts/paycontract.py
```
rockett-m, line 102: lgtm

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
...
+
+    contract = d_types + '|'
+
+    for elem in data_lines:
+        contract += elem
+
+    return contract
+
+
+contract = parseContract(file, funcname)
```
rockett-m, line 114: "file" and funcname" are not defined. Is this meant to be a comment on how to run it?

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
+# assumes all return types are strings
+def parseContract(file, funcName):
```
rockett-m, line 2: Would add type hints then
```suggestion
# assumes all return types are strings
def parseContract(file: str, funcName: str) -> str:
```

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
...
+
+    idx = 0
+    for i in range(len(data_lines)):
+        if("def " + funcName + "(" in data_lines[i]):
+            idx = i
+            break
+    data_lines = data_lines[idx:]
+
+
+    while(True):
```
rockett-m, line 32: no need for parentheses

```suggestion
    while True:
```

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
...
+
+    fin = -1
+    try:
+        fin = open(file)
+    except:
+        print("[ERROR] {fname} cannot be opened".format(fname = file))
+        return ""
+
+    data = fin.read()
+    fin.close()
```
rockett-m, line 13: The with handles file closing automatically
```suggestion
    data = ''
    try:
        with open(file, 'r') as fi:
            data = fi.read()
    except:
        print(f"[ERROR] {file} cannot be opened")
        return ""
```

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
...
+    pattern = re.compile(r'#.*?\n')
+    data = re.sub(pattern, '', data)
+    pattern = re.compile(r"'''.*?'''", re.DOTALL)
+    data = re.sub(pattern, '', data)
+    if(("def " + funcName + "(") not in data):
+        print("[ERROR] {fname} does not contain a definition for \"{func}\"".format(fname = file, func = funcName))
+        return ""
```
rockett-m, line 20: f-strings are useful here
```suggestion
    if f"def {funcName}(" not in data:
        print(f"[ERROR] {file} does not contain a definition for \"{funcName}\"")
        return ""
```

scripts/pythonContractConverter.py
```diff
@@ -0,0 +1,115 @@
...
+        if(i == '\t'):
+            tabChar = True
+            tabLength = 1
+            break
+        elif(i == ' '):
+            tabLength += 1
+        else:
+            break
+
+    startChar = '\t' if tabChar else ' ' * tabLength
```
rockett-m, line 50: clean way of initializing

HalosGhost:
Needs rebase.

Dockerfile
HalosGhost: Since #252 was merged, we require python 3.10 in-practice anyway (in terms of what's shipped by the image our docker images are based upon). I'm quite comfortable pushing this to 3.10.

Dockerfile
HalosGhost: Reasonable additions, but this is already a strong argument that we need a better maintainability story for runners and how we add them.

(i.e., the `scripts` directory is going to rapidly become unwieldy—let's be honest, it already is—should we continue adding runners without bound).

scripts/lua_bench.sh
HalosGhost: needs rebasing for #253

scripts/py_bench.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# A sample script for running the Python Bench test
+
+IP="127.0.0.1"
+PORT="8889"
+
+#cp ./scripts/pythonContractConverter.py ./build/tools/bench/parsec/py/
+
+rr record ./build/tools/bench/parsec/py/py_bench --component_id=0 --ticket_machine0_endpoint=$IP:7777 --ticket_machine_count=1 --shard_count=1 --shard0_count=1 --shard00_endpoint=$IP:5556 --agent_count=1 --agent0_endpoint=$IP:$PORT
```
HalosGhost, line 10: I _love_ [`rr`](https://rr-project.org/), but we don't currently require it or set it up in our environment-setup scripts. `./scripts/native-system-benchmark.sh` as added in #259 might offer some inspiration for how to optionally support it.

scripts/pythonContractConverter.py
HalosGhost: I fully agree that type-hints are a good practice, and make the commented assumption at least parsed (even if not enforced).

I might additionally suggest removing the comment if the type hints are added.

src/parsec/agent/agentd.cpp
```diff
@@ -142,6 +143,16 @@ auto main(int argc, char** argv) -> int {
             broker,
             log,
             cfg.value());
+    } else if(cfg->m_runner_type == cbdc::parsec::runner_type::py) {
```
HalosGhost, line 146: (optional) I think this full-block might be a candidate for replacement with `switch`/`case` (so that the number of supported runners doesn't necessarily change the start-up time).

~However, there's a deeper question that may be worth considering here: are runners a good candidate for pulling into a dynamically-linked library enabling the installation/support of new runners without recompilation? This is not a very small engineering effort; but something like [dylib](https://github.com/martin-olivier/dylib) might enable us to load the runners at runtime.~

~This might be a particularly positive way of pulling runners out of the main repo and modularizing them. Food for thought.~

cf. #261

src/parsec/agent/runners/py/impl.cpp
```diff
@@ -0,0 +1,413 @@
...
+        }
+        if(bail) {
+            return params;
+        }
+        // ---
+
+        // get parameters that are user inputs
+        size_t bufferOffset = 0;
+        auto it = 0;
+        // Known that there are 3 parts of the header
```
HalosGhost, line 194: This feels a touch magical. Is this guaranteed, python-version-dependent, or internally-specified?

If internally-specified, is there benefit to pulling these cases out as an enum?

HalosGhost (commented):
Some initial comments as I skimmed through pre-rebase and testing.

Solid conceptACK; look forward to testing thoroughly and getting it merged!

scripts/py_bench.sh
maurermi: This is an oversight on my end, should not be included here -- good catch

src/parsec/agent/agentd.cpp
maurermi: WRT the first comment, I would agree that we ought to make this a switch statement -- for this, I think what'd be best is defining an enum with runner types 

WRT the second comment, I think that's actually a very interesting idea -- and i am definitely for pulling runners out of the main repository. This may be a task for a later PR but I think this is good to keep in mind for sure

src/parsec/agent/runners/py/impl.cpp
maurermi: This is internally specified, and the input validation is handled just above this code block -- just to check, are you referring to the cases in the switch statement below this block?

scripts/pythonContractConverter.py
maurermi: So this file is actually just used to parse python functions such that arbitrary python functions can be formatted for usage as smart contracts. It cannot be run on its own, but is written to be used within `pyutil.cpp:formContract`. 

In that function, `file` and `funcname` are specified within the C++ scope and registered within the Python context, which is why they are not defined here.

I do certainly understand the confusion here, and I don't actually think this is the cleanest way to do this contract parsing -- though it is much more straightforward to define string parsing behavior within python as opposed to C++ which is why I did it. This was just written to leverage the already-in-use Python VM to handle some string operations with the ease of Python string operations.

scripts/pythonContractConverter.py
maurermi: This actually refers to the return types of functions which are parsed by `parseContract`, so this behavior cannot be represented as a type hint (or at least not a standard Pythonic type hint). That said, these type hints are correct and could be added without an issue in this context, if helpful.

scripts/pythonContractConverter.py
maurermi: The reason I included the import here instead of above is because python functions which are to be converted to smart contracts need to have their imports defined within the actual function, so this preserves that stylistic behavior. In this particular instance however the import can be defined outside of the function, so I would be happy to make that change. 

Further, imports need to be specified in the C++ scope (see `runners/py/pyutil.cpp:20`) which is interesting to note, because it allows for the system designer to restrict what python libraries are available for smart contracts.

Dockerfile
maurermi: Agreed, there is no reason that this needs to be 3.8 and cannot be 3.10 -- at the time of writing I was using python 3.8 but can run it just fine with python 3.10

src/parsec/agent/runners/py/impl.cpp
HalosGhost: Yeah, the actual line I highlighted for this comment was line 194 (the comment asserting that there are three parts of the header). 👍

scripts/pythonContractConverter.py
rockett-m: Ok, good context, thanks for explaining the logic

PR #244, opened 2024-01-08
Add CLI arguments to build.sh, specify build type

maurermi:
Adds CLI arguments to build.sh, and prints to stdout the type of build being performed. Default build is still Debug. Additonally supports a BUILD_DEBUG environment variable, and respects previously set values of CMAKE_BUILD_TYPE. 

Low urgency - I found this useful and hopefully others will.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,35 @@
 #!/bin/bash
+
 set -e
+
+help() {
+    if [ $# -gt 0 ]; then
+        printf 'Unexpected Argument (%s)\n' "$1"
+    fi
+    printf 'HELP: Usage: %s [Debug|Release|Profiling]\n' "$0"
+    exit 0
+}
+
+# Note:
+# CMAKE_BUILD_TYPE="Debug" adds "-O0 -g" flags by default
+# CMAKE_BUILD_TYPE="Release" adds "-O3 -DNDEBUG" by default
+if [[ "$BUILD_DEBUG" == "1" ]]; then
+    CMAKE_BUILD_TYPE="Debug"
+elif [[ "$BUILD_RELEASE" == "1" ]]; then
+    CMAKE_BUILD_TYPE="Release"
+elif [[ "$BUILD_PROFILING" == "1" ]]; then
+    CMAKE_BUILD_TYPE="Profiling"
+fi
+
+if [ $# -gt 0 ]; then
+    case "$1" in
+        Release|--release|-r) CMAKE_BUILD_TYPE="Release";;
+        Profiling|--profiling|-p) CMAKE_BUILD_TYPE="Profiling";;
+        Debug|--debug|-d) CMAKE_BUILD_TYPE="Debug";;
+        --help|-h) help;;
+        *) help $1;;
+    esac
+fi
 
 echo "Building..."
 
@@ -23,12 +53,12 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
     CMAKE_FLAGS+=" -DCMAKE_C_COMPILER=${XCODE_CMDLINE_DIR}/usr/bin/clang -DCMAKE_CXX_COMPILER=${XCODE_CMDLINE_DIR}/usr/bin/clang++ -DCMAKE_CXX_FLAGS=-isystem\ /usr/local/include -DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
 fi
 
-CMAKE_BUILD_TYPE="Debug"
-if [[ "$BUILD_RELEASE" == "1" ]]; then
-    CMAKE_BUILD_TYPE="Release"
-elif [[ "$BUILD_PROFILING" == "1" ]]; then
-    CMAKE_BUILD_TYPE="Profiling"
+if [[ -z $CMAKE_BUILD_TYPE ]]; then
+    echo "CMAKE_BUILD_TYPE not set, defaulting to debug"
+    CMAKE_BUILD_TYPE="Debug"
 fi
 
+echo "Building $CMAKE_BUILD_TYPE"
 eval "cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} ${CMAKE_FLAGS} .."
 make -j$CPUS
+


anders94 (approved):
LGTM

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
...
+    CMAKE_BUILD_TYPE="Debug"
+elif [[ "$BUILD_RELEASE" == "1" ]]; then
     CMAKE_BUILD_TYPE="Release"
 elif [[ "$BUILD_PROFILING" == "1" ]]; then
     CMAKE_BUILD_TYPE="Profiling"
 fi
 
+if [ $# -gt 0 ]; then
+    if [ "$1" == "Release" ]; then
+        CMAKE_BUILD_TYPE="Release"
```
rockett-m, line 36: [optional] Thinking it could be helpful to have a 'help' arg for users using the cli and/or reading the script. This would show the options of Release|Profiling|Debug for `build.sh` if and only if the user enters 'help' or '--help' exactly.
```suggestion
    if [[ "$1" == "help" || "$1" == "--help" ]]; then
```

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
...
+if [[ "$BUILD_DEBUG" == "1" ]]; then
+    CMAKE_BUILD_TYPE="Debug"
+elif [[ "$BUILD_RELEASE" == "1" ]]; then
     CMAKE_BUILD_TYPE="Release"
 elif [[ "$BUILD_PROFILING" == "1" ]]; then
     CMAKE_BUILD_TYPE="Profiling"
 fi
 
+if [ $# -gt 0 ]; then
+    if [ "$1" == "Release" ]; then
```
rockett-m: [optional] Thinking it could be helpful to have a 'help' arg for users using the cli and/or reading the script. This would show the options of Release|Profiling|Debug for `build.sh` if and only if the user enters 'help' or '--help' exactly.
```suggestion
    if [[ "$1" == "help" || "$1" == "--help" ]]; then
        echo "Usage: ./build.sh [Release|Profiling|Debug]"
        echo "If no argument is provided, the default build type is Debug."
        exit 0
    elif ...
```
Then would keep the existing 3 cases for Release, Profiling, Debug

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
...
+fi
+
+if [[ -z $CMAKE_BUILD_TYPE ]]; then
+    echo "CMAKE_BUILD_TYPE not set, defaulting to debug"
+    CMAKE_BUILD_TYPE="Debug"
+fi
+
+echo "Building $CMAKE_BUILD_TYPE"
 eval "cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} ${CMAKE_FLAGS} .."
 make -j$CPUS
```
rockett-m, line 63: [optional] We could print out error messages for debug if the cmake and make commands do not exit 0 as expected. 
This block could be put right after the `eval "cmake ...` line.
An identical copy of it could be after the `make ...` command except changing 'cmake' to 'make'.
```suggestion
if [[ $? -ne 0 ]]; then
    echo "cmake command failed"
    exit 1
fi
```

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
...
+    if [ "$1" == "Release" ]; then
+        CMAKE_BUILD_TYPE="Release"
+    elif [ "$1" == "Profiling" ]; then
+        CMAKE_BUILD_TYPE="Profiling"
+    elif [ "$1" == "Debug" ]; then
+        CMAKE_BUILD_TYPE="Debug"
+    fi
+fi
+
+if [[ -z $CMAKE_BUILD_TYPE ]]; then
```
rockett-m, line 56: Debug for a default arg makes sense; looks good

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
     CMAKE_FLAGS+="-DCMAKE_C_COMPILER=${XCODE_CMDLINE_DIR}/usr/bin/clang -DCMAKE_CXX_COMPILER=${XCODE_CMDLINE_DIR}/usr/bin/clang++ -DCMAKE_CXX_FLAGS=-isystem\ /usr/local/include -DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
 fi
 
-CMAKE_BUILD_TYPE="Debug"
-if [[ "$BUILD_RELEASE" == "1" ]]; then
+# Note:
```
rockett-m: Helpful note

scripts/build.sh
```diff
@@ -20,12 +20,33 @@ elif [[ "$OSTYPE" == "darwin"* ]]; then
...
+        CMAKE_BUILD_TYPE="Debug"
+    fi
+fi
+
+if [[ -z $CMAKE_BUILD_TYPE ]]; then
+    echo "CMAKE_BUILD_TYPE not set, defaulting to debug"
+    CMAKE_BUILD_TYPE="Debug"
+fi
+
+echo "Building $CMAKE_BUILD_TYPE"
```
rockett-m, line 61: Good move telling the user which build type is being chosen

scripts/build.sh
maurermi: This is an interesting point, I hadn't considered adding additional help statements here, as the cmake and make steps will give significant feedback as to why the compilation failed. Not sure that we necessarily need this here but can be persuaded.

scripts/build.sh
maurermi: This is a good point, now that the script takes arguments, it makes sense to have a help flag. Will add this in

maurermi: Question for you here, do you think it makes sense to pass in arguments with dashes? I may switch to `--release` from `Release` to stick with our typical convention

HalosGhost: As a general rule, if there's going to be more than one option available to pass to a script, help output is a kindness. To that end, if the options aren't positional (i.e., you can pass more than one and in any order), they really should follow the `-l`/`--long` pattern.

scripts/build.sh
HalosGhost: The suggested change seems to remove the actual build with `make` which is probably a mistake, but should not be committed.

Beyond that, testing to see if `cmake` failed rather than blindly running `make` probably isn't a bad idea.

scripts/build.sh
```diff
@@ -1,6 +1,33 @@
...
+    CMAKE_BUILD_TYPE="Profiling"
+fi
+
+if [ $# -gt 0 ]; then
+    if [ "$1" == "Release" ] || [ "$1" == "--release" ]; then
+        CMAKE_BUILD_TYPE="Release"
+    elif [ "$1" == "Profiling" ] || [ "$1" == "--profiling" ]; then
+        CMAKE_BUILD_TYPE="Profiling"
+    elif [ "$1" == "Debug" ] || [ "$1" == "--debug" ]; then
+        CMAKE_BUILD_TYPE="Debug"
```
HalosGhost: Note that `[` is a synonym for the `test` program (frequently a shell built-in). POSIX does not specify that `test` accept a `==` operator; the string equality operator is `=`.

Since this is a bash script, we should probably prefer `[[` (which has nicer ergonomics anyway) which *does* include a `==` operator, but if we are going to use `[`, we should use `=` for string comparisons.

However, I might suggest using a case statement here. The above lines might look instead like so:

```bash
case "$1" in
    Release|--release) CMAKE_BUILD_TYPE='Release';;
    Profiling|--profiling) CMAKE_BUILD_TYPE='Profiling';;
    Debug|--debug) CMAKE_BUILD_TYPE='Debug';;
    *) printf 'Unrecognized build type…\nUsage: %s [Debug|Release|Profiling]\n' "$0";;
esac
```

maurermi: @HalosGhost ACK; will update

maurermi:
@HalosGhost @rockett-m Made some updates to address review comments, would love another look over if possible!

scripts/build.sh
maurermi: We set this script to exit if any command fails, so we could print out this info but it would require not using `set -e`

HalosGhost (approved):
tACK. Does what it says on the tin; allow specifying build-type as a basic command-line argument.

Purely positional, but still much more pleasant than having to use environment variables; and interacts cleanly with env-vars so no tooling should break (cleanly allowing cli args to override env-vars as-typical).

LGTM. Merging!

PR #246, opened 2024-01-13
Parsec atomic commit fix

maurermi:
Fixes a bug which occurs when a smart contract does not request locks on all update keys, resulting in undesirable behavior. 

Aborts a transaction where the agent does not have tickets associated with all update keys. In the case that a contract does not prompt for all update keys to be locked, enforces that the transaction aborts (+ rolls back) and no updates are written, rather than those which are locked. Does not abort if the smart contract correctly requested locks and is still waiting on them, but still rolls back and retries.

### DIFF
--- a/src/parsec/util.cpp
+++ b/src/parsec/util.cpp
@@ -5,6 +5,7 @@
 
 #include "util.hpp"
 
+#include <future>
 #include <unordered_map>
 
 namespace cbdc::parsec {
@@ -202,6 +203,7 @@ namespace cbdc::parsec {
                 result_callback(false);
                 return;
             }
+
             auto ticket_number
                 = std::get<cbdc::parsec::ticket_machine::ticket_number_type>(
                     begin_ret);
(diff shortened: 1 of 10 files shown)


maurermi:
Thanks to @ayeshaali and @madars for identifying this

maurermi:
This PR also contains a new util called `get_row` which is also included in #241 to enable the integration tests included with this PR. If this is not wanted, we can keep only the commit which fixes this behavior.

src/parsec/util.cpp
```diff
@@ -183,6 +184,8 @@ namespace cbdc::parsec {
                 cfg.m_runner_type = runner_type::evm;
             } else if(val == "lua") {
                 cfg.m_runner_type = runner_type::lua;
+            } else if(val == "py") {
```
anders94: While I love this, I don't believe this should be in here - part of the other PR you are on AFAIK

src/parsec/util.hpp
```diff
@@ -25,6 +25,8 @@ namespace cbdc::parsec {
         lua,
         /// Ethereum-style transactions using EVM.
         evm,
+        /// Python Script using CPython
+        py,
```
anders94: Same issue here

src/parsec/util.hpp
```diff
@@ -71,6 +73,20 @@ namespace cbdc::parsec {
                  broker::key_type key,
                  broker::value_type value,
                  const std::function<void(bool)>& result_callback) -> bool;
+
+    /// Asynchronously get the value stored at key from the cluster.
+    /// Intended for testing purposes.
```
anders94: Nit-picky but potentially administrative functions as well like migrating keys? I wouldn't use that example but maybe the comment should say "testing and administration purposes" instead?

Yes, I did just add a review comment on a code comment... :)

anders94 (approved):
LGTM - nit-picks only.

src/parsec/util.cpp
maurermi: Good catch, thanks!

src/parsec/util.hpp
maurermi: I can get on board with that, will update

maurermi: For context, this comment is trying to convey that this is not the optimal method for reading from the shards as a system component, it is better to handle shard responses in the agent

HalosGhost:
utACK, clear and simple fix (core fix in 1e51249) to throw a reasonable error in the agent if a change-set includes modifications to keys which don't have tickets.

I also want to call out how easy it is to review this PR because of the pleasant separation of the commits. CI passed, and the [output](https://github.com/mit-dci/opencbdc-tx/actions/runs/7513646742/job/20455629118?pr=246#step:5:4120) for the added integration tests looks good:

```
Running integration tests...
Running main() from /build/googletest-j5yxiC/googletest-1.10.0/googletest/src/gtest_main.cc
[==========] Running 23 tests from 11 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from parsec_smart_contract_updates_test
[ RUN      ] parsec_smart_contract_updates_test.valid_updates
[       OK ] parsec_smart_contract_updates_test.valid_updates (1038 ms)
[ RUN      ] parsec_smart_contract_updates_test.invalid_updates
Error: 1-13 16:47:58.584] [ERROR] Update map contains keys not associated with tickets. Aborting.
Error: 1-13 16:47:58.585] [ERROR] Broker error for commit for 14
[2024-01-13 16:47:58.585] [WARN ] Exec failed
[       OK ] parsec_smart_contract_updates_test.invalid_updates (1036 ms)
[----------] 2 tests from parsec_smart_contract_updates_test (2074 ms total)
```

HalosGhost (approved):
tACK. Does what it says on the tin! Merging.

PR #247, opened 2024-01-18
Update docker workflows to support parsec

maurermi:
Adds the parsec architecture to the docker build workflows. 

Added this to the current workflow implementations, but was considering making parsec building a separate workflow entirely -- would welcome thoughts on this


HalosGhost (approved):
tACK.

Adds workflows for parsec to be built automatically in CI/CD and brought in-line with Atomizer/2PC. I walked through the build and produced image on @maurermi's fork, and it's looking good to me.

Merging.

PR #248, opened 2024-01-22
Add default platform for Docker builds

maurermi:
Addresses an issue seen with M-series Macs where Docker looks to pull images built for the linux/arm64/v8 platform instead of the linux/amd64 platform by default.

Specifies platform: linux/amd64 in docker-compose files to fix this issue for Mac users.

Minor additional change: also renames the image built / targeted by docker-compose-parsec-test.yml to opencbdc-tx-parsec instead of opencbdc-tx to match convention.

### DIFF
--- a/docker-compose-2pc.yml
+++ b/docker-compose-2pc.yml
@@ -7,6 +7,7 @@ services:
       context: .
       target: twophase
     image: opencbdc-tx-twophase
+    platform: linux/amd64
     tty: true
     restart: always
     command: ./build/src/uhs/twophase/sentinel_2pc/sentineld-2pc 2pc-compose.cfg 0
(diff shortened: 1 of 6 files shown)


maurermi:
Tested these changes on an M1 Mac and on my personal non-Mac system. 

Would be great to get approval from another M-series Mac user and a non-Mac user if possible.

HalosGhost:
utACK. Specifies amd64 as the platform for all docker images so that the right image is grabbed by-default across any platform.

Looks like we have a [tACK from Hung Nguyen](https://opencbdc.zulipchat.com/#narrow/stream/315624-General/topic/An.20error.20with.20.60docker.20compose.60.20and.20parsec/near/417553778) in the community. I don't have Apple Silicon to test on so I cannot give a tACK, but @maurermi if Hung's use serves as confirmation for you, I'm happy to approve and merge.

maurermi:
@HalosGhost That sounds good to me!

HalosGhost (approved):
Sounds good. Merging!

PR #252, opened 2024-03-08
fix: update docker base image to ubuntu 22.04

HalosGhost:
Having migrated to C++20, we ran into a version of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100346

By migrating to a more recent Ubuntu image, we benefit from more recent compiler versions which avoid the build errors.

The additional library linkages for the benchmarks seem to have become required on some platforms in addition.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Define image base arg
-ARG IMAGE_VERSION="ubuntu:20.04"
+ARG IMAGE_VERSION="ubuntu:22.04"
 
 # Define base image repo name
 ARG BASE_IMAGE="ghcr.io/mit-dci/opencbdc-tx-base:latest"
--- a/scripts/Dockerfile
+++ b/scripts/Dockerfile
@@ -1,8 +1,9 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
 COPY . .
 
 RUN apt update && apt dist-upgrade -y
-RUN ./configure.sh
+RUN ./install-build-tools.sh
+RUN ./setup-dependencies.sh
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -19,6 +19,8 @@ target_link_libraries(run_benchmarks ${GTEST_LIBRARY}
                                      watchtower
                                      locking_shard
                                      transaction
+                                     rpc
+                                     network
                                      common
                                      serialization
                                      crypto


scripts/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 ENV DEBIAN_FRONTEND noninteractive
 
 COPY . .
 
 RUN apt update && apt dist-upgrade -y
-RUN ./configure.sh
```
rockett-m, line 8: The configure file was removed in a recent commit; therefore we no longer need to reference it.  
I tested this change out and found that the build passed with this line removed.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Define image base arg
-ARG IMAGE_VERSION="ubuntu:20.04"
+ARG IMAGE_VERSION="ubuntu:22.04"
```
rockett-m, line 2: Updating the Ubuntu version enabled the build to no longer fail when running `scripts/build.sh`.  
Previously, the C++ library build was failing. I replicated @HalosGhost's changes to the Dockerfiles and CMakeLists file in a fresh Ubuntu 22.04 container with forked clone of the repo.  
After installing the necessary build tools/packages, I was able to call the setup dependencies script and run the build successfully.

scripts/Dockerfile
```diff
@@ -1,8 +1,9 @@
...
+RUN ./install-build-tools.sh
```
rockett-m, line 8: Before running a build by `./scripts/build.sh`, I manually ran `./install-build-tools.sh` and `./setup-dependencies.sh` in the command line. 
Adding these bash script calls in the Dockerfile saves time.

scripts/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
```
maurermi, line 1: `scripts/Dockerfile` does not seem to do anything, we can probably remove it. Was able to build docker containers and pass all tests without it.

maurermi (approved):
T-ACK. These changes fix the previously (recently) observed error in building using Docker. Tested by running `scripts/build-docker.sh` and running the test suite in Docker. Looks good to merge.

One note is that `scripts/Dockerfile` seems like it is not necessary to keep around. I am not sure what it does and had no issue building the docker containers without it.

scripts/Dockerfile
rockett-m: Agreed. I ran a successful build with this file deleted.

scripts/Dockerfile
HalosGhost: I did a bit of spelunking through some of the old history of the repo (before release), and located where this Dockerfile came from. As best I can recover, it served as a method of getting a development environment up with the current source tree imported to make it easy to have a reproducible/consistent setup.

While not strictly necessary for us to keep, I'm not yet certain if some of our ancillary tooling (namely, the Test Controller) depends on it. For now, I'm opting to keep it in and updated so that the dev use-case remains practical; but I would not oppose its removal once we've managed to ensure nothing else would break.

HalosGhost:
cf. https://github.com/mit-dci/opencbdc-tx/pull/252#discussion_r1520012001 for my reasoning for leaving this in in this PR; once we've finished vetting that it won't break any other tooling we have (most obviously the TC), I'd be happy to accept a PR to remove it.

PR #253, opened 2024-03-23
Usability updates for lua bench testing

maurermi:
Add in script for running lua bench, and add
configurable logging to the lua bench test.

Also updates .gitignore to ignore installed
lua files but not files in a 'lua/' subdirectory.

This lua bench script is a more fleshed out version
of the one added in PR #241

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,7 @@ NuRaft*/
 curl*/
 jsoncpp*/
 ethash*/
-lua*/
+lua-*/
 benchmark-results/
 CMakeFiles/
 plots/
--- a/tools/bench/parsec/lua/lua_bench.cpp
+++ b/tools/bench/parsec/lua/lua_bench.cpp
@@ -32,6 +32,7 @@ auto main(int argc, char** argv) -> int {
         log->error("Error parsing options");
         return 1;
     }
+    log->set_loglevel(cfg->m_loglevel);
 
     auto args = cbdc::config::get_args(argc, argv);
     auto n_wallets = std::stoull(args.back());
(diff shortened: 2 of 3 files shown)


tools/bench/parsec/lua/lua_bench.cpp
```diff
@@ -32,6 +32,7 @@ auto main(int argc, char** argv) -> int {
         log->error("Error parsing options");
         return 1;
     }
+    log->set_loglevel(cfg->m_loglevel);
```
rockett-m, line 35: Looks good

.gitignore
```diff
@@ -43,7 +43,7 @@ NuRaft*/
 curl*/
 jsoncpp*/
 ethash*/
-lua*/
+lua-*/
```
rockett-m, line 46: Less greedy regex is preferable as you entered

HalosGhost:
utACK; looks like a straightforward addition of a wrapper benchmarking script for the lua runner along with a touch more logging and an update to the ignore file. Do you have a suggested procedure for a tACK?

maurermi:
@HalosGhost This is indeed a very small change, to test I would suggest is to run parsec locally, and then run the Lua bench script (potentially as follows):

```
(configure, build, etc...)
./scripts/parsec-run-local.sh --runner_type=lua --port=8889
(wait for spawn ...)
./scripts/lua_bench.sh --loglevel=TRACE
```
I assert that this _should_ work and run the Lua bench test locally with trace logging

HalosGhost (approved):
tACK. Builds, all tests pass, and the procedure above successfully runs a lua-runner benchmark, producing a tx-sample output.

Looks like a really pleasant QoL improvement to me. Merging!

PR #255, opened 2024-04-10
Publish github pages to external repository

maurermi:
This PR addresses the growing gh-pages branch and re-writes the github pages action to direct pages to an external repository. 
Rather than cloning a single branch to avoid pulling down gh-pages which can be annoying to work with when the repo has multiple branches, having gh-pages in a separate repository may be more useful. 

This also makes gh-pages opt-in instead of opt-out.

PRing this for consideration and discussion, as making this change would require slightly more than merging upstream. 

cc: @madars

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -30,4 +30,5 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          external_repository: mit-dci/opencbdc-tx-pages
           publish_dir: ./doxygen_generated/html
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -27,8 +27,8 @@ jobs:
           enable-latex: false
           additional-packages: graphviz
       - name: Deploy GitHub Pages
-        uses: peaceiris/actions-gh-pages@v3
+        uses: peaceiris/actions-gh-pages@v4
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          deploy_key: ${{ secrets.TX_PAGES_DEPLOY_KEY }}
           external_repository: mit-dci/opencbdc-tx-pages
           publish_dir: ./doxygen_generated/html


HalosGhost:
Immense Concept-ACK. What steps do we need to take to tACK / deploy? Does the repo need to be created in-advance and are any permissions changes necessary?

maurermi:
This is a good question, I believe the repo does need to be created in advance, and it looks like the `github_token` affiliated with the action in this repo would not work for pushing to an external repo. It looks like the way to go about doing this would be to create a SSH deploy key, and give the GH action permission to write to the newly created 'pages' repo.

(Potentially Useful Links: )
https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-deploy-to-external-repository-external_repository 
https://docs.github.com/en/authentication/connecting-to-github-with-ssh/managing-deploy-keys
https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-create-ssh-deploy-key

HalosGhost:
Okay, I've just tested this using a couple of basic repos to make sure I understand the flow. Here's a checklist for us to implement:

- [X] Create [documentation repository](https://github.com/mit-dci/opencbdc-tx-pages)
- [X] Generate deployment key
- [X] Add deployment secret key to -tx, and deployment public key to docs repo
- [x] Update workflow to target new repo
   - ~~to potentially make it easier to host docs for multiple repos later on, I think we should target a branch specific to the origin (e.g., `tx`)~~ keeping suggested repo name for now (potentially revisit for unified documentation in the future)
- [x] Unpublish the gh-pages deployment from -tx, and publish the new gh-pages deployment from the docs repo
- [x] Delete the gh-pages branch from this repo

@maurermi mind taking a brief look at my docs MVP (https://github.com/HalosGhost/doc-test and https://github.com/HalosGhost/doc-test-ext) to see if you agree?

maurermi:
@HalosGhost Those experimental repos look good on my end

.github/workflows/pages.yml
```diff
@@ -30,4 +30,5 @@ jobs:
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          external_repository: mit-dci/opencbdc-tx-pages
           publish_dir: ./doxygen_generated/html
```
HalosGhost:
```suggestion
        uses: peaceiris/actions-gh-pages@v4
        with:
          deploy_key: ${{ secrets.TX_PAGES_DEPLOY_KEY }}
          external_repository: mit-dci/opencbdc-tx-pages
          publish_dir: ./doxygen_generated/html
```

I believe this change is sufficient to enable publishing to the default branch of the docs repository (largely copied from my docs MVP test above, tweaked appropriately). @maurermi if it makes sense, I can add the new secret and variable name changes to my test repo and deploy from that repo first.

HalosGhost (changes requested):
I've left a suggestion which, I believe, will leave the PR in a state ready-to-merge. Double-checking much appreciated! Let me know if the additional testing I proposed feels useful.

.github/workflows/pages.yml
maurermi: Yes that sounds good. I can commit this change as well

HalosGhost:
@maurermi I'm not bothered by it in this case, but note that git-trailers are expected to be in a contiguous block separated from the commit-body by a newline. [This has bitten us before](https://github.com/mit-dci/opencbdc-tx/blob/trunk/NEWS.md#2022-04-05-force-push-against-trunk).

HalosGhost (approved):
tACK. I've tested deployment to the target repo using a very close variant of the workflow shown here. Let's get the docs ripped out!

Merging.

PR #257, opened 2024-04-23
docs: mention new GH-Pages deployment in README

HalosGhost:
Before merging #255, the gh-pages branch of this repository was used for our live API reference instance. Unfortunately, that branch houses multiple binaries which were rebuilt on every-push, leading to the branch growing in-size without bound, and dramatically increasing the storage and network burden for all clones and forks.

Following #255 and the deletion of the gh-pages branch, the repo can be cloned almost instantly, and the burden is fully offloaded to a second repository which no one should need to clone under normal circumstances.

This commit simply adds reference to the new deployment (since GH's web interface is not smart enough to link to the external repo automatically).

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ Note that if you have not already installed the xcode cli tools you will need to
 ```
 # Run the Code
 
+The API Reference is now housed in [an external repository](https://github.com/mit-dci/opencbdc-tx-pages/).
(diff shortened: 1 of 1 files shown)


maurermi (approved):
tACK. Simple docs update to address new location of GH-pages. Links work on my end -- LGTM

PR #258, opened 2024-04-23
Pr/189 update

HalosGhost:
This simply rebases and updates #189.

I've done a bit of local testing, and this dramatically reduces the impact of enabling sentinel attestations.

@anders94 had a suggestion to prefer `n_threads` == `std::thread::hardware_concurrency() - 1`; I'm going to run a few tests to see if one or the other has an obvious performance benefit, and then I believe this is likely ready-to-merge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if(DEFINED CMAKE_PREFIX_PATH)
 endif()
 
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-    add_compile_options(-fprofile-arcs -ftest-coverage)
+    add_compile_options(-fprofile-arcs -ftest-coverage -Og -ggdb3)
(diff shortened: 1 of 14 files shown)


HalosGhost:
I've not yet addressed Anders's suggestion. Hoping to test that later this week.

HalosGhost:
@anders94 I did some initial (read: *not* thorough) testing reducing the validation-thread-count in the coordinators/shards by 1, and see negligible difference (if anything slightly higher latency / lower throughput); but I've not tested at particularly large configurations.

Do you have time later today or tomorrow to hop on a call and I can walk through the preliminary results I have and we can see if it feels like I should run a few more tests (i.e., just to eliminate the possibility of noise, or at larger configurations to potentially amplify the effects)?

anders94:
Yep, let's do it.

PR #259, opened 2024-04-23
Enable easy-to-run test-controller-style benchmarks locally

HalosGhost:
This is a WIP script that allows you to take a crafted configuration file (including one generated by the test controller), and execute a benchmark against that configuration using one process-per-component.

Needed before merging:

- [X] Support configurable test runtime
- [X] Preserve test artifacts out-of-tree
- [X] Cache and reuse preseeds (depends on a `<commit_hash,preseed_count,shard_count>` triple for uniqueness)
- [X] Generate basic performance plots
- [X] Support debugging and performance tooling
- [X] Support arbitrary configuration files from 2PC and Atomizer (including those generated by the test controller for large-scale tests)
- [x] Expose a pleasant command-line interface for easy feature-discovery and use

### DIFF
--- a/scripts/native-system-benchmark.sh
+++ b/scripts/native-system-benchmark.sh
@@ -1,0 +1,363 @@
+#!/usr/bin/env bash
+
+RR='rr record --'
+GDB='gdb -ex run --args'
+VALGRIND='valgrind --leak-check=full'
+DBG="${DBG:-$GDB}"
+
+# runs for DURATION seconds
+# if DURATION is set to inf/infinity, run indefinitely
+DURATION=30
+CWD=$(pwd)
+COMMIT=$(git rev-parse --short HEAD)
+TL=$(git rev-parse --show-toplevel)
+RT="${TL:-$CWD}"
+BLD="$RT"/build
+SEEDDIR="$BLD"/preseeds
+TESTDIR="$BLD"/test-$(date +"%s")
+
+IFS='' read -r -d '' usage <<'EOF'
+Usage: %s [options]
+
+Options:
+  -h, --help              print this help and exit
+  -c, --config=PATH       use PATH as the test configuration
+  -s, --samples=TIME      run test for TIME seconds (defaults to 30)
+                          (or indefinitely if set to inf, or infinity)
+
+Cleanup:
+  --clean                 delete all previous test and preseed artifacts
+  --clean-tests           delete all previous test artifacts
+  --clean-preseeds        delete all previous preseed artifacts
+
+Debugging Options:
+  --debug                 run each component under a debugger
+                          (likely requires a Debug build to be useful)
+  --profile               run each component under perf
+                          (likely requires a Profiling build to be useful)
+  --leak-check            run each component under valgrind
+
+  -d, --debugger=CMD      specify the debugger CMD
+                          ("gdb" and "rr" are short-cuts for sensible defaults
+                          for those debuggers)
+
+Note: --debug, --profile, and --leak-check are mutually-exclusive
+EOF
+
+_help=
+if [[ $# -eq 0 ]]; then
+    _help=1
+fi
+
+_err=0
+while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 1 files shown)


scripts/native-system-benchmark.sh
```diff
@@ -0,0 +1,358 @@
...
+
+# locate and move to test directory
+mkdir -p "$TESTDIR"
+printf 'Running test from %s\n' "$TESTDIR"
+cd "$TESTDIR" || exit
+
+# normalizes ports for local execution
+IFS='' read -r -d '' normalize <<'EOF'
+BEGIN {
+    i = 5000
```
maurermi: [suggestion] a lot of _stuff_ unofficially uses port 5000 (in my case, it conflicts with docker-registry) so I would suggest we start higher, (5002 seems much emptier: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers)

scripts/native-system-benchmark.sh
```diff
@@ -0,0 +1,358 @@
+#!/usr/bin/env bash
+
+RR='rr record --'
+GDB='gdb -ex run --args'
+VALGRIND='valgrind --leak-check=full'
+DBG="${DBG:-$GDB}"
+
+# runs for DURATION seconds
+# if DURATION is set to 0/inf/infinity, run indefinitely
```
maurermi: [suggestion] Not sure if it's standard to have tests with duration set to 0 run indefinitely, but I would be inclined to only support `DURATION=inf` and `DURATION=infinity` to have tests run indefinitely (maybe also `DURATION=''`).

scripts/native-system-benchmark.sh
```diff
@@ -0,0 +1,358 @@
...
+            esac
+            shift "$shft_cnt";;
+        -c*|--config*)
+            if [[ "$optarg" = /* ]]; then
+                ORIG_CFG="${optarg}"
+            else
+                ORIG_CFG="$CWD/$optarg"
+            fi
+            shift "$shft_cnt";;
+        -h|--help) _help=1; shift "$shft_cnt";;
```
maurermi, line 96: Should have unrecognized arguments print the help statement or exit. Currently, if an errant argument is provided the script gets stuck here

scripts/native-system-benchmark.sh
```diff
@@ -0,0 +1,358 @@
...
+            kill -SIGTERM -- "$i"
+        done
+    fi
+
+    if test -n "$(find "$TESTDIR" -maxdepth 1 -name '*.perf' -print -quit)"; then
+        printf 'Generating Flamegraphs\n'
+        for i in "$TESTDIR"/*.perf; do
+            waitpid -t 5 -e $(lsof -Qt "$i") &>/dev/null
+            perf script -i "$i" | stackcollapse-perf.pl > "${i/.perf/.folded}"
+            flamegraph.pl "${i/.perf/.folded}" > "${i/.perf/.svg}"
```
maurermi, line 183: Might want to ensure that FlameGraph is installed before running this part

maurermi (changes requested):
Mostly looks good to me, suggested a couple changes but I very much welcome this addition. 

One additional question, would it be possible to alert the user, or terminate the test if a component dies during the test? I think that could be quite useful especially for long running tests.

scripts/native-system-benchmark.sh
HalosGhost: This was just a stylistic choice because a benchmark duration of `0` seconds is effectively just a test that the system comes up correctly, and is otherwise non-sensical. I'm not bothered by allowing that use-case though if it feels helpful.

scripts/native-system-benchmark.sh
HalosGhost: Essentially nothing uses 298xx; switching to that.

maurermi (approved):
T-ACK. Looks good to me! Definitely a pleasant upgrade for running the system locally

PR #260, opened 2024-04-26
Add read locks to the Lua runner

maurermi:
This PR adds support for requesting read locks from Lua contracts. Now values yielded from contracts can include an optional second integer parameter (0 = read, else = write).

### DIFF
--- a/scripts/gen_bytecode.lua
+++ b/scripts/gen_bytecode.lua
@@ -1,5 +1,10 @@
 function gen_bytecode()
     pay_contract = function(param)
+        -- 0 is a read lock, 1 is a write lock in C++ scope
+        Locks = {
+            READ = 0,
+            WRITE = 1,
+        }
         from, to, value, sequence, sig = string.unpack("c32 c32 I8 I8 c64", param)
 
         function get_account_key(name)
@@ -8,9 +13,12 @@ function gen_bytecode()
 
         function get_account(name)
             account_key = get_account_key(name)
-            account_data = coroutine.yield(account_key)
-            if #account_data > 0 then
-                return string.unpack("I8 I8", account_data)
+
+            account_data = coroutine.yield(account_key, Locks.WRITE)
+            if string.len(account_data) > 0 then
+                account_balance, account_sequence
+                = string.unpack("I8 I8", account_data)
+                return account_balance, account_sequence
             end
             return 0, 0
         end
--- a/src/parsec/agent/runners/lua/impl.cpp
+++ b/src/parsec/agent/runners/lua/impl.cpp
@@ -138,25 +138,56 @@ namespace cbdc::parsec::agent::runner {
         return buf;
     }
 
+    auto lua_runner::get_stack_integer(int index) -> std::optional<int64_t> {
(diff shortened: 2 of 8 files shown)


HalosGhost:
CC @ayeshaali

scripts/gen_bytecode.lua
```diff
@@ -10,7 +10,9 @@ function gen_bytecode()
 
         function get_account(name)
             account_key = get_account_key(name)
-            account_data = coroutine.yield(account_key)
+
+            -- 0 is read lock, 1 is write lock
```
narula: perhaps consider using something more enum-like? https://unendli.ch/posts/2016-07-22-enumerations-in-lua.html

src/parsec/agent/runners/lua/impl.cpp
```diff
@@ -138,25 +138,51 @@ namespace cbdc::parsec::agent::runner {
         return buf;
     }
 
+    auto lua_runner::get_stack_integer(int index) -> std::optional<int64_t> {
+        if(lua_isinteger(m_state.get(), index) != 1) {
```
eolesinski, line 142: would it make sense to do this instead:

```
auto lua_runner::get_stack_integer(int index) -> std::optional<int64_t> {
         if(lua_isinteger(m_state.get(), index)) {
             return lua_tointeger(m_state.get(), index);
         }
         return std::nullopt;
     }
```

src/parsec/agent/runners/lua/impl.cpp
```diff
@@ -138,25 +138,51 @@ namespace cbdc::parsec::agent::runner {
...
-            if(n_results != 1) {
-                m_log->error("Contract yielded more than one key");
+            if(n_results > 2) {
+                m_log->error("Contract yielded more than two keys");
                 m_result_callback(error_code::yield_count);
                 return;
             }
+
+            auto lock_level = broker::lock_type::write;
+            if(n_results == 2) {
```
eolesinski, line 164: also it might make sense to add a check here to handle the case where  the call to `lua_resume` does not yield any results (unless it is the expectation that this will be handled in the contract specification)

src/parsec/agent/runners/lua/impl.cpp
maurermi: Good catch that perhaps cleaner, though on personal preference I am going to opt to flip the branches

maurermi:
Update: Responded to comments from @eolesinski, additionally caught an error from testing in `scripts/gen_bytecode.lua`, and rebased onto trunk

src/parsec/agent/runners/lua/impl.cpp
maurermi: Ah, apparently this is disliked by the linter! Reverting

maurermi:
@HalosGhost Added in a unit test that should make verification easier here

HalosGhost (approved):
This is an excellent addition, enabling dramatically more performant/parallelization contracts in the lua runner (namely due to enabling multiple, concurrent read-locks on a given key).

PR #268, opened 2024-06-13
removed references to the deleted scripts/configure.sh file and made …

rockett-m:
We no longer have `scripts/configure.sh` in the codebase as it has been replaced by `scripts/install-build-tools.sh` and `scripts/setup-dependencies.sh`. 

So removed references to the deprecated `configure.sh` file where it could be confusing to new people onboarding and filled in with the two build scripts we use in its place.

### DIFF
--- a/.github/workflows/docker-pull.yml
+++ b/.github/workflows/docker-pull.yml
@@ -26,7 +26,6 @@ jobs:
         with:
           files: |
             Dockerfile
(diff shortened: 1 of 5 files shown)


HalosGhost (approved):
Thank you for cleaning up a few of these dangling references! Looks good to me.

Merging

PR #269, opened 2024-06-14
Add method for emplacing lua contracts from CLI

maurermi:
Adds a way to emplace lua contracts into parsec from the CLI

Potentially a useful development tool

### DIFF
--- a/tools/bench/parsec/lua/contract_to_bytecode.lua
+++ b/tools/bench/parsec/lua/contract_to_bytecode.lua
@@ -1,0 +1,13 @@
+function gen_bytecode(func)
+    c = string.dump(func, true)
+    tot = ""
+    for i = 1, string.len(c) do
+        hex = string.format("%x", string.byte(c, i))
+        if string.len(hex) < 2 then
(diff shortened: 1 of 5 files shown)


maurermi:
@madars

tools/bench/parsec/lua/contract_to_bytecode.lua
```diff
@@ -0,0 +1,13 @@
+function gen_bytecode(func)
+    c = string.dump(func, true)
+    tot = ""
+    for i = 1, string.len(c) do
```
eolesinski, line 4: since strings are immutable, string concatenation in a for loop is generally discouraged due to inefficiency (it requires the creation of an entirely new string for each iteration).

It looks like Lua has a "string buffer" equivalent by using tables https://www.lua.org/pil/11.6.html and then just returning `table.concat` (PIL, pp. 110 and E 5.8 on pp. 41)

also it looks like you can just use the length operator `#` (PIL, pp. 24) rather than using a method call for the length

PR #273, opened 2024-06-26
made shell scripts all have 4 spaces for indents not 2; issue #272

### LINKED ISSUE #272: Make shell scripts have consistent indentation of 4 space chars
Roughly half the shell scripts use 2 spaces for indents and the rest use 4 spaces.

Our python and c++ code use 4 spaces as the indent, so it would be helpful to enforce this for our shell scripts too

rockett-m:
Now all shell scripts have indentation at 4 spaces. Previously it was mixed between 2 and 4 spaces for indents.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -21,7 +21,7 @@ if [ -n "$newline_files" ] ; then
 fi
 
 if [ -n "$whitespace_files" ] || [ -n "$newline_files" ] ; then
-  exit 1
+    exit 1
 fi
 
 check_format_files=$(git ls-files | grep -E "tools|tests|src|cmake-tests" \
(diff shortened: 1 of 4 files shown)


scripts/lint.sh
```diff
@@ -4,7 +4,7 @@ set -e
 echo "Linting..."
 
 check_files=$(git ls-files \
-              | grep -v -E ".jpg|.svg|3rdparty" | cat)
```
maurermi: I think this was intentional for alignment

maurermi (changes requested):
One requested change -- otherwise I think this is OK assuming we want to enforce 4 spaces instead of 2

scripts/lint.sh
rockett-m: ACK - fixed

rockett-m:
@HalosGhost is this good to merge?

scripts/native-system-benchmark.sh
```diff
@@ -20,26 +20,26 @@ IFS='' read -r -d '' usage <<'EOF'
...
-                          for those debuggers)
+    --debug                 run each component under a debugger
+                            (likely requires a Debug build to be useful)
+    --profile               run each component under perf
+                            (likely requires a Profiling build to be useful)
+    --leak-check            run each component under valgrind
+
+    -d, --debugger=CMD      specify the debugger CMD
+                            ("gdb" and "rr" are short-cuts for sensible defaults
+                            for those debuggers)
```
HalosGhost: This is a heredoc, a literal string that is manually-formatted to print on the terminal sensibly. Personally, I think this increase of indentation is a bit jarring (it's pretty atypical for help output to be indented so much), but I'm not particularly strong-feeling about it. However, as a matter of practice and habit, changing indentation level inside a string should absolutely not be treated the same as changing indentation level in code.

HalosGhost:
Additionally, instead of saying `issue #the-number`, it's much nicer if you say `closes #the-number` or `fixes #the-number` in commit messages; GitHub automatically picks those up and and links PRs to issues (closing the issue automatically on-merge). I've manually linked this one.

scripts/native-system-benchmark.sh
rockett-m: ACK, will revert to keep this as is

rockett-m (commented):
closing the review, thanks

HalosGhost (approved):
unifies indentation for shell scripts.

Looks good to me.

HalosGhost:
@maurermi when you have a moment; mind double-checking that there's nothing left that you'd want to see changed?

maurermi (approved):
ACK, just unifies indentation and makes no significant or concerning changes, looks good to me

PR #282, opened 2024-07-15
lua code improvements

eolesinski:
This PR makes various improvements to the `gen_bytecode.lua` code:

- removes extra variables
- replaces function calls to string length with operator # (much faster)
- replaces for loop string concatenation with lua "string buffer" idiom 
- replaces conditional with two digit, left-padding format specifier

### DIFF
--- a/scripts/gen_bytecode.lua
+++ b/scripts/gen_bytecode.lua
@@ -3,18 +3,14 @@ function gen_bytecode()
         from, to, value, sequence, sig = string.unpack("c32 c32 I8 I8 c64", param)
 
         function get_account_key(name)
-            account_prefix = "account_"
-            account_key = account_prefix .. name
-            return account_key
+            return "account_" .. name
         end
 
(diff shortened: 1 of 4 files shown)


maurermi (changes requested):
T-ACK. This PR improves the code-practices in the Lua script which we use for our Lua performance benchmark. I have verified that the script works as expected using our local bench test, and also verified that the bytecode generated is the same as before. 

This is ready to merge pending one change -- which is fixing the inconsistency of tabs / spaces in `scripts/gen_bytecode.lua`. Once this is changed I think this is good to go!

One further note, we have other lua scripts checked into this repository, and I think applying the same cleansing changes to those would be wise as well (and potentially would be good to include in this PR)

maurermi (approved):
T-ACK. Thanks for fixing the other lua files and the tab/space inconsistency. Looks like a good change to me

PR #285, opened 2024-07-18
ci: get docker merge passing; fixes #283

### LINKED ISSUE #283: CI/CD breakage
@rockett-m, it looks like the [docker merge](https://github.com/mit-dci/opencbdc-tx/actions/runs/9980786103/job/27583948824) CI/CD step is failing from the merge of #265. It's not spurious (rerunning the job failed as well). Help resolving would be appreciated.

rockett-m:
> Changed docker merge stage to run on pull-request in addition to merge to see if it will pass or fail before a merge takes place. `.github/workflows/docker-push.yml`
Although it is for testing, it might be valuable to keep this to avoid future issues

> Edited the`.dockerignore` and `Dockerfile` so that `scripts/activate-venv.sh` is found
Error:
```console
...
251.1 scripts/install-build-tools.sh: line 87: /scripts/activate-venv.sh: No such file or directory
-----------------------------------
Dockerfile:23
-----------------------------------
21 |      WORKDIR /opt/tx-processor
22 | 
23 |  >>> RUN scripts/install-build-tools.sh
24 |      RUN scripts/setup-dependencies.sh
----------------------------------
ERROR: failed to solve: process "/bin/sh -c scripts/install-build-tools.sh" did not complete successfully: exit code: 1
...
```

> Added '.' to `scripts/install-build-tools.sh` sourcing the venv instead of 'source' due to Michael's point about sh not being able to run source.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,6 +13,7 @@ charts/
 # Prevent copying scripts that are unused in docker
 scripts/
 !scripts/test.sh
+!scripts/activate-venv.sh
 !scripts/install-build-tools.sh
 !scripts/setup-dependencies.sh
 !scripts/test-transaction.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,12 +8,14 @@ ARG BASE_IMAGE="ghcr.io/mit-dci/opencbdc-tx-base:latest"
 FROM $IMAGE_VERSION AS base
 
 # set non-interactive shell
-ENV DEBIAN_FRONTEND noninteractive
-ENV CMAKE_BUILD_TYPE Release
-ENV BUILD_RELEASE 1
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CMAKE_BUILD_TYPE=Release
+ENV BUILD_RELEASE=1
 
 RUN mkdir -p /opt/tx-processor/scripts
 
+COPY requirements.txt /opt/tx-processor/requirements.txt
+COPY scripts/activate-venv.sh /opt/tx-processor/scripts/activate-venv.sh
 COPY scripts/install-build-tools.sh /opt/tx-processor/scripts/install-build-tools.sh
 COPY scripts/setup-dependencies.sh /opt/tx-processor/scripts/setup-dependencies.sh
 
--- a/scripts/build-docker.sh
+++ b/scripts/build-docker.sh
@@ -1,5 +1,5 @@
+#!/usr/bin/env bash
 
-#!/bin/bash
 set -e
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
--- a/scripts/install-build-tools.sh
+++ b/scripts/install-build-tools.sh
@@ -51,7 +51,8 @@ create_venv_install_python() {
         if rm -rf -- "${ENV_LOCATION}"; then
             echo "Removed existing virtual environment"
         else
-            echo "Failed to remove existing virtual environment"; exit 1
+            echo "Failed to remove existing virtual environment"
+            exit 1
         fi
     fi
     # install pip for linux
@@ -82,10 +83,14 @@ create_venv_install_python() {
     if "${PY_LOC}" -m venv "${ENV_NAME}"; then
         echo "Virtual environment '${ENV_NAME}' created"
     else
-        echo "Virtual environment creation failed"; exit 1
+        echo "Virtual environment creation failed"
+        exit 1
     fi
     # activate virtual environment
-    source "${ROOT}/scripts/activate-venv.sh"
+    if ! . "${ROOT}/scripts/activate-venv.sh"; then
+        echo "Failed to activate virtual environment"
+        exit 1
+    fi
     # install python packages
     if pip install -r "${ROOT}/requirements.txt"; then
         echo "Success installing python packages"


rockett-m:
@HalosGhost and/or @maurermi would you be able to review this? Thanks

.github/workflows/docker-merge.yml
```diff
@@ -6,6 +6,10 @@ on:
   push:
     branches:
       - 'trunk'
+  # to test pass/failure before a merge
+  pull_request:
```
maurermi: I understand why this was added in, but I am not sure that we want this here. 

I am not 100% sure what the reason is that we have different workflows for pull, push, and merge, but my gut reaction is we should not modify when they are run and instead synchronize these workflows (if they've gotten out of sync). 

@HalosGhost would be good to get your insight here (or out of band) if possible.

@rockett-m I think this should be removed from this commit for now

maurermi (commented):
UT-ACK. This looks correct, will give it a test shortly, but I have high confidence. Thanks for patching this quickly.

.dockerignore
```diff
@@ -3,6 +3,7 @@ run
 plots
 .vscode
 .deps
+!requirements.txt
```
HalosGhost: is `requirements.txt` ignored by `docker` by-default? If not, is there a reason to add this?

.github/workflows/docker-merge.yml
HalosGhost: I need to remind myself of why we had this separation; it was definitely intentional, but I cannot recall what the purpose was. Let me dig into this and see if I can recover an answer.

So, the workflows are intentionally separated because they do different things.
* `push` just runs on non-`trunk` branches so that images are made available for non-`trunk` branches off this upstream repo
* `pull` operates only on pull-requests and is designed to build the images using the changes in the PR to make sure they work
* `merge` operates only on merging to `trunk` and manages the publishing of new images

I agree with @maurermi; if anything, it looks like perhaps `merge` and `pull` have gotten out-of-sync (though nothing really jumps out at me as to what) rather than that `merge` should be run on PR---this should be removed.

.dockerignore
rockett-m: Right, so for activate-venv.sh and requirements.txt I encountered failures where those files could not be found (and are needed by install-build-tools.sh). So it was necessary to add the 'COPY' lines and this ! exception argument to the best of my understanding

.github/workflows/docker-merge.yml
rockett-m: If we don't run docker merge before a merge takes place then we can have future CI/CD breakage

Dockerfile
```diff
@@ -7,10 +7,16 @@ ARG BASE_IMAGE="ghcr.io/mit-dci/opencbdc-tx-base:latest"
...
+# make shell run with bash not sh
+SHELL ["/bin/bash", "-c"]
+
 # set non-interactive shell
-ENV DEBIAN_FRONTEND noninteractive
-ENV CMAKE_BUILD_TYPE Release
-ENV BUILD_RELEASE 1
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CMAKE_BUILD_TYPE=Release
+ENV BUILD_RELEASE=1
```
rockett-m, line 13: Fix for these legacy warnings 
```console
3 warnings found (use --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 11)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 12)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 13)
```

.dockerignore
maurermi: I think this should work without `requirements.txt` in `.dockerignore` but you are right an explicit `COPY` command is required in `Dockerfile` for `requirements.txt`

Dockerfile
```diff
@@ -7,10 +7,16 @@ ARG BASE_IMAGE="ghcr.io/mit-dci/opencbdc-tx-base:latest"
...
+
```
HalosGhost: Are these necessary?

Dockerfile
```diff
@@ -7,10 +7,16 @@ ARG BASE_IMAGE="ghcr.io/mit-dci/opencbdc-tx-base:latest"
...
+COPY requirements.txt /opt/tx-processor/requirements.txt
+COPY scripts/activate-venv.sh /opt/tx-processor/scripts/activate-venv.sh
```
HalosGhost: These two lines should be moved down to be with the other copy-scripts-to-container lines.

Dockerfile
rockett-m: they were already there, I just added the = sign so there aren't warnings

.github/workflows/docker-merge.yml
HalosGhost: That's because, in this case, `pull` somehow didn't fail despite `activate-venv.sh` not being copied to the container. `merge` did fail; i.e., the two actions have a gap in functionality. *That* is what should be fixed; not having `merge` run on pull-requests.

Dockerfile
rockett-m: ACK - moved

.dockerignore
rockett-m: Let me run without this line and see what happens
Edit: passes, good dd

.github/workflows/docker-merge.yml
rockett-m: ACK - makes sense

Edit - red herring - ignore

doing a comparison between the docker-pull and docker-merge files, I see a point of concern

> Limiting fetch depth in docker-pull (to only the most recent 2 commits) - necessary?
```
    ########################
    # CI Setup             #
    ########################
    - name: Checkout code
      uses: actions/checkout@v3
        with:                                        # only in docker-pull
          fetch-depth: "2"                           # only in docker-pull
```
https://git-scm.com/docs/git-fetch

HalosGhost: Note: I'm comfortable getting this merged as a fast-follow without having solved the gap between `pull` and `merge`, but this PR should not add running `merge` on PRs.

rockett-m (commented):
feedback updated

I will remove the additional docker-merge run on pull-request when we are good on the changes pre-merge

docker-build.sh runs successfully on ubuntu 22.04 disk image with these changes fyi

.github/workflows/docker-merge.yml
HalosGhost: I wouldn't think the workflow doing a shallow clone would cause any issue here (it will still have all the files present at both of those commits)—I would still expect the docker build to fail if the file isn't copied to the container as-required.

scripts/build-docker.sh
```diff
@@ -1,5 +1,5 @@
 
-#!/bin/bash
+#!/usr/bin/env bash
```
rockett-m, line 1: should be on line 1. do we prefer /bin/bash or the /usr/bin/env bash?

.github/workflows/docker-merge.yml
rockett-m: TACK - this was not the issue

scripts/build-docker.sh
HalosGhost: `#!/usr/bin/env bash` is the most portable `bash` shebang; and it should be preferred unless there is an explicit reason not to (I'm unaware of a general one).

HalosGhost: And yes, it should be on line-1.

rockett-m:
Just ran and 8/8 checks passed. I had the docker merge stage included to see if it would pass.

I have now removed just the docker-merge on pull-request and re-pushed the commit. Should be good to go now

HalosGhost (approved):
in a seemingly related issue `act` is now broken testing against `trunk`, so I cannot reproduce this working locally; however, I saw the passing CI test, and I see that the `push` workflow finished correctly on your fork. I'll merge this in-hopes that it solves the immediate CI/CD problem, but a fix to generally narrow the gap between `pull` and `merge` docker workflows, ideally also fixing `act` for testing will be much appreciated.

PR #286, opened 2024-07-19
fix: include the prefix directory in .dockerignore

maurermi:
Noticed on macOS that ./scripts/build-docker.sh was failing. Our Dockerfile runs ./scripts/setup-dependencies.sh in the container before running `COPY . .`, which if prefix/ is not ignored, will copy in the libraries built on the host.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@ run
 plots
 .vscode
 .deps
+prefix
 
 # E2E test results
 testruns


rockett-m:
TACK - this worked for me running on mac and ubuntu

HalosGhost:
@eolesinski I'd appreciate another macOS sign-off/review if you have the time (I don't have an easy way to give a t-ACK here).

eolesinski (approved):
t-ACK - this change successfully fixes the error that was occurring when running the `build-docker.sh` script

Tested by running `build-docker.sh` before this change and encountered an error. Pulled in the changes from this PR and reran the `build-docker.sh` and it worked.

HalosGhost (approved):
Approved (by-proxy t-ACK from @eolesinski and @rockett-m)

Issue #97, opened 2022-05-01
Some local tests may fail if port number is not available


davebryson:
### Affected Branch

trunk

### Basic Diagnostics

- [X] I've pulled the latest changes on the affected branch and the issue is still present.

- [ ] The issue is reproducible in docker

### Description

Testing outside of Docker may fail if a port number is needed for the test and the port is in use.  In some cases, this may result in a seg fault  which can be misleading.  For example, the unit tests in `raft_test.app` create 3 nodes using the port range 5000-5002.  The port range has caused seg faults on my machine as port 5000 was in use.  Changing the ports resolves the issue.

I'm not sure what the best fix for this is.  Finding available ports (in code) can be a little tricky and unreliable.  The other alternative is to just test in Docker and specify that testing outside of Docker is not recommended....

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

HalosGhost:
I've run into this before, and we do have one potential solution for it housed under #3 (particularly in the task `Add support for OS-assigned ephemeral port numbers and leverage them in-tests`).

It should be possible to listen on port 0 (which should fetch an available port).

HalosGhost:
via #289.

PR #289, opened 2024-07-24
change default ports

eolesinski:
This PR addresses issue #97 by updating the port numbers used in the `raft_test.cpp` unit tests to a higher range (29800-29802) to avoid conflicts with commonly used ports.

Port conflicts with 5000 are frequent, especially with the docker-registry and Apple's AirPlay Receiver on MacOS. To address this issue, the default ports are updated to 298xx, following the convention outlined in #259.

Tested: tested by running `./scripts/test.sh` 

**EDIT:** This change now updates all test endpoints that use ports at 5xxx


HalosGhost:
It looks like there are a ~lot of~few other tests which use 5xxx portnumbers; are you hoping to address those all in this PR as well?

In particular, the `cfg` files used by the integration test suites.

eolesinski:
@HalosGhost Yep - I can go through the other tests and make those changes

maurermi (approved):
T-ACK. This is useful when running on my local system since port 5000 is often occupied. I approve of these changes but second @HalosGhost 's comment that this should be replicated elsewhere in the code.

maurermi (approved):
T-ACK. Updates the ports-used-for-test to be outside the range of typically used ports. MacOS AirPlay users rejoice.

HalosGhost (approved):
T-ACK. Just swaps out the default ports we use to something less crowded. Hopefully defaults to causing far fewer headaches!

Thanks @eolesinski

PR #297, opened 2024-07-31
Update CMakeLists.txt

JohnbelMDev:


HalosGhost:
Hello!

The most recent revision of C++ is C++23, and we just upgraded to target 20 not long ago. Is there a reason you'd suggest updating the target version to `30`?

HalosGhost:
I think this might have been unintentional. Closing.

PR #298, opened 2024-08-07
ci: add shellcheck gh action, lint fixes, parallel acts

rockett-m:
This adds a shellcheck github workflow to check lint our shell files after discussion with @HalosGhost 

I fixed a bunch of warnings / errors / info messages from shellcheck across our *.sh scripts.

Edit: added
I assume we will want a `scripts/shellcheck.sh` script to tidy up the `.github/workflows/ci.yml :: shellcheck` stage - and to have an easy way for users to run shellcheck / view reports.

`act -j shellcheck --container-architecture linux/amd64`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Setup Build Env
+      - name: Install Build Tools
         run: sudo ./scripts/install-build-tools.sh
       - name: Setup Local Dependencies
         run: ./scripts/setup-dependencies.sh
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Setup Build Env
+      - name: Install Build Tools
         run: sudo ./scripts/install-build-tools.sh
       - name: Setup Local Dependencies
         run: ./scripts/setup-dependencies.sh
@@ -50,7 +50,7 @@ jobs:
     name: Pylint
     runs-on: ubuntu-22.04
     continue-on-error: true
-    timeout-minutes: 10
+    timeout-minutes: 5
     strategy:
       matrix:
         python-version: ["3.10"]
@@ -62,10 +62,25 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Setup Build Env
+      - name: Install Build Tools
         run: sudo ./scripts/install-build-tools.sh
       - name: Lint with Pylint
         run: ./scripts/pylint.sh
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-22.04
+    continue-on-error: true
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install shellcheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+      - name: Lint with Shellcheck
+        run: ./scripts/shellcheck.sh
   unit-and-integration-test:
     name: Unit and Integration Tests
     runs-on: ubuntu-22.04
@@ -74,7 +89,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
-      - name: Setup Build Env
+      - name: Install Build Tools
         run: sudo ./scripts/install-build-tools.sh
       - name: Setup Local Dependencies
         run: ./scripts/setup-dependencies.sh
@@ -84,7 +99,7 @@ jobs:
         run: ./scripts/test.sh
       - name: Shorten SHA
         id: vars
-        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+        run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
       - uses: actions/upload-artifact@v4
         if: ${{ !env.ACT }}
         name: Archive Test Results
@@ -114,4 +129,3 @@ jobs:
           name: OpenCBDC Transaction Processor docs for ${{ steps.vars.outputs.sha_short }}
           path: ./doxygen_generated/html/*
           retention-days: 7
-
--- a/scripts/benchmarks.sh
+++ b/scripts/benchmarks.sh
@@ -64,6 +64,7 @@ done
 # to the location of this script, the user can run this script from any folder.
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 REPO_TOP_DIR="${SCRIPT_DIR}/.."
+BUILD_DIR=
 if [[ -z "${BUILD_DIR+x}" ]]
 then
     BUILD_DIR="${REPO_TOP_DIR}/build"
@@ -78,7 +79,8 @@ fi
 # If the build folder is a relative path, convert it to an absolute path
 # to avoid potential relative path errors and to improve readability
 # if the path is written to stdout.
-export BUILD_DIR=$(cd "$BUILD_DIR"; pwd)
+BUILD_DIR=$(cd "$BUILD_DIR"; pwd)
+export BUILD_DIR
 echo "Build folder: '${BUILD_DIR}'"
 echo
 
--- a/scripts/build-docker.sh
+++ b/scripts/build-docker.sh
@@ -12,7 +12,7 @@ DOCKER_IMAGE_TAG_TWOPHASE=${DOCKER_IMAGE_TAG:-opencbdc-tx-twophase}
 git submodule init && git submodule update
 
 # Build docker image
-docker build --target base -t $DOCKER_IMAGE_TAG_BASE -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target builder --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_BUILDER -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target twophase --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_TWOPHASE -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target atomizer --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_ATOMIZER -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
+docker build --target base -t "$DOCKER_IMAGE_TAG_BASE" -f "$SCRIPT_DIR"/../Dockerfile "$SCRIPT_DIR"/..
+docker build --target builder --build-arg BASE_IMAGE=base -t "$DOCKER_IMAGE_TAG_BUILDER" -f "$SCRIPT_DIR"/../Dockerfile "$SCRIPT_DIR"/..
+docker build --target twophase --build-arg BASE_IMAGE=base -t "$DOCKER_IMAGE_TAG_TWOPHASE" -f "$SCRIPT_DIR"/../Dockerfile "$SCRIPT_DIR"/..
+docker build --target atomizer --build-arg BASE_IMAGE=base -t "$DOCKER_IMAGE_TAG_ATOMIZER" -f "$SCRIPT_DIR"/../Dockerfile "$SCRIPT_DIR"/..
--- a/scripts/create-e2e-report.sh
+++ b/scripts/create-e2e-report.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -e
-SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
 TESTRUN_PATH=$1
 
 function readAndFormatLogs() {
@@ -11,15 +11,15 @@ function readAndFormatLogs() {
         return
     fi
 
-    for logfile in $(ls $logdir); do
+    for logfile in "$logdir"/*; do
         logfile_path="$logdir/$logfile"
-        logfile_content=$(cat $logfile_path)
+        logfile_content=$(<"$logfile_path")
         message+="\n<details>\n<summary>$logfile</summary>\n\n\`\`\`\n$logfile_content\n\`\`\`\n</details>\n"
     done
     echo "$message"
 }
 
-testrun_logs="\n<details>\n<summary>View Testrun</summary>\n\n\`\`\`\n$(cat $TESTRUN_PATH/testrun.log)\n\`\`\`\n</details>\n\n"
-container_logs=$(readAndFormatLogs $TESTRUN_PATH/logs)
+testrun_logs="\n<details>\n<summary>View Testrun</summary>\n\n\`\`\`\n$(cat "$TESTRUN_PATH"/testrun.log)\n\`\`\`\n</details>\n\n"
+container_logs=$(readAndFormatLogs "$TESTRUN_PATH"/logs)
 
 printf "# E2E Results\n# TestRun Logs\n%b\n\n# Container Logs\n%b\n" "$testrun_logs" "$container_logs"
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -14,6 +14,7 @@ if [ -n "$whitespace_files" ]; then
     printf '%s\n' "${whitespace_files[@]}"
 fi
 
+# shellcheck disable=SC2016
 newline_files=$(printf '%s' "${check_files[@]}" | xargs -r -I {} bash -c 'test "$(tail -c 1 "{}" | wc -l)" -eq 0 && echo {}' | cat)
 
 if [ -n "$newline_files" ] ; then
@@ -25,9 +26,12 @@ if [ -n "$whitespace_files" ] || [ -n "$newline_files" ] ; then
     exit 1
 fi
 
-check_format_files=$(git ls-files | grep -E "tools|tests|src|cmake-tests" \
-                     | grep -E "\..*pp")
-clang-format --style=file --Werror --dry-run ${check_format_files[@]}
+check_format_files=$(git ls-files | \
+                     grep -E "tools|tests|src|cmake-tests" | \
+                     grep -E "\..*pp")
+
+echo "${check_format_files}" | \
+    xargs -n1 -I{} clang-format --style=file --Werror --dry-run {}
 
 if ! command -v clang-tidy &>/dev/null; then
     echo "clang-tidy does not appear to be installed"
@@ -46,6 +50,6 @@ fi
 
 # use python from the virtual environment for clang-tidy
 if source "./scripts/activate-venv.sh"; then
-    python /usr/local/bin/run-clang-tidy.py -p ${BUILD_DIR} "tests/.*/.*\.cpp|src/.*/.*\.cpp|tools/.*/.*\.cpp"
+    python /usr/local/bin/run-clang-tidy.py -p "$BUILD_DIR" "tests/.*/.*\.cpp|src/.*/.*\.cpp|tools/.*/.*\.cpp"
     deactivate
 fi
--- a/scripts/native-system-benchmark.sh
+++ b/scripts/native-system-benchmark.sh
@@ -107,7 +107,8 @@ case "$DURATION" in
 esac
 
 if [[ -n "$_help" ]]; then
-    printf "$usage" "$(basename $0)"
+    # shellcheck disable=SC2059
+    printf "$usage" "$(basename "$0")"
     exit "$_err"
 fi
 
@@ -151,7 +152,7 @@ on_int() {
     printf 'Interrupting all components\n'
     trap '' SIGINT # avoid interrupting ourself
     for i in $PIDS; do # intentionally unquoted
-        if [[ -n "RECORD" ]]; then
+        if [[ -n "$RECORD" ]]; then
             kill -SIGINT -- "-$i"
         else
             kill -SIGINT -- "$i"
@@ -178,7 +179,7 @@ on_int() {
     if [[ -x "$(which flamegraph.pl)" && -x "$(which stackcollapse-perf.pl)" && -n "$(find "$TESTDIR" -maxdepth 1 -name '*.perf' -print -quit)" ]]; then
         printf 'Generating Flamegraphs\n'
         for i in "$TESTDIR"/*.perf; do
-            waitpid -t 5 -e $(lsof -Qt "$i") &>/dev/null
+            waitpid -t 5 -e "$(lsof -Qt "$i")" &>/dev/null
             perf script -i "$i" | stackcollapse-perf.pl > "${i/.perf/.folded}"
             flamegraph.pl "${i/.perf/.folded}" > "${i/.perf/.svg}"
             rm -- "${i/.perf/.folded}"
@@ -194,7 +195,7 @@ on_int() {
 
     printf 'Terminating any remaining processes\n'
     for i in $PIDS; do # intentionally unquoted
-        if [[ -n "RECORD" ]]; then
+        if [[ -n "$RECORD" ]]; then
             kill -SIGTERM -- "-$i"
         else
             kill -SIGTERM -- "$i"
@@ -253,7 +254,7 @@ run() {
     COMP=
     case "$RECORD" in
         perf)
-            $@ &> "$PROC_LOG" &
+            "$@" &> "$PROC_LOG" &
             COMP="$!"
             perf record -F 99 -a -g -o "$PNAME".perf -p "$COMP" &> "$PERF_LOG" &
             PERFS="$PERFS $!";;
@@ -261,7 +262,7 @@ run() {
             ${DBG} "$@" &> "$PROC_LOG" &
             COMP="$!";;
         *)
-            $@ &> "$PROC_LOG" &
+            "$@" &> "$PROC_LOG" &
             COMP="$!";;
     esac
 
@@ -287,9 +288,9 @@ seed() {
     if test ! -e "$SEEDDIR"/"$preseed_id"; then
         printf 'Creating %s\n' "$preseed_id"
         mkdir -p -- "$SEEDDIR"/"$preseed_id"
-        pushd "$SEEDDIR"/"$preseed_id" &> /dev/null
+        pushd "$SEEDDIR"/"$preseed_id" &> /dev/null || exit
         PID=$(PNAME=seeder BLOCK=1 run "$(getpath seeder)" "$CFG")
-        popd &> /dev/null
+        popd &> /dev/null || exit
     fi
 
     printf 'Using %s as seed\n' "$preseed_id"
@@ -320,12 +321,13 @@ launch() {
                 for node in $(seq 0 $(( "$raft" - 1 )) ); do
                     export PNAME="$1${id}_$node"
                     PID=$(run "$(getpath "$1")" "$CFG" "$id" "$node")
+                    # shellcheck disable=SC2013
                     for ep in $(awk -F'[":]' "/$PNAME.*endpoint/ { print \$3 }" "$CFG"); do
                         "$RT"/scripts/wait-for-it.sh -q -t 5 -h localhost -p "$ep"
                     done
                     printf 'Launched logical %s %d, replica %d [PID: %d]\n' "$1" "$id" "$node" "$PID"
-                    if [[ -n "RECORD" ]]; then
-                        PIDS="$PIDS $(getpgid $PID)"
+                    if [[ -n "$RECORD" ]]; then
+                        PIDS="$PIDS $(getpgid "$PID")"
                     else
                         PIDS="$PIDS $PID"
                     fi
@@ -333,12 +335,13 @@ launch() {
             else
                 export PNAME="$1${id}"
                 PID=$(run "$(getpath "$1")" "$CFG" "$id")
+                # shellcheck disable=SC2013
                 for ep in $(awk -F'[":]' "/$PNAME.*endpoint/ { print \$3 }" "$CFG"); do
                     "$RT"/scripts/wait-for-it.sh -q -t 5 -h localhost -p "$ep"
                 done
                 printf 'Launched %s %d [PID: %d]\n' "$1" "$id" "$PID"
-                if [[ -n "RECORD" ]]; then
-                    PIDS="$PIDS $(getpgid $PID)"
+                if [[ -n "$RECORD" ]]; then
+                    PIDS="$PIDS $(getpgid "$PID")"
                 else
                     PIDS="$PIDS $PID"
                 fi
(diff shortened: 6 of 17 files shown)


scripts/test-transaction.sh
```diff
@@ -11,27 +11,27 @@ wallet1="wallet1.dat"
 client1="client1.dat"
 
 # Create wallet0 and save wallet id to var, mint
-wallet0_id=$($BUILD_DIR/src/uhs/client/client-cli $config_file $client0 $wallet0 mint 10 10 | grep -E '^[^\W]+$')
+wallet0_id=$("${BUILD_DIR}/src/uhs/client/client-cli" "${config_file}" "${client0}" "${wallet0}" mint 10 10 | grep -E '^[^\W]+$')
 
-echo "Wallet0_PK: $wallet0_id"
+echo "Wallet0_PK: ${wallet0}_id"
```
rockett-m: wallet0_id is unused.
Should it be on line 16 "Wallet0_PK: ${wallet0}_id"?

I see wallet1_id on line 27.
On line 24 we have "Wallet1_PK: ${wallet1}_id", which makes me think line 16 is intentional ...

benchmarks/transactions.cpp
```diff
@@ -41,21 +41,21 @@ void reset_wallets(cbdc::transaction::wallet& w1,
...
     std::optional<cbdc::transaction::full_tx> maybe_tx{};
     state.ResumeTiming();
-    maybe_tx = sender.send_to(n_in * 2, reciever.generate_key(), true).value();
+    maybe_tx = sender.send_to(n_in * 2, receiver.generate_key(), true).value();
     state.PauseTiming();
     if(maybe_tx.has_value()) {
         sender.confirm_transaction(*maybe_tx);
-        reciever.confirm_transaction(*maybe_tx);
+        receiver.confirm_transaction(*maybe_tx);
         return true;
```
rockett-m: Typo fix unrelated to the shell files

scripts/lint.sh
```diff
@@ -1,51 +1,62 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 # Usage: ./scripts/lint.sh
 echo "Linting..."
 
-check_files=$(git ls-files \
-              | grep -v -E ".jpg|.svg|3rdparty" | cat)
+# GIT_FILES=$(git ls-files)
```
rockett-m: This should be deleted before submission

.github/workflows/ci.yml
```diff
@@ -66,6 +66,35 @@ jobs:
...
+          echo "Shellcheck report: shellcheck-report.txt"
+          cat shellcheck-report.txt
+      - name: Detect Shellcheck errors
+        run: |
+          if grep -qE "error" shellcheck-report.txt; then
+            echo "Shellcheck found errors in report: shellcheck-report.txt"
+            exit 1
+          else
+            echo "Shellcheck found no fatal errors in report: shellcheck-report.txt"
+          fi
```
rockett-m: Can make a script like pylint and consolidate this if desired

scripts/test-transaction.sh
HalosGhost: `wallet0_id` was used on this line; the change you made here (adding the braces) excludes `_id` from the variable name making it unusued. This is not correct or desired.

You made the same change on line 24.

If, instead, you mean that `wallet0_id` is never used in the actual transaction flow, that's because `wallet0` is Alice's wallet, and she doesn't need to invoke her own address (which is stored in her wallet file) in order to pay Bob (`wallet1`).

.github/workflows/ci.yml
HalosGhost: Is there a reason to prefer manually installing and running `shellcheck` in a bare container over using a shellcheck GH action? (e.g., https://github.com/marketplace/actions/shellcheck)

benchmarks/transactions.cpp
HalosGhost: Please extract unrelated fixes into a separate commit (fine to be in the same PR).

scripts/benchmarks.sh
```diff
@@ -69,7 +69,7 @@ then
     BUILD_DIR="${REPO_TOP_DIR}/build"
 fi
 
-if [[ ! -d "$BUILD_DIR" ]]
+if [[ ! -d "${BUILD_DIR}" ]]
```
HalosGhost: A lot of the added braces in this PR aren't really doing anything; this is a good example. Was this recommended by shellcheck?

scripts/build-docker.sh
```diff
@@ -12,7 +12,7 @@ DOCKER_IMAGE_TAG_TWOPHASE=${DOCKER_IMAGE_TAG:-opencbdc-tx-twophase}
...
 
 # Build docker image
-docker build --target base -t $DOCKER_IMAGE_TAG_BASE -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target builder --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_BUILDER -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target twophase --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_TWOPHASE -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
-docker build --target atomizer --build-arg BASE_IMAGE=base -t $DOCKER_IMAGE_TAG_ATOMIZER -f $SCRIPT_DIR/../Dockerfile $SCRIPT_DIR/..
+docker build --target base -t "${DOCKER_IMAGE_TAG_BASE}" -f "${SCRIPT_DIR}/../Dockerfile" "${SCRIPT_DIR}/.."
+docker build --target builder --build-arg BASE_IMAGE=base -t "${DOCKER_IMAGE_TAG_BUILDER}" -f "${SCRIPT_DIR}/../Dockerfile" "${SCRIPT_DIR}/.."
+docker build --target twophase --build-arg BASE_IMAGE=base -t "${DOCKER_IMAGE_TAG_TWOPHASE}" -f "${SCRIPT_DIR}/../Dockerfile" "${SCRIPT_DIR}}/.."
+docker build --target atomizer --build-arg BASE_IMAGE=base -t "${DOCKER_IMAGE_TAG_ATOMIZER}" -f "${SCRIPT_DIR}/../Dockerfile" "${SCRIPT_DIR}}/.."
```
HalosGhost: It looks like there are errant, additional closing braces on these lines.

.github/workflows/ci.yml
rockett-m: Ahh I didn't realize there was a built-in shellcheck.
Since the script is included now, it will be slightly easier for users to test their changes locally. We could use the gh action shellcheck - open to discussion
Edit: after trying many of the shellchecks from other repos failed, so agreement to make a script and call it command-line shellcheck instead.

scripts/native-system-benchmark.sh
```diff
@@ -46,80 +46,106 @@ EOF
...
-                ORIG_CFG="$CWD/$optarg"
-            fi
-            shift "$shft_cnt";;
-        -h|--help) _help=1; shift "$shft_cnt";;
-        *)
-            printf 'Unrecognized option: %s\n' "$1"
-            _help=1; _err=1;
-            break;;
-    esac
+	optarg=
```
HalosGhost: None of the other files use tabs for indentation; is there a reason this portion has been changed to tabs / reformatted?

scripts/native-system-benchmark.sh
```diff
@@ -46,80 +46,106 @@ EOF
...
-    '') DURATION=30;;
+inf | infinity) DURATION=infinity ;;
+'') DURATION=30 ;;
 esac
 
-if [[ -n "$_help" ]]; then
-    printf "$usage" "$(basename $0)"
-    exit "$_err"
+if [[ -n ${_help} ]]; then
+	printf "%s %s" "${usage}" "$(basename "$0")"
```
HalosGhost: This is not equivalent. `usage` includes a `%s` as it is a format string intended for use as the first argument to `printf`.

scripts/native-system-benchmark.sh
```diff
@@ -46,80 +46,106 @@ EOF
...
+	exit "${_err}"
 fi
 
-if [[ -z "$ORIG_CFG" ]]; then
-    printf '%s\n' 'No config specified; exiting'
-    exit 0
+if [[ -z ${ORIG_CFG} ]]; then
```
HalosGhost: braces were added but quotes were removed even though this is a time where quoting is important.

scripts/setup-dependencies.sh
```diff
@@ -160,8 +160,11 @@ rm libmicrohttpd-0.9.75.tar.gz
 cd libmicrohttpd-0.9.75
 mkdir build
 cd build
-../configure --prefix="${PREFIX}" --disable-curl --disable-examples --disable-doc --disable-shared --disable-https
-make -j $CPUS
+../configure "--prefix=${PREFIX}" --disable-curl --disable-examples --disable-doc --disable-shared --disable-https
```
HalosGhost: Why move the quote to before the `--`?

HalosGhost (changes requested):
There are a lot of style-only changes included here (namely, a dramatic but non-universal adoption of bracing variable expansions); I'm fine with these being included (they are, at worst, benign so long as the braces don't change the expansion, which some of the included ones do), but it's a ton of noise making the review much harder.

In the future, any style-only changes should be in a separate commit so they can be reviewed separately; ideally including a script which could be used to apply the style change so that the reviewer can verify the changes are entirely mechanical style modifications.

rockett-m:
@HalosGhost I did see the braces recommendation a lot via shellcheck. A warning not an error. I tried to clear as many warnings as I could for this PR

https://www.shellcheck.net/wiki/SC2250

benchmarks/transactions.cpp
rockett-m: done in #299

scripts/setup-dependencies.sh
rockett-m: ACK, looks better how it was, changed back for all similar

scripts/native-system-benchmark.sh
rockett-m: ACK, unintentional, reverted

scripts/native-system-benchmark.sh
rockett-m: ACk, fixed

scripts/native-system-benchmark.sh
rockett-m: ACK, reverted

scripts/test-transaction.sh
rockett-m: ACK, thanks and fixed

scripts/build-docker.sh
rockett-m: ACK, fixed

scripts/benchmarks.sh
rockett-m: ACK, fixed most

scripts/native-system-benchmark.sh
```diff
@@ -151,8 +166,8 @@ on_int() {
     printf 'Interrupting all components\n'
     trap '' SIGINT # avoid interrupting ourself
     for i in $PIDS; do # intentionally unquoted
-        if [[ -n "RECORD" ]]; then
-            kill -SIGINT -- "-$i"
+        if [[ -n "$RECORD" ]]; then
```
HalosGhost, line 155: Good catch! This is a bug!

rockett-m:
Divide and conquer PR new plan
Commit 1: ci.yml and shellcheck.sh and any fatal errors across *.sh files (so the stage can pass)
Commit 2: shellcheck info and warning messages fixed to *.sh files
Commit 3: style fixes etc and shebang update to *.sh scripts
Commit 4: use NUM_CORES and xargs in lint.sh for parallelization to `lint.sh`

scripts/shellcheck.sh
HalosGhost, line 1: I don't know that this script is really providing a lot of benefit; It's a lot of boilerplate for what is essentially just

```
git ls-files '*.sh' | xargs -n 1 -P "$CPUS" shellcheck
```

But it adds the ceremony of dep-checking, writing the output to a file, and some ad-hoc inspection of that file. The only strict benefit is that it offers a shortcut for people not familiar with shellcheck.

Especially since I expect the GH action that uses this script to be *much slower* than the native action (because the native action doesn't need to spin up a bare ubuntu container, clone the repo, install all the dependencies, and then run this script).

This is the strongest argument I've seen so far for a command-runner (the least interesting use of `Make`) instead of the ad-hoc script paradigm we've had so far in this repo.

I'd love comments/thoughts from @maurermi as well. At the very least, I think the GH action should be the native form instead of this; and that immediately calls into question the benefit this script offers.

scripts/create-e2e-report.sh
```diff
@@ -11,7 +11,7 @@ function readAndFormatLogs() {
         return
     fi
 
-    for logfile in $(ls $logdir); do
+    for logfile in "$logdir"/*; do
         logfile_path="$logdir/$logfile"
         logfile_content=$(cat $logfile_path)
```
HalosGhost:
```suggestion
        logfile_content=$(<"$logfile_path")
```

HalosGhost (changes requested):

scripts/shellcheck.sh
rockett-m: I find it very helpful for making shell script changes and being able to run `scripts/shellcheck.sh` to generate the report right away without having to remember the command above and then how to tee to a file etc.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+
+Usage: ./scripts/shellcheck.sh [-e|--exclude-code=CODE] [-S|--severity=LEVEL] [-v|--view]
+
+example: ./scripts/shellcheck.sh -e SC1091 -e SC1090 -S warning -v
+
+EOF
+}
+
+parse_cli_args() {
+    echo; echo "Command line arguments: $0 $*"; echo
```
HalosGhost: Are we planning to include this?

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+        elif [[ "$1" =~ ^-[^-] && $# -gt 1 && ! "$2" =~ ^- ]]; then
+            optarg="$2"
+            shft_cnt=2
+        elif [[ "$1" =~ ^-[^-] ]]; then
+            optarg="${1/??/}"
+        fi
+
+        case "$1" in
+            -S*|--severity*)
+                # don't let the user enter -S LEVEL more than once
```
HalosGhost: This is bad practice; it's much better to give higher-precedence to later-specified arguments (i.e., `--level=warn --level=error --level=info` should set `level` equal to `info`). This makes it so that aliases and functions can hard-code arguments to programs but still allow the caller to override them (creating an ad-hoc default-mechanism). Plus, it's actually just simpler: each specification of the argument just overwrites the previous value, and that's it!

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+                    printf "${RED}[Error]${RST_COLOR} Severity level already set to: '${CYAN}%s${RST_COLOR}'\n" "${SEVERITY}"
+                    _err=1
+                # valid if -S has any of 'info', 'warning', 'error'
+                elif [[ "${optarg}" == "info" || "${optarg}" == "warning" || "${optarg}" == "error" ]]; then
+                    SEVERITY="${optarg}"
+                else
+                    # continue and disregard invalid severity level
+                    printf "${RED}[Error]${RST_COLOR} invalid severity level: '${CYAN}%s${RST_COLOR}'\n" "${optarg}"
+                    _err=1
+                fi
```
HalosGhost:
```bash
case "$optarg" in
    info|warning|error) SEVERITY="$optarg";;
    *) <snip>…
esac
```

When in-doubt, prefer less boilerplate.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env bash
+
+CYAN="\e[36m"
+YELLOW="\e[33m"
+BLUE="\e[34m"
+RED="\e[31m"
+GREEN="\e[32m"
+RST_COLOR="\e[0m"
```
HalosGhost: I'm always a bit hesitant to use color in shell output. If nothing else, I think it has to be used sparingly, and this strikes me as a bit excessive. Is there a specific motivation for having such colorized output?

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+                else
+                    # continue and disregard invalid severity level
+                    printf "${RED}[Error]${RST_COLOR} invalid severity level: '${CYAN}%s${RST_COLOR}'\n" "${optarg}"
+                    _err=1
+                fi
+                shift "$shft_cnt"
+                ;;
+            -e*|--exclude-code*)
+                # strip whitespace from optarg
+                CODE=$(echo "${optarg}" | tr -d '[:space:]')
```
HalosGhost: This appears to be redundant since the next line ensures the full argument meets the format `SC[0-9]{4}`.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+                _help=1
+                ;;
+            *)
+                printf "${RED}[Error]${RST_COLOR} Unrecognized option: %s\n" "$1"
+                _err=1
+                shift "$shft_cnt"
+                ;;
+        esac
+
+        _invalid_arg_count=$(( "${_invalid_arg_count}" + 1 ))
```
HalosGhost:
```suggestion
        _invalid_arg_count=$(( _invalid_arg_count + 1 ))
```
In bash, there is a shortcut for accessing variable contents without `$` in arithmetic contexts; and, in fact, this tends to behave better. When you don't need to rely on parameter expansion, this is a good plan.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+        if [[ "$_err" -eq 1 ]]; then
+            printf "${RED}[Error]${RST_COLOR} Invalid argument: %s\n" "${optarg}"
+        fi
+
+    done
+
+    if [[ "$_invalid_arg_count" -gt 0 ]]; then
+        # print the help message just once if one or more invalid arguments
+        echo; echo -e "${RED}[Error]${RST_COLOR} Some invalid arguments were detected."; echo
+        show_usage
```
HalosGhost: If a command invocation is invalid, the program should stop. Continuing when you know something was specified wrongly is most likely to result in you doing something different than what was intended (and has been a major source of bugs and vulnerabilities in many projects).

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+
+check_shellcheck_install() {
+    if ! command -v shellcheck &>/dev/null; then
+        echo -e "${RED}[ERROR]${RST_COLOR} shellcheck is not installed."
+        echo "Run 'sudo ./scripts/install-build-tools.sh' to install shellcheck."
+        exit 1
+    fi
+}
+
+get_num_cores() {
```
HalosGhost, line 133: Why does this set a variable instead of returning a result?

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+    if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        NUM_CORES=$(grep -c ^processor /proc/cpuinfo)
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        NUM_CORES=$(sysctl -n hw.ncpu)
+    else
+        NUM_CORES=1
+    fi
+}
+
+run_shellcheck() {
```
HalosGhost: `shellcheck` supports it's own severity filter (`-S`, the same as this script uses). Why do we not just forward that argument directly? A whole lot of this script could be simplified.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+        echo "git is not installed. Using find to compile list of shell scripts..."; echo
+        if [[ -z "$EXCLUDE_CODES" ]]; then
+            find "$ROOT" -name '*.sh' -print0 | xargs -0 -n 1 -P "$NUM_CORES" shellcheck -s bash > "$SHELLCHECK_REPORT"
+        else
+            find "$ROOT" -name '*.sh' -print0 | xargs -0 -n 1 -P "$NUM_CORES" shellcheck -s bash --exclude="$EXCLUDE_CODES" > "$SHELLCHECK_REPORT"
+        fi
+    fi
+}
+
+check_report() {
```
HalosGhost: Can you clarify for me why you prefer having the report saved to a file instead of just sending to `stdout`? This adds a lot of complexity and requires you to deal with things like cache-invalidation.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,253 @@
...
+    # check if shellcheck report was created
+    check_report
+    # display report contents if view set in cli
+    view_report
+    # determine status by parsing shellcheck report to see if any messages
+    # of the severity level or more strict are present to determine failure (true errors)
+    determine_pass_fail
+}
+
+main "$@"
```
HalosGhost:
```suggestion
check_shellcheck_install
parse_cli_args "$@"
get_num_cores
run_shellcheck
check_report
view_report
determine_pass_fail
```

If even all of these make it in, there's no benefit to including a `main` function. And the comments here are all redundant as the functions being called are actually well-named (though, the control flow here should probably be extracted out of the internal functions so that, for example, `view_report` only even gets called if the user asked for it).

HalosGhost (changes requested):

scripts/shellcheck.sh
rockett-m: Optional, I could go either way on it

scripts/shellcheck.sh
rockett-m: ACK, I'll update the logic to allow this

scripts/shellcheck.sh
HalosGhost: Personally, I'd view this as unhelpful noise unless guarded by a debugging option.

scripts/shellcheck.sh
rockett-m: I find its much easier to understand the key insights quickly with the color. A friend who works in UI actually recommended I add it. We do have color output in some of the other scripts such as `./scripts/install-build-tools.sh` and `./scripts/setup-dependencies.sh`
failing:
shellcheck-output
passing:
shellcheck-pass
The shellcheck report itself `./shellcheck-report.txt` does not have color.

scripts/shellcheck.sh
rockett-m: It should be actually at the before stage to not always += 1.         
```console
_invalid_arg_count=$(( "${_invalid_arg_count}" + "${_err}" ))
```
ACK, changing to
```console
_invalid_arg_count=$(( _invalid_arg_count + _err ))
```
Edit: changed to exit immediately on error

scripts/shellcheck.sh
rockett-m: ACK, I'll put it under the help section so it only prints if triggered.

scripts/shellcheck.sh
rockett-m: ACK

scripts/shellcheck.sh
HalosGhost: The report can have color, because `shellcheck` supports configurable colorization.

Other scripts do have color, and we should unify them at some point to behave well for accessibility reasons (color can be helpful, but it can also be terrible; the most accessible option is to support configuration). However, most of those are either run-once, run-super-rarely, or are intended for pipelines and not really for humans (like the end-to-end kubernetes test). If we want to support colorization in this tool, we should at least support a basic way of disabling/enabling it (as `shellcheck` does).

The gold standard is to offer `-c|--color[=WHEN]` where `WHEN` can be one of `auto` (which disables color when the output is going to a pipe or file, and is a bit tricky to really get right), `always`, and `never` (defaulting to `auto`); and, to support the [`NO_COLOR`](https://no-color.org/) environment variable. We should, at least, support the env-var if not something more flexible.

scripts/shellcheck.sh
HalosGhost: I do not think this is helpful to have in the help output by-default.

scripts/shellcheck.sh
rockett-m: I allow for both. Send to file always, and `--view` option to see stdout.
When I am fixing errors I can search very easily in a file, but the stdout could exceed 1000 lines of output and be harder to view the errors of a code. Can grep the stdout but error messages take different amounts of line space so would have to play with `grep -B2 -A4` etc.
shellcheck-file

HalosGhost: You barely need to add anything to support outputting to a file (because shells support redirection). Is overriding this default behavior (how almost all other command-line utilities work) of-use? Why not just redirect the output to a file as you need to?

scripts/shellcheck.sh
rockett-m: ACK, thanks about the function naming.
I put this inside `check_report()`
```console
    if [[ "$VIEW" == "True" ]]; then
        view_report
    fi
```
and removed that logic from `view_report()` so it will shows the report if called.

scripts/shellcheck.sh
rockett-m: ACK - fixed

scripts/shellcheck.sh
rockett-m: Updated to printf the decimal value instead

scripts/shellcheck.sh
rockett-m: ACK - did that as well, but added logic to count waived severities and error codes and summarize all

scripts/shellcheck.sh
rockett-m: ACK - updated to be more efficient now

scripts/shellcheck.sh
rockett-m: It creates two files now so this wouldn't be trivial anymore

scripts/shellcheck.sh
rockett-m: ACK - gave this a shot

rockett-m:
@HalosGhost Incorporated your suggestions. Added more intelligent coloring, waived error codes/severity messages, summary report.

Can be viewed here: https://github.com/mit-dci/opencbdc-tx/actions/runs/10501604065/job/29091863763?pr=298

The waivers are more powerful when looking beyond errors (majority of the info warnings are about variable quotations 'SC2086')
```console
$ ./scripts/shellcheck.sh -C auto -S info -e SC2034 -e SC2086 -e SC1010
```
sh-chk

Edit: I will add a path to the full report as well in the next push

scripts/shellcheck.sh
HalosGhost: It now creates two files by running `shellcheck` twice every time, once at the requested severity level, and one bypassing the user's choice to make a "full" report. Why do this?

rockett-m: The default for shellcheck is 'style' and that seems too strict to make a github action/prevent code from getting in. My script defaults to 'error'.

There are over 100 messages of severity stricter than 'error' (warning, info, style) which I don't want myself or others to be blind to when turning in shell scripts or modifying them.

Very good to have the full report for commit 2 to mow down (warning, info, style) messages. And future reference. Plus being able to see if a certain code is a nuisance more than a help and to deliberate over making it an exclude if so.

HalosGhost: “It would be better to give contributors more feedback than is enforced or asked for” does not feel sufficient to me to rationalize the additional complication of bypassing the standard convention of printing output to `stdout` (which can be captured by file-redirection should the caller desire). Moreover, even though deriving from a positive motivation, taking an argument to specify severity level and then ignoring it intentionally (note that this will duplicate effort entirely if the user specifies the severity to be equal to `shellcheck`'s default) comes close to being user-hostile (and our users in this case, are contributors)!

If you feel strongly that we should prefer a less restrictive severity level by-default, make the argument for why (and we would instead default to that severity level, perhaps disabling specific codes we find less-than-useful).

HalosGhost (changes requested):
I'll make a full review after the file-output/severity issue has been appropriately resolved or justified.

scripts/shellcheck.sh
rockett-m: Error is the least restrictive severity level as anything more strict gets ignored in pass/fail of the stage. 
What if we changed it to warning or info as default and used whichever one in the gh action? Can discuss in slack/zoom if need be.

.github/workflows/ci.yml
```diff
@@ -62,10 +62,25 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - name: Install shellcheck
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y shellcheck
+      - name: Lint with Shellcheck
+        run: ./scripts/shellcheck.sh -S error --color=auto
```
rockett-m: severity `-S` likely to get more strict

scripts/shellcheck.sh
HalosGhost: My phrasing might have been confusing. I meant less-restrictive in the sense of “print more diagnostics”. As I alluded to, I'm comfortable with “raising” the default from error to something higher, but I would ask that you make an argument for *why* we should prefer that. I appreciate unifying standards across the repository, and improving the CI/CD, *and* raising the bar for quality of code committed; but, for a research project predominantly focused on engineering novel solutions and not-at-all currently driven by productionization, these improvements are quality-of-life, not top-priority or of most research-interest. Adding friction for contributors absolutely *must* be justified; it isn't free.

scripts/shellcheck.sh
```diff
@@ -0,0 +1,322 @@
+#!/usr/bin/env bash
+
+START_TIME=$(date "+%s")
+ROOT="$(cd "$(dirname "$0")"/.. && pwd)"
+SHELLCHECK_REPORT="${ROOT}/shellcheck-report.txt"
+SHELLCHECK_REPORT_ALL_MSGS="${ROOT}/shellcheck-report-all-msgs.txt"
+
+SEVERITY="error"
```
rockett-m: default severity likely to get more strict

scripts/shellcheck.sh
```diff
@@ -0,0 +1,322 @@
...
+    printf "%d\n" "$CORE_COUNT"
+}
+
+run_shellcheck() {
+    # check if git is installed
+    if command -v git &>/dev/null; then
+        if [[ -z "$EXCLUDE_CODES" ]]; then
+            git ls-files '*.sh' | xargs -n 1 -P "$NUM_CORES" shellcheck -s bash -C"$COLOR" > "$SHELLCHECK_REPORT_ALL_MSGS"
+            git ls-files '*.sh' | xargs -n 1 -P "$NUM_CORES" shellcheck -s bash -C"$COLOR" -S "$SEVERITY" > "$SHELLCHECK_REPORT"
+        else
```
rockett-m: can just pass empty or populated exclude codes into args and it will work, so can get rid of extra if/else

scripts/shellcheck.sh
```diff
@@ -0,0 +1,243 @@
...
+        echo; echo -e "${RED}Shellcheck failed. Please fix the issues and try again.${RST_COLOR}"
+    fi
+    exit "$EXIT_STATUS"
+}
+
+main() {
+    check_shellcheck_install
+    parse_cli_args "$@"
+
+    if [[ "$COLOR" != "never" ]]; then
```
HalosGhost, line 220: This is fine, but `auto` should be removed from the help output, it's not being handled (which is okay! handling auto-coloring is a bit sticky).

PR #300, opened 2024-08-21
Convert assert in create into a warning

maurermi:
Turn the assert statement in create into a warning, this restores the ability to import tokens through Metamask.

Details: When importing tokens, Metamask sends a create call without sender information, triggering an assert failure.


maurermi:
@HalosGhost This should fix the error we discussed on a call last Friday

HalosGhost (approved):
Replaces an `assert` with a valid/expected code-path to handle MetaMask's behavior changing unexpectedly.

t-ACK, I was able to get all the way through importing tokens into MetaMask without error or crash.

PR #306, opened 2024-10-21
fix: don't track raft logs

nicol-ii:
Add generated raft logs to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,11 @@ plots/
 .deps/
 .libs/
 
+runtime_locking_shard*_raft*
+runtime_locking_shard*_raft*/
+ticket_machine_raft_*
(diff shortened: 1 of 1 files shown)


maurermi (approved):
T-ACK; This change simply adds generated files to the .gitignore, so they are not mistakenly checked in during development. Assuming this passes CI tests, this should be good to go

maurermi:
@HalosGhost I would say this is good to go, but will defer to you on merging it in

HalosGhost (approved):
t-ACK. I'm happy to merge this, but it looks like it only covers the generated raft logs from parsec. I'd gladly also accept a PR to ignore the UHS architectures' raft logs (as well as all other files generated during our component runtimes).

Will Merge.

PR #307, opened 2024-10-28
upgrade: Upgrade nuraft to 2.1.0

maurermi:
Upgrades NuRaft to v2.1.0. 

This upgrade caused a failing assertion in the `raft_node_test_non_blocking` unit test. I claim that this assertion in the non_blocking test was actually misguided, and as written, should have always failed in the non_blocking context (or more accurately, the assertion gave insufficient information about whether the task completed correctly). 

This test fails because when getting the result code of a replication which is asynchronous, if the result has not been written yet, the return value is not cmd_result_code::OK (it becomes `cmd_result_code::RESULT_NOT_EXIST_YET`). In the non-blocking context, this should not prompt a failure, specifically with `node::replicate_sync` written as it is currently. The reason this does not fail with v1.3 is because `cmd_result_code::RESULT_NOT_EXIST_YET` did not exist, and `get_result_code()` returned the default value of `cmd_result_code`, which appeared to be `cmd_result_code::OK`. 

Would appreciate feedback about whether we should re-write `replicate_sync` to be blocking in a non-blocking node, and wait to see if the result code is truly OK before proceeding. It does not seem like this has been the case to this point (but I have a hunch that it should be), but happy to be wrong on this point. This change is not included in this PR at the time of writing.

### DIFF
--- a/scripts/setup-dependencies.sh
+++ b/scripts/setup-dependencies.sh
@@ -39,7 +39,7 @@ make -j$CPUS
 make install
 cd ..
 
-NURAFT_VERSION="1.3.0"
+NURAFT_VERSION="2.1.0"
 echo -e "${green}Building NuRaft from sources...${end}"
 wget https://github.com/eBay/NuRaft/archive/v${NURAFT_VERSION}.tar.gz
 rm -rf "NuRaft-${NURAFT_VERSION}-${CMAKE_BUILD_TYPE}"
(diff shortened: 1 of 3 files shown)


maurermi:
@HalosGhost would be great to get your opinion in particular on what to do about `replicate_sync`

maurermi:
Bumping this out of draft

maurermi:
Unfortunately, I believe that the failing CI test is related to the image that test is run against, which is based on trunk. As a dependency was updated in this PR, that will continue to fail.

HalosGhost (approved):
t-ACK; this is a really simple modification to handle updating our NuRaft dependency to 2.1.0.

PR #308, opened 2024-10-30
chore: fix some warnings on clang 16.0.0

maurermi:
Addresses three minor "shadow-uncaptured-local" warnings. Noticed that these are raised by Clang 16.0.0 which prevents compilation with -Werror. Minor QOL change for users of Clang 16.0.0


HalosGhost (approved):
Really basic change to satisfy newer clang's opinions about shadowing locals. Looks good to me. 👍

PR #313, opened 2024-11-27
fix: makes parsec endpoint names scalable; fixes #312

### LINKED ISSUE #312: Use a number separator for parsec `shard##_endpoint` to prevent double-digit confusion
### Affected Branch

trunk

### Basic Diagnostics

### Description

For endpoints in parsec, the option parsing for the following machines {agent, ticket machine, and shard} goes like this:

`src/parsec/util.cpp`, line 63
```console
for(size_t i = 0; i < count; i++) {
    auto ep_key = component_name + std::to_string(i) + "_endpoint";
}
```

- Agents aren't replicated, so `agent0_endpoint` to `agent{n-1}_endpoint` works.

- The ticket machine can be replicated but there is typically just one, so a single number works.

- We can have many shards, and shard replicas; however, so the above code should be modified for double-digit compatibility.

In the trivial case (1 shard, 0 replicas) the representation `shard00_endpoint` is fair enough.
For many shards; however, we should use an underscore to separate the values like 2pc does. 
https://github.com/mit-dci/opencbdc-tx/blob/96062b5b9a3cd9406ea4e0174829be8d0e80f18a/2pc-compose.cfg#L18

For example, we can't tell which shard or replica `shard111_endpoint` refers to if we have 50 shards and 15x replication. The underscore fixes that: `shard11_1_endpoint`. Even if we don't plan on double-digit replication, the separation helps readability.

The potential for many shards can be seen at the bottom of the diagram. 
[parsec-diagram](https://github.com/mit-dci/opencbdc-tx/blob/trunk/docs/parsec_architecture_diagram.svg)

### Code of Conduct

rockett-m:
Before: `shard##_endpoint`
After:     `shard#_#_endpoint`

This PR let's parsec use the same naming convention as 2pc for shard endpoints.

Also, credit to @madars for pointing out the unnecessary `--shard00_raft_endpoint=$IP:5557`.
https://github.com/mit-dci/opencbdc-tx/blob/96062b5b9a3cd9406ea4e0174829be8d0e80f18a/scripts/parsec-run-local.sh#L46
That could be a separate commit or PR based on reviewer discretion

### DIFF
--- a/docker-compose-parsec-test.yml
+++ b/docker-compose-parsec-test.yml
@@ -9,7 +9,7 @@ services:
     image: opencbdc-tx-parsec
     platform: linux/amd64
     tty: true
-    command: ./scripts/wait-for-it.sh -s agent0:8080 -t 60 -- ./build/tools/bench/parsec/evm/evm_bench --shard_count=1 --shard0_count=1 --shard00_endpoint=shard0:5556 --node_id=0 --component_id=0 --agent_count=1 --agent0_endpoint=agent0:8080 --ticket_machine_count=1 --ticket_machine0_endpoint=ticket0:7777 --loadgen_accounts=8192 --loadgen_txtype=erc20 --telemetry=1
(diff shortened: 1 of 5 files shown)


rockett-m:
@maurermi @madars could you review this? Thanks

PR #314, opened 2025-02-25
feat: expose secp hashing to lua environment

nicol-ii:
feat: expose secp hashing to lua environment

Exposes a hash function to the Lua smart contract execution environment.
Additionally adds a handful of library version bumps and other cleanup.


maurermi:
This is great! @nicol-ii if you can fix up the linter errors then we should be able to push this forward some

scripts/gen_bytecode.lua
```diff
@@ -43,7 +43,11 @@ function gen_bytecode()
         from_balance, from_seq = get_account(from)
         payload = sig_payload(to, value, sequence)
         check_sig(from, sig, payload)
-        if sequence < from_seq then
+	
+	hash = make_hash("Hello World!")
+	print(hash)      
+
+	if sequence < from_seq then
```
HalosGhost: Presumably, this is meant to test that the newly-exposed function works; can this be verified using a real test instead of adding IO to the bytecode generation?

On that note, when testing the exposure of a hash-function (or any data-transformation) specifically, it's probably a good idea to verify that it works against known values (i.e., I'd suggest ``assert``ing that the hash from ``"Hello World!"`` is what you expect).

src/parsec/agent/runners/lua/impl.cpp
```diff
@@ -299,4 +300,28 @@ namespace cbdc::parsec::agent::runner {
...
+            }
+        size_t sz{};
+        const auto* str = lua_tolstring(L, 1, &sz);
+        assert(str != nullptr);
+        auto sha = CSHA256();
+        auto unsigned_str = std::vector<unsigned char>(sz);
+        std::memcpy(unsigned_str.data(), str, sz);
+        sha.Write(unsigned_str.data(), sz);
+        cbdc::hash_t computed_hash{};
+        sha.Finalize(computed_hash.data());
```
HalosGhost: `cbdc::hash_data()` is effectively this snippet refactored into a method; I might suggest using it here directly.

HalosGhost (changes requested):
A simple and clear exposure of internal hashing to the Lua runner. There are just a couple instances of small tidying that occur to me. Well done!

PR #315, opened 2025-03-11
fix: patch lua unit test

maurermi:
Noticed that a lua test was not actually running correctly, this fixes that.

To verify, run `./scripts/test.sh -ni -nc --gtest_filter="*lua*"` and notice that two tests pass. These tests were not running previously.


maurermi:
Puzzling to me what's wrong here, I'll need to pick this apart

PR #316, opened 2025-03-19
chore: pin all non-docker/non-GH action versions

HalosGhost:
This avoids the issue of malleable version tags in the event of repo compromise.

Note that this leaves the tagged versions from official GH/docker actions out of a hope to minimize churn/maintenance-burden.

Includes a NEWS post.

cf. https://www.cve.org/CVERecord?id=CVE-2025-30066


maurermi (approved):
ACK; Pins a few workflows to stable versions, looks like they all passed in CI so this should be good to go in.

HalosGhost:
All builds and runs fine for me as well, so I'm calling it good.

